Issue #38, opened 2014-10-07
"hosts" parameter in JSON request is ignored when signing a new certificate


ghost:
I'm using the following boilerplate JSON request, `csrjson`:

``` javascript
{
    "hosts": [
        "example.com",
        "www.example.com"
    ],
    "key": {
        "algo": "rsa",
        "size": 2048
    },
    "names": [
        {
            "C": "US",
            "L": "San Francisco",
            "O": "Internet Widgets, Inc.",
            "OU": "WWW",
            "ST": "California"
        }
    ]
}
```

After creating a suitable CA certificate, when attempting to generate and sign a new client certificate:

```
cfssl gencert example.com csrjson | cfssljson -bare
```

I would expect it to include `example.com` and `www.example com` as Subject Alternative Names. But instead, it only includes the specified hostname, `example.com`:

```
luna@luna-VirtualBox:/tmp/tmp$ openssl x509 -in cert.pem -noout -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 7662139354499468246 (0x6a556325ac7b4bd6)
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=US, O=Internet Widgets, Inc., OU=WWW, L=San Francisco, ST=California
        Validity
            Not Before: Oct  7 02:59:22 2014 GMT
            Not After : Oct  7 03:04:22 2015 GMT
        Subject: C=US, O=Example Organisation, L=San Francisco, ST=California, CN=example.com
        Subject Public Key Info:
... snip ...
            X509v3 Subject Alternative Name: 
                DNS:example.com
    Signature Algorithm: sha256WithRSAEncryption
... snip ...
```

I believe this is not the intended behavior. Would anyone please confirm this?

PR #49, opened 2014-10-28
Use all host names in the request for signing.

kisom:
This is based on #38.


thebluesnevrdie:
This is still broken in version 1.6.1

PR #73, opened 2015-01-19
Initial CLI Version of `cfssl scan`

jacobhaven:
Adds new `scan` command, that performs a variety of connectivity and TLS protocol tests.

Unfortunately, some of these require importing the go standard library's `crypto/tls` package's source in order to access private members. These are used in `scan/tls/cfsslscan_*.go` files to export additional functionality not supplied by the `crypto/tls` API.

Support for the standard CFSSL HTTP API is forthcoming, but comments on this initial structure might be useful first.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -44,7 +44,6 @@ type Command struct {
 	Main func(args []string, c Config) error
 }
 
-// Parsed command name
 var cmdName string
 
 // usage is the cfssl usage heading. It will be appended with names of defined commands in cmds
--- a/cli/config.go
+++ b/cli/config.go
@@ -43,6 +43,9 @@ type Config struct {
 	Reason            int
 	RevokedAt         string
 	Interval          int64
+	List              bool
+	Family            string
+	Scanner           string
 }
 
 // registerFlags defines all cfssl command flags and associates their values with variables.
@@ -74,6 +77,9 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.IntVar(&c.Reason, "reason", 0, "Reason code for revocation")
 	f.StringVar(&c.RevokedAt, "revoked-at", "now", "Date of revocation (YYYY-MM-DD)")
 	f.Int64Var(&c.Interval, "interval", int64(4*helpers.OneDay), "Interval between OCSP updates, in seconds (default: 4 days)")
+	f.BoolVar(&c.List, "list", false, "list possible scanners")
+	f.StringVar(&c.Family, "family", "", "scanner family regular expression")
+	f.StringVar(&c.Scanner, "scanner", "", "scanner regular expression")
 
 	if pkcs11.Enabled {
 		f.StringVar(&c.Module, "pkcs11-module", "", "PKCS #11 module")
--- a/cli/scan/scan.go
+++ b/cli/scan/scan.go
@@ -1,0 +1,106 @@
+package scan
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"os"
+	"regexp"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/cloudflare/cfssl/cli"
+	"github.com/cloudflare/cfssl/log"
+	"github.com/cloudflare/cfssl/scan"
+)
+
+var scanUsageText = `cfssl scan -- scan a host for issues
+Usage of scan:
+        cfssl scan [-family regexp] [-scanner regexp] HOST+
+        cfssl scan -list [-family regexp] [-scanner regexp]
+
+Arguments:
+        HOST:    Host(s) to scan (including port)
+Flags:
+`
+var scanFlags = []string{"list", "family", "scanner"}
+
+// regexpLoop iterates through each scan Family and Scanner registered in scan.AllFamilies.
+// familyFunc is run on each Family with a name matching the family flag's regexp,
+// then scannerFunc is run on each Scanner in that Family that matches the scanner flag's regexp
+func regexpLoop(familyRegexp, scannerRegexp *regexp.Regexp,
+	familyFunc func(*scan.Family),
+	scannerFunc func(*scan.Scanner)) {
+	for _, family := range scan.AllFamilies {
+		if familyRegexp.MatchString(family.Name) {
+			familyFunc(family)
+			for _, scanner := range family.Scanners {
+				if scannerRegexp.MatchString(scanner.Name) {
+					scannerFunc(scanner)
+				}
+			}
+		}
+	}
+}
+
+// indentPrintln prints a multi-line block with the specified indentation.
+func indentPrintln(block string, indentLevel int) {
+	w := tabwriter.NewWriter(os.Stdout, 0, 4, 4, ' ', 0)
+	scanner := bufio.NewScanner(strings.NewReader(block))
+	for scanner.Scan() {
+		fmt.Fprintf(w, "%s%s\n", strings.Repeat("\t", indentLevel), scanner.Text())
+	}
+	w.Flush()
+}
+
+func scanMain(args []string, c cli.Config) (err error) {
+	familyRegexp, err := regexp.Compile(c.Family)
+	scannerRegexp, err := regexp.Compile(c.Scanner)
+	if err != nil {
+		return
+	}
+	if c.List {
+		regexpLoop(
+			familyRegexp, scannerRegexp,
+			func(f *scan.Family) {
+				fmt.Println(f)
+			},
+			func(s *scan.Scanner) {
+				indentPrintln(s.String(), 1)
+			},
+		)
+	} else {
+		// Execute for each HOST argument given
+		for len(args) > 0 {
+			var host string
+			host, args, err = cli.PopFirstArgument(args)
+			if err != nil {
+				return
+			}
+			// If no port is specified, default to 443
+			if _, _, err := net.SplitHostPort(host); err != nil {
+				host = net.JoinHostPort(host, "443")
+			}
+			log.Infof("Scanning %s...", host)
+			regexpLoop(
+				familyRegexp, scannerRegexp,
+				func(f *scan.Family) {
+					if log.Level == log.LevelDebug && c.Scanner == "" {
+						fmt.Printf("[%s]\n", f.Name)
+					}
+				},
+				func(s *scan.Scanner) {
+					grade, output, _ := s.Scan(host)
+					fmt.Printf("%s: %s\n", s.Name, grade)
+					if log.Level == log.LevelDebug && output != nil {
+						indentPrintln(output.String(), 1)
+					}
+				},
+			)
+		}
+	}
+	return
+}
+
+// Command assembles the definition of Command 'scan'
+var Command = &cli.Command{UsageText: scanUsageText, Flags: scanFlags, Main: scanMain}
--- a/scan/connectivity.go
+++ b/scan/connectivity.go
@@ -1,0 +1,81 @@
+package scan
+
+import (
+	"errors"
+	"net"
+	"strings"
+
+	"github.com/cloudflare/cf-tls/tls"
+)
+
+// Connectivity contains scanners testing basic connectivity to the host
+var Connectivity = &Family{
+	Name:        "Connectivity",
+	Description: "Scans for basic connectivity with the host through DNS and TCP/TLS dials",
+	Scanners: []*Scanner{
+		{
+			"DNSLookup",
+			"Host can be resolved through DNS",
+			dnsLookupScan,
+		},
+		{
+			"TCPDial",
+			"Host accepts TCP connection",
+			tcpDialScan,
+		},
+		{
+			"TLSDial",
+			"Host can perform TLS handshake",
+			tlsDialScan,
+		},
+	},
+}
+
+// lookupAddrs is a list of host's addresses returned by DNS lookup
+type lookupAddrs []string
+
+func (addrs lookupAddrs) String() string {
+	return strings.Join(addrs, "\n")
+}
+
+// dnsLookupScan tests that DNS resolution of the host returns at least one address
+func dnsLookupScan(host string) (grade Grade, output Output, err error) {
+	host, _, err = net.SplitHostPort(host)
+	if err != nil {
+		return
+	}
+
+	var addrs lookupAddrs
+	addrs, err = net.LookupHost(host)
+	if err != nil {
+		return
+	}
+
+	if len(addrs) == 0 {
+		err = errors.New("no addresses found for host")
+	}
+	grade, output = Good, addrs
+	return
+}
+
+// tcpDialScan tests that the host can be connected to through TCP.
+func tcpDialScan(host string) (grade Grade, output Output, err error) {
+	conn, err := Dialer.Dial(Network, host)
+	if err != nil {
+		return
+	}
+	conn.Close()
+	grade = Good
+	return
+}
+
+// tlsDialScan tests that the host can perform a TLS Handshake.
+func tlsDialScan(host string) (grade Grade, output Output, err error) {
+	conn, err := tls.DialWithDialer(Dialer, Network, host, defaultTLSConfig(host))
+	if err != nil {
+		return
+	}
+	conn.Close()
+	grade = Good
+	return
+}
--- a/scan/scan_common.go
+++ b/scan/scan_common.go
@@ -1,0 +1,116 @@
+package scan
+
+import (
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/cloudflare/cf-tls/tls"
+	"github.com/cloudflare/cfssl/log"
+)
+
+var (
+	// Network is the default network to use.
+	Network = "tcp"
+	// Dialer is the default dialer to use, with a 1s timeout.
+	Dialer = &net.Dialer{Timeout: time.Second}
+)
+
+// Grade gives a subjective rating of the host's success in a scan.
+type Grade int
+
+const (
+	// Bad describes a host with serious misconfiguration or vulnerability.
+	Bad Grade = iota
+	// Legacy describes a host with non-ideal configuration that maintains support for legacy clients.
+	Legacy
+	// Good describes host performing the expected state-of-the-art.
+	Good
+	// Skipped descibes the "grade" of a scan that has been skipped.
+	Skipped
+)
+
+// String gives the name of the Grade as a string.
+func (g Grade) String() string {
+	switch g {
+	case Bad:
+		return "Bad"
+	case Legacy:
+		return "Legacy"
+	case Good:
+		return "Good"
+	case Skipped:
+		return "Skipped"
+	default:
+		return "Invalid"
+	}
+}
+
+// Output is the result of a scan, to be stored for potential use by later Scanners.
+type Output interface {
+	fmt.Stringer
+}
+
+// Scanner describes a type of scan to perform on a host.
+type Scanner struct {
+	// Name provides a short name for the Scanner.
+	Name string
+	// Description describes the nature of the scan to be performed.
+	Description string
+	// scan is the function that scans the given host and provides a Grade and Output.
+	scan func(host string) (Grade, Output, error)
+}
+
+// Scan performs the scan to be performed on the given host and stores its result.
+func (s *Scanner) Scan(host string) (Grade, Output, error) {
+	grade, output, err := s.scan(host)
+	if err != nil {
+		log.Infof("%s: %s", s.Name, err)
+		return grade, output, err
+	}
+	return grade, output, err
+}
+
+// String gives the name of the Scanner, and its description if loglevel is 0.
+func (s *Scanner) String() string {
+	ret := fmt.Sprintf("%s", s.Name)
+	if log.Level == log.LevelDebug {
+		ret += fmt.Sprintf(": %s", s.Description)
+	}
+	return ret
+}
+
+func defaultTLSConfig(host string) *tls.Config {
+	h, _, err := net.SplitHostPort(host)
+	if err != nil {
+		h = host
+	}
+	return &tls.Config{ServerName: h, InsecureSkipVerify: true}
+}
+
+// Family defines a set of related scans meant to be run together in sequence.
+type Family struct {
+	// Name is a short name for the Family.
+	Name string
+	// Description gives a short description of the scans performed scan/scan_common.goon the host.
+	Description string
+	// Scanners is a list of scanners that are to be run in sequence.
+	Scanners []*Scanner
+}
+
+// String gives the name of the Family, and its description if loglevel is 0.
+func (f *Family) String() string {
+	ret := fmt.Sprintf("%s", f.Name)
+	if log.Level == 0 {
+		ret += fmt.Sprintf(": %s", f.Description)
+	}
+	return ret
+}
+
+// AllFamilies contains each scan Family that is defined
+var AllFamilies = []*Family{
+	Connectivity,
+	TLSHandshake,
+	TLSSession,
+	PKI,
+}
--- a/scan/tls_handshake.go
+++ b/scan/tls_handshake.go
@@ -1,0 +1,120 @@
+package scan
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/cloudflare/cf-tls/tls"
+)
+
+// TLSHandshake contains scanners testing host cipher suite negotiation
+var TLSHandshake = &Family{
+	Name:        "TLSHandshake",
+	Description: "Scans for host's SSL/TLS version and cipher suite negotiation",
+	Scanners: []*Scanner{
+		{
+			"CipherSuite",
+			"Determines host's cipher suites accepted and prefered order",
+			cipherSuiteScan,
+		},
+	},
+}
+
+func sayHello(host string, ciphers []uint16, vers uint16) (cipherIndex int, err error) {
+	tcpConn, err := net.Dial(Network, host)
+	if err != nil {
+		return
+	}
+	config := defaultTLSConfig(host)
+	config.MinVersion = vers
+	config.MaxVersion = vers
+	config.CipherSuites = ciphers
+	conn := tls.Client(tcpConn, config)
+	serverCipher, serverVersion, err := conn.SayHello()
+	conn.Close()
+	if err != nil {
+		return
+	}
+
+	if serverVersion != vers {
+		err = fmt.Errorf("server negotiated protocol version we didn't send: %s", tls.Versions[serverVersion])
+		return
+	}
+
+	var cipherID uint16
+	for cipherIndex, cipherID = range ciphers {
+		if serverCipher == cipherID {
+			return
+		}
+	}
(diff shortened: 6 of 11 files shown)


scan/scan_common.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Scanner describes a type of scan to perform on a host.
+type Scanner struct {
+	Name        string
+	Description string
+	scan        func(host string) (Grade, Output, error)
+}
+
+// Scan defines the scan to be performed on the given host.
+func (s *Scanner) Scan(host string) (grade Grade, output Output, err error) {
```
lziest: Looks good. One style issue: For exported function, we keep the return list nameless, so the doc will be consistent with standard GoDoc. Also, it makes each return statement declare exactly what should be returned. It's much easier to maintain afterwards, and reduces potential bugs.

scan/tls_ciphersuite_scanner.go
```diff
@@ -0,0 +1,65 @@
+package scan
+
+import (
+	"errors"
+	"net"
+
+	"github.com/jacobhaven/cipherscan/tls"
```
lziest: Some compilation error regarding this package.

scan/tls_session_scanner.go
```diff
@@ -0,0 +1,63 @@
...
+	addrs, err := net.LookupHost(host)
+	if err != nil {
+		return
+	}
+	sessions := newSessionCache()
+	output = sessions
+	config := &tls.Config{ClientSessionCache: sessions}
+	for _, addr := range addrs {
+		var conn *tls.Conn
+		conn, err = tls.Dial(Network, addr, config)
```
lziest: I recommend to use DialWithDialer, so we can control the timeout and easy to tweak. It's otherwise the system default timeout, which could be like 30 seconds.

cfssl_scan.go
```diff
@@ -0,0 +1,82 @@
...
+	"regexp"
+
+	"github.com/cloudflare/cfssl/scan"
+)
+
+var scanUsageText = `cfssl scan -- scan a host for issues
+
+Usage of scan:
+        cfssl scan [-short] [SCANFAMILY] HOST
+        cfssl scan [-short] [-scanner regexp] HOST
```
grittygrease: Please put an example here. The documentation is confusing, is SCANFAMILY a regexp? What is regexp matching against? What does [SCANFAMILY]* represent?

scan/tls_ciphersuite_scanner.go
grittygrease: Don't we want the version from cfssl/scan/tls here?

jacobhaven: Ah, well that would explain something. \*updating\*

cfssl_scan.go
jacobhaven: I've refactored this to be a normal argument `-scanfamily regexp`

cfssl.go
```diff
@@ -67,6 +67,9 @@ var Config struct {
 	domain            string
 	ip                string
 	remote            string
+	short             bool
+	list              bool
+	scanner           string
```
grittygrease: Need to register flags and command (CLIScan) below.

scan/connectivity_scanner.go
```diff
@@ -0,0 +1,57 @@
...
+	Description: "Check for basic connectivity with the host",
+	Scanners: []*Scanner{
+		DNSLookupScanner,
+		TCPDialScanner,
+		TLSDialScanner,
+	},
+}
+
+// DNSLookupScanner tests that DNS resolution of the host returns at least one address
+var DNSLookupScanner = &Scanner{
```
kisom: Should use the field names in initialisation.

scan/connectivity_scanner.go
```diff
@@ -0,0 +1,57 @@
+package scan
+
+import "net"
+
+// ConnectivityScans contains scanners testing basic connectivity to the host
+var ConnectivityScanners = &Family{
+	Name:        "ConnectivityScanners",
```
kisom: Style note: if Family denotes a family of scanners, you can probably drop the "Scanners" part of the name. if this is the name that the user selects the family with on the commandline, I would just use "connectivity" to make it easier and consistent with the other args.

scan/connectivity_scanner.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func dnsLookupScan(host string) (grade Grade, output Output, err error) {
+	var conn net.Conn
+	addrs, err := net.LookupHost(host)
+	if err != nil || len(addrs) == 0 {
+		output = []string{}
+		return
+	}
+	defer conn.Close()
```
kisom: This shouldn't be deferred; the conn is no longer in use and should be closed right away.

scan/scan_common.go
```diff
@@ -0,0 +1,78 @@
...
+	// Legacy describes a host with non-ideal configuration that maintains support for legacy clients.
+	Legacy
+	// Good describes host performing the expected state-of-the-art.
+	Good
+)
+
+// Output is the result of a Scan, to be stored for potential use by later Scanners.
+type Output interface{}
+
+// Scanner describes a type of scan to perform on a host.
```
kisom: One thing I'm trying to move to in CFSSL is to make sure structure fields are documented as well.

scan/scan_common.go
```diff
@@ -0,0 +1,78 @@
...
+	output = outputs[len(outputs)-1]
+	return
+}
+func (h *scanHistory) Store(scannerName, host string, output Output) {
+	key := scannerHostPair{scannerName, host}
+	h.history[key] = append(h.history[key], output)
+}
+
+// Family defines a set of related scans meant to be run together in sequence
+type Family struct {
```
kisom: These field comments are good. As a Go style note though, they should be complete sentences.

scan/connectivity_scanner.go
jacobhaven: I didn't , this was a a relic of before this test was split from another one (It actually caused a bug and was removed)

cfssl_scan.go
```diff
@@ -0,0 +1,82 @@
...
+        cfssl scan [-list] [SCANFAMILY]*
+
+Arguments:
+        HOST:    Host to scan (including port)
+        SCANFAMILY:	 Regexp selecting family of related scanners to run
```
grittygrease: Inconsistency with spaces v. tabs.

cfssl_scan.go
```diff
@@ -0,0 +1,82 @@
...
+						fmt.Printf("\t%s:\n\t\t%s", scanner.Name, scanner.Description)
+					}
+				}
+			}
+		}
+	} else {
+		if Config.scanner != "" {
+			scannerRegexp = regexp.MustCompile(Config.scanner)
+		}
+		for len(args) > 1 {
```
grittygrease: This logic needs some work. calling cfssl scan with no arguments causes scanFamily.Name to be nil and 
			matches := scanFamilyRegexp.MatchString(scanFamily.Name)
to cause a panic.

scan/tls/alert.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2009 The Go Authors. All rights reserved.
```
grittygrease: It would be very helpful to have some documentation here. At the minimum we need a revision of the original code this was copied from and a diff to see which changes are custom.

grittygrease: It might even make sense to leave this as its own repository (forked from crypto/tls) so that we can track the diffs and keep it up to date with Go's tls changes.

kisom:
You've checked in the CFSSL binary. Please remove this (and make sure the commits are squashed so that it doesn't appear in any commits).

scan/tls_handshake_scanners.go
```diff
@@ -0,0 +1,65 @@
...
+	tcpConn, err := net.Dial(Network, host)
+	if err != nil {
+		return
+	}
+	conn := tls.Client(tcpConn, &tls.Config{ServerName: host, CipherSuites: ciphers})
+	defer conn.Close()
+	serverCipher, serverVersion, err = conn.SayHello()
+	for _, cipherID := range ciphers {
+		if serverCipher == cipherID {
+			err = errors.New("Server negotiated ciphersuite we didn't send.")
```
kisom: As a style note, Go error strings are generally not capitalised and don't end in punctuation.

scan/tls/alert.go
jacobhaven: This isn't a bad idea. To set it up as a submodule. I don't think we'll need to make any changes to the original `crypto/tls` files. I've just needed to access some functionality in creating the helper functions in `tls/cfsslscan_*`

grittygrease: No need to use a submodule in Go, you can just include from Github. For safety, the module name could be changed to scan-tls, cf-tls or something like that.

cfssl_scan.go
```diff
@@ -0,0 +1,103 @@
...
+					if err != nil {
+						return
+					}
+				}
+			}
+		}
+	}
+	return
+}
+func scanMain(args []string) (err error) {
```
jgrahamc: You've mixed the user of a named return argument and returns with arguments. Choose one or the other. I'd suggest making the returned value just 'error' and using explicit 'return foo'.

cfssl_scan.go
```diff
@@ -0,0 +1,103 @@
...
+        cfssl scan -list [-short] [-scanfamily regexp] [-scanner regexp]
+
+Arguments:
+        HOST:    Host(s) to scan (including port)
+Flags:
+`
+var scanFlags = []string{"short", "list", "family", "scanner"}
+var scanFamilyRegexp, scannerRegexp *regexp.Regexp
+
+func regexpLoop(familyFunc func(*scan.Family) error, scannerFunc func(*scan.Scanner) error) (err error) {
```
jgrahamc: I think its worth documenting these arguments as they are rather complex and it's not immediately clear what's expected of a familyFunc or a scannerFunc.

scan/connectivity.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// dnsLookupScan tests that DNS resolution of the host returns at least one address
+func dnsLookupScan(host string) (grade Grade, output Output, err error) {
+	host, _, err = net.SplitHostPort(host)
+	if err != nil {
+		return
+	}
+	addrs, err := net.LookupHost(host)
+	if err != nil || len(addrs) == 0 {
```
jgrahamc: If err == nil but len(addrs) == 0 this will return nil. I think you want something else.

cfssl.go
```diff
@@ -94,6 +98,10 @@ func registerFlags() {
 	cfsslFlagSet.StringVar(&Config.domain, "domain", "", "remote server domain name")
 	cfsslFlagSet.StringVar(&Config.ip, "ip", "", "remote server ip")
 	cfsslFlagSet.StringVar(&Config.remote, "remote", "", "remote CFSSL server")
+	cfsslFlagSet.BoolVar(&Config.short, "short", false, "so not run long/intensive scans")
```
jgrahamc: Not sure what 'so no run long/intensive scans' means.

cfssl.go
```diff
@@ -67,6 +67,10 @@ var Config struct {
 	domain            string
 	ip                string
 	remote            string
+	short             bool
+	list              bool
+	family            string
```
jgrahamc: These need documenting. It's very unclear how family and scanner interact.

scan/scan_common.go
```diff
@@ -0,0 +1,118 @@
...
+func (s *Scanner) String() string {
+	ret := fmt.Sprintf("\t%s", s.Name)
+	if !Short {
+		ret += fmt.Sprintf(":\n\t\t%s", s.Description)
+	}
+	return ret
+}
+
+type scannerHostPair struct{ scannerName, host string }
+type scanHistory struct {
```
jgrahamc: If all you have in scanHistory is this map then just do
```
type scanHistory map[scannerHostPair][]Output
```
and simplify your life.

scan/tls_handshake.go
```diff
@@ -0,0 +1,109 @@
...
+func (ciphers cipherList) String() (list string) {
+	for _, cipher := range ciphers {
+		list += fmt.Sprintf("%s(%#04x), ", tls.CipherSuites[cipher], cipher)
+	}
+	if len(list) > 2 {
+		list = list[:len(list)-2]
+	}
+	return list
+}
+func (ciphers cipherList) Append(cipher uint16) {
```
jgrahamc: This function seems a bit pointless.

scan/tls_session.go
```diff
@@ -0,0 +1,87 @@
...
+		{
+			"SessionResumeScanner",
+			"Host is able to resume sessions accross all addresses.",
+			sessionResumeScan,
+		},
+	},
+}
+
+// sessionCache is a mock ClientSessionCache used to record all stores (i.e. new session tickets)
+type sessionCache struct {
```
jgrahamc: This could just be 
```
type sessionCache map[string][]*tls.ClientSessionState
```

scan/tls_session.go
```diff
@@ -0,0 +1,87 @@
...
+	cache map[string][]*tls.ClientSessionState
+}
+
+func newSessionCache() *sessionCache {
```
jgrahamc: If you make the change above then this function is unnecessary and you can just do 
```
make(sessionCache)
```
whenever one is needed.

scan/tls_session.go
```diff
@@ -0,0 +1,87 @@
...
+		return
+	}
+	mostRecent = states[len(states)-1]
+	return
+}
+func (s *sessionCache) Put(sessionKey string, cs *tls.ClientSessionState) {
+	s.cache[sessionKey] = append(s.cache[sessionKey], cs)
+}
+
+// SessionResumeScan tests that host is able to resume sessions accross all addresses.
```
jgrahamc: across

scan/tls_session.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return
+	}
+	ips, err := net.LookupIP(host)
+	if err != nil {
+		return
+	}
+	sessions := newSessionCache()
+	output = sessions
+	config := &tls.Config{ClientSessionCache: sessions, InsecureSkipVerify: true}
```
jgrahamc: Why 
```
InsecureSkipVerify: true
```
? I think that needs documenting.

scan/tls_session.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, ip := range ips {
+		var addr string
+		if ip.To4() != nil {
+			addr = ip.String()
+		} else {
+			addr = "[" + ip.String() + "]"
+		}
+		if len(port) == 0 {
```
jgrahamc: Why not use
```
port == ""
```
instead of
```
len(port) == 0
```

scan/tls_session.go
```diff
@@ -0,0 +1,87 @@
...
+			port = ":443"
+		} else if port[0] != ':' {
+			port = ":" + port
+		}
+		fmt.Println()
+		var conn *tls.Conn
+		conn, err = tls.Dial(Network, addr+port, config)
```
jgrahamc: There's actual a function called
```
net.JoinHostPort
```
http://golang.org/pkg/net/#JoinHostPort

scan/tls_session.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			return
+		}
+		conn.Close()
+		if len(sessions.cache) != 1 {
+			return
```
jgrahamc: Is it OK that the error returned here is nil?

cfssl_scan.go
jacobhaven: Updated with:
```
// regexpLoop iterates through each scan Family and Scanner registered in scan.AllFamilies.
// familyFunc is run on each Family with a name matching the family flag's regexp,
// then scannerFunc is run on each Scanner in that Family that matches the scanner flag's regexp
```

cfssl.go
jacobhaven: Essentially you should probably only use one or the other. Using both takes the intersection. How does this sound?
```
  -family="": regular expression selecting specific scan families to run (defaults to all)
  -scanner="": regular expression selecting specific scanners to run (defaults to all)
```

cfssl.go
jacobhaven: It was intended as "don't run long and/or intensive scans". But, preferably we won't have those, or you can select to not run the entire suite. The `short` argument has been removed, and and replaced with a `verbose` output argument.

cfssl_scan.go
jacobhaven: Most of the other cfssl Mains use `(err error)`, so I'll stick with that, but fix the `return err`/`return nil` instances

scan/connectivity.go
jacobhaven: Good point. Split into two separate errors.

scan/scan_common.go
jacobhaven: Good point, done.

scan/scan_common.go
jacobhaven: Done.

scan/scan_common.go
jacobhaven: True. At first I thought there might be more included in there. I've refactored a large part of this.

scan/tls_handshake.go
jacobhaven: Fair :)

scan/tls_session.go
jacobhaven: I've actually removed this type entirely and used `tls.NewLRUClientSessionCache()` instead

scan/tls_session.go
jacobhaven: I suppose there's no reason not to verify here, so I will.

scan/tls_session.go
jacobhaven: Actually, port should *always* be populated (already defaulted to 443 if empty), so this is completely unnecessary.

jacobhaven:
Rebased on new `cmd` and `cli` structure and cleaned up commit log, added new scan families for secure HTTP headers and (extending the scope of the tool), and scanner to scrape CIDR IP ranges for new Intermediate CAs to update [cfssl_trust](https://github.com/cloudflare/cfssl_trust) for Bundler.

grittygrease:
Unfortunately, some of the go code copied from the standard library does not follow the official style guidelines and fails go vet. Please check the travis output and fix.

scan/https.go
```diff
@@ -0,0 +1,86 @@
...
+		fing := base64.StdEncoding.EncodeToString(digest[:])
+		if strings.Contains(hpkpHeader, fing) {
+			output = fingerprint(fing)
+			grade = Good
+		}
+	}
+	return
+}
+
+// hstsScan tests that the host serves correct HPKP headers.
```
grittygrease: Do you also want to check for expiry time here? Short period HSTS is less valuable than long period HSTS.

jacobhaven: Possibly. I mainly couldn't find an easy way to parse the header values.

grittygrease:
@jacobhaven Can you add tests for these packages? Even a nearly empty test file in each package makes sure that the overall coverage takes this code into account.

scan/scan_common.go
```diff
@@ -0,0 +1,142 @@
...
+const (
+	// Bad describes a host with serious misconfiguration or vulnerability.
+	Bad Grade = iota
+	// Legacy describes a host with non-ideal configuration that maintains support for legacy clients.
+	Legacy
+	// Good describes host performing the expected state-of-the-art.
+	Good
+)
+
+func (g Grade) String() string {
```
lziest: Minor, add doc.

scan/scan_common.go
```diff
@@ -0,0 +1,142 @@
...
+func (s *Scanner) Scan(host string) (Grade, Output, error) {
+	grade, output, err := s.scan(host)
+	if err != nil {
+		log.Infof("%s: %s", s.Name, err)
+		return grade, output, err
+	}
+	History.store(s.Name, host, output)
+	return grade, output, err
+}
+func (s *Scanner) String() string {
```
lziest: Minor, add doc and a blank line

scan/scan_common.go
```diff
@@ -0,0 +1,142 @@
...
+type Family struct {
+	// Name is a short name for the Family.
+	Name string
+	// Description gives a short description of the scans performed on the host.
+	Description string
+	// Scanners is a list of scanners that are to be run in sequence.
+	Scanners []*Scanner
+}
+
+func (f *Family) String() string {
```
lziest: Minor, add doc.

scan/tls_handshake.go
```diff
@@ -0,0 +1,111 @@
...
+	var cipherID uint16
+	for cipherIndex, cipherID = range ciphers {
+		if serverCipher == cipherID {
+			return
+		}
+	}
+	err = fmt.Errorf("server negotiated ciphersuite we didn't send: %s", tls.CipherSuites[serverCipher])
+	return
+}
+func allCiphersIDs() []uint16 {
```
lziest: Add blank lines for style consistency.

cli/scan/scan.go
```diff
@@ -0,0 +1,121 @@
...
+func indentPrintln(block string, indentLevel int) {
+	w := tabwriter.NewWriter(os.Stdout, 0, 4, 4, ' ', 0)
+	scanner := bufio.NewScanner(strings.NewReader(block))
+	for scanner.Scan() {
+		fmt.Fprintf(w, "%s%s\n", strings.Repeat("\t", indentLevel), scanner.Text())
+	}
+	w.Flush()
+}
+func scanMain(args []string, c cli.Config) (err error) {
+	familyRegexp := regexp.MustCompile(c.Family)
```
lziest: Minor issue: do we want to panic here if the input is not valid regex? Just use Compile and throw error here?

cli/cli.go
```diff
@@ -94,6 +97,9 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.StringVar(&c.IP, "ip", "", "remote server ip")
 	f.StringVar(&c.Remote, "remote", "", "remote CFSSL server")
 	f.StringVar(&c.Label, "label", "", "key label to use in remote CFSSL server")
+	f.BoolVar(&c.List, "list", false, "list possible scanners")
+	f.StringVar(&c.Family, "family", "", "family regular expression")
```
lziest: I prefer 'scanner family regular expression'

lziest:
Other than those minor issues, LGTM

grittygrease:
Once Zi's comments are addressed and the _test.go files are added, I think this is ready to merge.

cli/scan/scan.go
kisom: I think that we should use Compile,m and return an error instead of panicking.

scan/connectivity.go
```diff
@@ -0,0 +1,80 @@
...
+func (addrs lookupAddrs) Describe() string {
+	return "List of host's addresses returned by DNS lookup"
+}
+
+// dnsLookupScan tests that DNS resolution of the host returns at least one address
+func dnsLookupScan(host string) (grade Grade, output Output, err error) {
+	host, _, err = net.SplitHostPort(host)
+	if err != nil {
+		return
+	}
```
kisom: It's a style nit, but I find the code is more readable if you put a blank line here (and L53).

scan/https.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+			return conn, err
+		},
+	}
+	client := http.Client{Transport: tr}
+	resp, err := client.Get("https://" + host)
+	if err != nil {
+		return
+	}
+	hpkpHeader := resp.Header.Get("Public-Key-Pins")
```
kisom: Style nit: I think it's more readable if you put a newline between the '}' and this line.

scan/tls_handshake.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return
+	}
+	conn := tls.Client(tcpConn, &tls.Config{MinVersion: vers, MaxVersion: vers, ServerName: host, CipherSuites: ciphers, SessionTicketsDisabled: true, ClientSessionCache: nil})
+	serverCipher, serverVersion, err := conn.SayHello()
+	conn.Close()
+	if err != nil {
+		return
+	}
+	if serverVersion != vers {
```
kisom: I'd add a blank line here.

scan/tls_handshake.go
```diff
@@ -0,0 +1,111 @@
...
+		err = fmt.Errorf("server negotiated protocol version we didn't send: %s", tls.Versions[serverVersion])
+		return
+	}
+	var cipherID uint16
```
kisom: and probably here, too.

scan/tls_handshake.go
```diff
@@ -0,0 +1,111 @@
...
+		ciphers = append(ciphers, cipherID)
+	}
+	return ciphers
+}
+
+type cipherVersions struct {
+	cipherID uint16
+	versions []uint16
+}
+type cipherVersionList []cipherVersions
```
kisom: Should add a blank line.

scan/tls_handshake.go
```diff
@@ -0,0 +1,111 @@
...
+	for i, c := range cvList {
+		versStrings := make([]string, len(c.versions))
+		for j, vers := range c.versions {
+			versStrings[j] = tls.Versions[vers]
+		}
+		cvStrings[i] = fmt.Sprintf("%s\t%s", tls.CipherSuites[c.cipherID], strings.Join(versStrings, ", "))
+	}
+	return strings.Join(cvStrings, "\n")
+}
+func (cvList cipherVersionList) Describe() string {
```
kisom: Blank line above this.

scan/tls_session.go
```diff
@@ -0,0 +1,55 @@
...
+		return
+	}
+	config := &tls.Config{ClientSessionCache: tls.NewLRUClientSessionCache(1), ServerName: hostname}
+	var conn *tls.Conn
+	conn, err = tls.Dial(Network, host, config)
+	if err != nil {
+		return
+	}
+	conn.Close()
+	for _, ip := range ips {
```
kisom: I'd add a blank line about this one (after the `conn.Close()`).

kisom:
Only nits I had were stylistic; other than those, LGTM.

kisom:
Also, @jacobhaven you'll need to resolve the merge conflicts before I can merge this.

grittygrease:
@jacobhaven Can you address Kyle's comments so we can merge and move on to adding an API?

cli/config.go
```diff
@@ -72,6 +75,9 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.IntVar(&c.Reason, "reason", 0, "Reason code for revocation")
 	f.StringVar(&c.RevokedAt, "revoked-at", "now", "Date of revocation (YYYY-MM-DD)")
 	f.Int64Var(&c.Interval, "interval", int64(4*helpers.OneDay), "Interval between OCSP updates, in seconds (default: 4 days)")
+	f.BoolVar(&c.List, "list", false, "list possible scanners")
+	f.StringVar(&c.Family, "family", "", "family regular expression")
```
lziest: Can we use "scanner family regular expression" for the flag usage string?

PR #77, opened 2015-01-29
Support PKCS#11 for certificate signing

bifurcation:
This PR adds support for signing certificates with a private key stored on a PKCS#11 token.  

In terms of code, the main impacts here are:
1. Adding a "Pkcs11Key" type that implements crypto.Signer with a token-backed key
2. Tweaking the signer.NewSigner interface to allow either file-backed or token-backed keys
3. Wiring all this through to the CLI

This extension has a couple dependency impacts:
* [PKCS#11 interface library](https://github.com/miekg/pkcs11)
* Latest version of "crypto/x509", which allows the use of [crypto.Signer with CreateCertificate](https://github.com/golang/go/commit/fe40cdd756763982db6815ee87318f418218fcf5)

I have tested this using the PKCS#11 capabilities on a [Yubikey Neo-N](https://www.yubico.com/products/yubikey-hardware/yubikey-neo/) and with [SoftHSM](https://www.opendnssec.org/softhsm/).

Obviously, the references to "github.com/bifurcation/cfssl" will need to be changed before merging.

### DIFF
--- a/api/api_generator.go
+++ b/api/api_generator.go
@@ -10,12 +10,12 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/bifurcation/cfssl/signer"
 	"github.com/cloudflare/cfssl/api/client"
 	"github.com/cloudflare/cfssl/config"
 	"github.com/cloudflare/cfssl/csr"
 	"github.com/cloudflare/cfssl/errors"
 	"github.com/cloudflare/cfssl/log"
-	"github.com/cloudflare/cfssl/signer"
 )
 
 // Validator is a type of function that contains the logic for validating
@@ -155,7 +155,7 @@ func NewCertGeneratorHandler(validator Validator, caFile, caKeyFile, remote stri
 		}
 	}
 
-	if cg.signer, err = signer.NewSigner(caFile, caKeyFile, cfg); err != nil {
+	if cg.signer, err = signer.NewSignerFromFile(caFile, caKeyFile, cfg); err != nil {
 		if remote == "" {
 			return nil, err
 		}
--- a/api/api_signer.go
+++ b/api/api_signer.go
@@ -5,12 +5,12 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/bifurcation/cfssl/signer"
 	"github.com/cloudflare/cfssl/api/client"
 	"github.com/cloudflare/cfssl/auth"
 	"github.com/cloudflare/cfssl/config"
 	"github.com/cloudflare/cfssl/errors"
 	"github.com/cloudflare/cfssl/log"
-	"github.com/cloudflare/cfssl/signer"
 )
 
 // A SignHandler accepts requests with a hostname and certficate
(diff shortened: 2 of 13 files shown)


kisom:
Thanks for this PR! It's a big one, so it'll take a little bit to code review. We're also in the middle of refactoring our signer code, which will change the interface a bit. Once that's in, I can help get this integrated. I'd also propose that the PKCS11 code go in separate packages and have that functionality enabled with a build flag.

bifurcation:
Any idea what the ETA is on the signer refactor?  Putting this in a separate module sounds very sensible.  I have no idea how to do build flags in golang, though, so I will require some advising on that point.

kisom:
It's being worked on now, and should be within a week. I've got a pretty good idea in my head how to take get this working with those changes, and I can help out with the build flags. If you're curious, Dave Cheney has some [useful](http://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool) [writeups](http://dave.cheney.net/2014/09/28/using-build-to-switch-between-debug-and-release).

kisom:
@bifurcation I think we're ready to start integrating this in. I can take what you have and build it into an appropriate module (and credit you with the commit), or I can help guide you with the changes that need to be made.

bifurcation:
If you're in the groove, go for it!  I don't even really care much about
the credit :)  I'm working on getting OCSP working.

bifurcation:
Looking at the refactor, it looks like the right thing to do is:
* Make a top-level pkcs11 module
* Hide it behind a build flag
* Add pkcs11 config flags to the cli.go
* Have the cli/sign and cli/ocspsign (TODO) modules
  * import pkcs11 and use it if enabled by the build flag and instructed by CLI flags
  * throw an error if the CLI flags are set by the build flag wasn't 

Agreed?

kisom:
That's in line with what I'm planning to do, with these minor variations:

- add a `signer/pkcs11` package with build flags
- add a `cmd/cfssl-pkcs11` package that will provide a PKCS #11 CLI
(this can be specifically tailored to accepting PKCS #11 keys).

How's that sound?

bifurcation:
Not quite right.

ISTM that the pkcs11 module should be at the top level, since PKCS#11 keys might be used for functions other than signing certificates.  (In particular, signing OCSP responses.)  There would be two files in there:
* // +build pkcs11 -- defines actual pkcs11 implementation, pkcs11.Enabled = true
* // +build !pkcs11 -- defines stub pkcs11 implementation (throwing error if invoked), pkcs11.Enabled = false

Then the CLI module can just call pkcs11, and if the module's not actually there, they'll get an error.  No need to fork the CLI.

I assume you're trying to minimize the impact on the CLI.  Unfortunately, if you fork that, you also have to fork cli/cli.go, cli/sign/sign.go, and the eventual cli/ocspsign/ocspsign.go, since they all have to have different behavior for PKCS#11.  That doesn't seem worth it, given that the impact on those files should be relatively small (see the mods to cfssl_sign.go in the current patch set).

Happy to whiteboard this next week if you want.

kisom:
This has been shifted to #95.

PR #82, opened 2015-02-01
Remove mercurial dependence

dr107:
Edit the import of the ocsp to use the new, git location, thus
ending cfssl's dependence on mercurial


grittygrease:
Looks good.

PR #83, opened 2015-02-01
Fix lint errors

dr107:
Fix various golint errors caused by variable names with underscores
rather than camlCase


kisom:
LGTM. I love seeing golint used on the code base more.

PR #86, opened 2015-02-02
Remove dependency on client from signer

grittygrease:
This is a continuation of the previous refactor. The main reason for it is to remove the dependencies on the client library from the signer and config packages.

- add a signer/universal package
- move the local and remote signers to their own packages
- fix dependency on client/api from config


lziest:
LGTM

PR #87, opened 2015-02-04
Extract parts of local signer code into signer

kisom:
There are certain helper functions that will be useful for a wide range of potential signers. These components have been extracted from the local signer and placed into the signer package.


lziest:
LGTM

PR #89, opened 2015-02-05
Use crypto.Signer for private keys.

kisom:
This PR changes the private keys used in the signer package from `interface{}` to `crypto.Signer`. Once the changes to `crypto/x509` make it into mainline Go, this will permit a more fleixble signature system.

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -4,6 +4,7 @@ package helpers
 
 import (
 	"bytes"
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/rsa"
 	"crypto/x509"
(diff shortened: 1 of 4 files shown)


lziest:
LGTM

signer/local/local_test.go
```diff
@@ -285,7 +286,7 @@ func TestCAIssuing(t *testing.T) {
 			}
 			keyBytes, _ := ioutil.ReadFile(interKeys[j])
 			interKey, _ := helpers.ParsePrivateKeyPEM(keyBytes)
-			interSigner := &Signer{interCert, interKey, CAPolicy, signer.DefaultSigAlgo(interKey)}
+			interSigner := &Signer{interCert, interKey.(crypto.Signer), CAPolicy, signer.DefaultSigAlgo(interKey)}
```
grittygrease: Is this necessary now that ParsePrivateKeyPEM uses crypto.Signer?

grittygrease:
There are a lot of changes missing here. For example the DefaultSigAlgo function, and everywhere else in the code an explicit rsa.PrivateKey is used.

PR #90, opened 2015-02-06
Fix the CLI after refactor.

grittygrease:
Re-enable the ability to use flags in the CLI.


kisom:
There's no usage information in cfssl anymore.

```
(2) <straka:kyle> $ ./cfssl                  
No command is given.
Usage of ./cfssl:
  -loglevel=1: Log level
```

It would be useful to have a list of the commands that are available.

kisom:
LGTM.

PR #91, opened 2015-02-06
crypto.Signer

bifurcation:
Currently, private keys are represented as type interface{}.  It is better clearer to use crypto.Signer.  That's the direction that crypto/x509 is headed, and it's compatible with the current version.


kisom:
Good catch, but there's a PR for this in #89 .

PR #95, opened 2015-02-11
Add PKCS #11 support to CFSSL.

kisom:
This commit is based on a previous pull request by Richard Barnes (@bifurcation).

### DIFF
--- a/signer/pkcs11/doc.go
+++ b/signer/pkcs11/doc.go
@@ -1,0 +1,4 @@
+// Package pkcs11 implements support for PKCS #11 signers. If the
+// package has not been built with the `pkcs11` tag, the `New`
+// function will be a stub.
+package pkcs11
--- a/signer/pkcs11/pkcs11_stub.go
+++ b/signer/pkcs11/pkcs11_stub.go
@@ -1,0 +1,27 @@
+// +build !pkcs11
+
+package pkcs11
+
+import (
+	"github.com/cloudflare/cfssl/config"
+	"github.com/cloudflare/cfssl/errors"
+	"github.com/cloudflare/cfssl/signer"
+)
+
+// Config contains configuration information required to use a PKCS
+// #11 key.
+type Config struct {
+	Module string
+	Token  string
+	PIN    string
+	Label  string
+}
+
+// New always returns an error. If PKCS #11 support is needed, the
+// program should be built with the `pkcs11` build tag.
+func New(caCertFile string, policy *config.Signing, cfg *Config) (signer.Signer, error) {
+	return nil, errors.New(errors.PrivateKeyError, errors.Unknown)
+}
+
+// Enabled is set to true if PKCS #11 support is present.
+const Enabled = false
--- a/api/api_generator.go
+++ b/api/api_generator.go
@@ -154,8 +154,10 @@ func NewCertGeneratorHandler(validator Validator, caFile, caKeyFile string, poli
 	}
 
 	root := universal.Root{
-		CertFile: caFile,
-		KeyFile:  caKeyFile,
+		Config: map[string]string{
+			"ca-file":     caFile,
+			"ca-key-file": caKeyFile,
+		},
 	}
 	if cg.signer, err = universal.NewSigner(root, policy); err != nil {
 		log.Errorf("setting up signer failed: %v", err)
--- a/api/api_signer.go
+++ b/api/api_signer.go
@@ -30,8 +30,10 @@ func NewSignHandler(caFile, caKeyFile string, policy *config.Signing) (http.Hand
 	s := new(SignHandler)
 
 	root := universal.Root{
-		CertFile: caFile,
-		KeyFile:  caKeyFile,
+		Config: map[string]string{
+			"cert-file": caFile,
+			"key-file":  caKeyFile,
+		},
 	}
 	if s.signer, err = universal.NewSigner(root, policy); err != nil {
 		log.Errorf("setting up signer failed: %v", err)
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -42,59 +42,8 @@ type Command struct {
 	Main func(args []string, c Config) error
 }
 
-// Config is a type to hold flag values used by cfssl commands.
-type Config struct {
-	Hostname          string
-	CertFile          string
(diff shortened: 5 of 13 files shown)


grittygrease:
This is really cool.

signer/pkcs11/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package pkcs11 implements support for PKCS #11 signers. If the
+// package has not been built with the `pkcs11` tag, the `New`
+// function will be a stub.
+package pkcs11
```
bifurcation: This package might want to be at the top level, rather than under signer.  We're going to want to use PKCS#11 for OCSP signing as well.   So I guess the choice is whether you want the dependencies to go:
1. ocsp -> signer/pkcs11
2. ocsp -> pkcs11

signer/pkcs11/pkcs11_stub.go
```diff
@@ -0,0 +1,24 @@
...
+	Token  string
+	PIN    string
+	Label  string
+}
+
+// New always returns an error. If PKCS #11 support is needed, the
+// program should be built with the `pkcs11` build tag.
+func New(caCertFile string, policy *config.Signing, cfg *Config) (signer.Signer, error) {
+	return nil, errors.New(errors.PrivateKeyError, errors.Unknown)
+}
```
bifurcation: It seems like if you have this level of adaptation, then the split in the CLI code (and associated duplication of non-PKCS#11 code) is unnecessary.  The only function it serves is to hide the PKCS#11 options from the CLI, which could be done more cleanly with something like:

In pkcs11_stub.go:
```
const Enabled = false
```

In pkcs11.go:
```
const Enabled = true
```

In cli/config.go :: registerFlags:
```
if pkcs11.Enabled { /* register p11 flags */ }
```

cli/config_pkcs11.go
```diff
@@ -0,0 +1,83 @@
...
+	Profile            string
+	IsCA               bool
+	IntDir             string
+	Flavor             string
+	Metadata           string
+	Domain             string
+	IP                 string
+	Remote             string
+	Label              string
+	Module, Token, PIN string
```
bifurcation: Why are just these three on the same line?

cli/config_pkcs11.go
```diff
@@ -0,0 +1,83 @@
...
+	PKCS11Label        string
```
bifurcation: The other usage of "label" actually seems consistent with the PKCS#11 semantic, in the sense that the HSM is a remote signing server.  I would be inclined just to re-use "label".

signer/pkcs11/pkcs11key/pkcs11key.go
```diff
@@ -0,0 +1,235 @@
...
+		return
+	}
+	ps.publicKey = rsa.PublicKey{n, e}
+
+	return
+}
+
+// Destroy tears down a PKCS11Key.
+//
+// This method must be called before the PKCS11Signer is GC'ed, in
```
bifurcation: s/PKCS11Signer/PKCS11Key/g

signer/pkcs11/pkcs11key/pkcs11key.go
```diff
@@ -0,0 +1,235 @@
...
+	return &ps.publicKey
+}
+
+// Sign performs a signature using the PKCS #11 key.
+func (ps *PKCS11Key) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) (signature []byte, err error) {
+	// Verify that the length of the hash is as expected
+	hash := opts.HashFunc()
+	hashLen := hash.Size()
+	if len(msg) != hashLen {
+		err = errors.New("input size does not match hash function output size")
```
bifurcation: Should these errors switch to using cferr?  Actually probably not, since it might be nice to be able to re-use this file outside of CFSSL.

signer/universal/universal_signers.go
```diff
@@ -0,0 +1,9 @@
+// +build !pkcs11
+
+package universal
+
+// Without any build flags, the only local signer that should be
+// activated is the file-backed local signer.
+var localSignerList = []localSignerCheck{
+	fileBackedSigner,
+}
```
bifurcation: Again, this split seems like overkill.  Instead:
1. Get rid of localSignerCheck
2. Push any checking logic in fileBackedSigner / pkcs11Signer into the local.NewSigner()FromFile / pkcs11.New().
3. Replace calls to localSignerCheck methods with just the constructors.

That way, when the build flags are off, the PKCS#11 stuff will just fail.

bifurcation:
Overall, this looks pretty OK.  

Before this gets merged, though, there needs to be some refactoring to minimize duplicated code between the pkcs11 / !pkcs11 branches.  Basically, the only place that split should be salient at build time should be in the pkcs11 module itself -- everything else should be able to ask whether PKCS#11 is enabled.

The other note is that we're going to want to use PCKS#11 for OCSP signing as well.  So we should consider whether we want the PKCS11Key struct to live where it does now, or somewhere else.  (This is really just a question of what we want the dependency graph to look like.)

cli/config_pkcs11.go
kisom: I thought about reusing it at first. However, a command line option can only go to one variable; if we re-use the "-label" flag, it's inconsistent with the other three "pkcs11-" prefixed options. This makes it less obvious that the flag also serves as a PKCS11 label.

cli/config_pkcs11.go
kisom: That's just me using my style and not following the rest of the file's style. This is fixed.

signer/pkcs11/pkcs11_stub.go
kisom: I ended up using the Enabled approach to a lot of this, and created a new error reason "Unavailable"; attempting to use PKCS11 without support built in returns this code.

signer/pkcs11/pkcs11key/pkcs11key.go
kisom: I left it as standard errors, and let the wrapping occur in the `pkcs11` package.

kisom:
The comments should be all addressed now.

PR #105, opened 2015-02-17
Switch from sign to authsign when appropriate.

bifurcation:
Currently, there is no way to trigger authentication from `cfssl serve`.  It seems like the right answer in the long run is to simplify "Sign" and "AuthSign" so that they can authenticate or not depending on the profile being used.  But this PR at least lets you create an authenticated-only server (which is going to be a requirement for LE).


grittygrease:
I'm fine with this quick hack to get AuthSign working and I agree that the creation of the endpoints needs to be dependent on the profile in the long term.

grittygrease:
Actually, NewAuthSignHandler is not going to return an error in the right circumstances to make this change valid. In the case where all profiles have a provider, but for which all keys are remote, this will cause the /sign endpoint to be disabled but the /authsign endpoint to be enabled.

To have a cfssl server with /authsign enabled and /sign disabled, you can make this logic contingent on the -remote parameter being unset. If -remote is set, we should try to enable both.

The expected behavior is outlined here:
https://github.com/cloudflare/cfssl/issues/107

bifurcation:
How about the following approach:

* SignHandler will only respond if `profile.Provider != nil`
* AuthSignHandler will respond only if profile.Provider != nil`
* New[Auth]SignHandler will error if there are no profiles it will respond for
* `cfssl serve` will create both

api/api_signer.go
```diff
@@ -26,30 +26,51 @@ type SignHandler struct {
...
+func NewSignHandlerFromSigner(signer signer.Signer) (h HTTPHandler, err error) {
+	policy := signer.Policy()
+	if policy == nil {
+		err = errors.New(errors.PolicyError, errors.InvalidPolicy)
+		return
+	}
+
+	// Sign will only respond for profiles that have no auth provider.
+	// So if all of the profiles require authentication, we return an error.
+	haveUnauth := (policy.Default.Provider == nil)
```
kisom: Maybe we should do this before calling NewSignHandler (and not necessarily error).

kisom:
I'm not sure we should fail if there's no unauthenticated endpoints, but I'm fine with the changes for now to get us moved forward.

PR #106, opened 2015-02-17
When no metadata, bundle ubiquity should be false. Add some extra ubiqui...

lziest:
...ty tests for better coverage.


grittygrease:
LGTM

PR #108, opened 2015-02-20
Test local signer

dr107:
Bring test coverage for local signer from ~60% to 92%

There's a few lines in local.go that I've been struggling to hit. I'd appreciate some tips on how I might hit them.

coverage-0
coverage-1
coverage-2


signer/local/local_test.go
```diff
@@ -107,6 +108,161 @@ func TestNewSignerFromFileNoUsageInPolicy(t *testing.T) {
...
+		t.Fatal("srsly?")
+	}
+	cert, err := helpers.ParseCertificatePEM(pem)
+	if err != nil {
+		t.Fatal("srsly?")
+	}
+
+	badcert := *cert
+	badcert.PublicKey = nil
+	profl := config.SigningProfile{Usage: []string{"FUCK", "YOU"}}
```
grittygrease: Looks like this was frustrating to create :) Let's try "Certificates" "Rule"

signer/local/local_test.go
```diff
@@ -107,6 +108,161 @@ func TestNewSignerFromFileNoUsageInPolicy(t *testing.T) {
...
+}
+
+func TestSign(t *testing.T) {
+	testSign(t)
+	s, err := NewSignerFromFile("testdata/ca.pem", "testdata/ca_key.pem", nil)
+	if err != nil {
+		t.Fatal("Failed to produce signer")
+	}
+
+	//pem, _ := ioutil.ReadFile("../helpers/testdata/cert.pem")
```
grittygrease: Please do not commit commented-out code.

grittygrease:
Hey Dan, looks like there's something that we need to fix with the coveralls configuration. I'll take a look.

As for the missing lines:
- The code for sign for which s.ca is nil (causing initroot=1) and for which the incoming DER is malformed can't be reached

- add a test case for Sign with req a valid base64 block, but for which the data is a malformed certificate
the csr package should be able to help you construct these:
- add a test case for Sign with an empty subject, but for which the certificate has an IP as the hostname
- add a test case for Sign with a subject and a domain name as one of the Subject.Hosts

signer/local/local_test.go
```diff
@@ -107,6 +108,161 @@ func TestNewSignerFromFileNoUsageInPolicy(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	validReq := signer.SignRequest{Hostname: testHostName, Request: string(pem)}
+	_, err = s.Sign(validReq)
+	if err == nil {
+		t.Fatal("A bad case failed to raise an error")
+	}
+
+	validReq = signer.SignRequest{Hostname: "74.67.211.31", Request: string(pem)}
```
dr107: I actually tried the hostname as IP thing here, but it still didn't touch that line.

grittygrease: This is the signer.SignRequest with an IP. Try creating a CSR with an IP as a hostname, here you're using the testdata/cert.pem.

dr107: Got it. Covered the line.

PR #120, opened 2015-03-04
Simplify SigningProfile with remote signer.

lziest:
Just remove expiry and usage requirement for remote signing profile. Add a bunch tests.


kisom:
LGTM.

PR #125, opened 2015-03-08
Additional configuration parameters for CA policies

bifurcation:
For Let's Encrypt, we require a few additional policy knobs to generate certificates with the profile we want.

* Specifying certificate policies (as OIDs)
* Specifying the OCSP No Check extension
* Specifying the amount of back-dating applied to a certificate

This PR adds those controls to config.SigningProfile, and adapts signer.FillTemplate() to put them into action.

As of this commit, the patch is untested, but builds and passes `go vet` and `golint`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,9 +3,13 @@ package config
 
 import (
 	"crypto/x509"
+	"encoding/asn1"
 	"encoding/json"
 	"errors"
 	"io/ioutil"
+	"regexp"
+	"strconv"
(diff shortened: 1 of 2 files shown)


bifurcation:
OK, I have now verified that these options work, and generate certificates with the expected context.

kisom:
LGTM.

*Note*: Build failure is due to well-known coveralls failure for forked repo PRs.

PR #130, opened 2015-03-16
Fix remote signing bugs

bifurcation:
Recent changes around hostname handling broke remote signing by creating a discontinuity in how the API client and server expressed lists of hostnames.

* Nil pointer dereference on subject copy
* Requests without `hostnames` rejected even when `hosts` was present

This PR fixes these two bugs.


kisom:
LGTM.

Build failure is due to well-known coveralls issue with forked repos. All other tests pass.

PR #131, opened 2015-03-16
Update multiroot CA with recent changes to structs.

kisom:


cmd/multirootca/api.go
```diff
@@ -164,11 +164,11 @@ func dispatchRequest(w http.ResponseWriter, req *http.Request) {
 	}
 
 	signReq := signer.SignRequest{
```
lziest: Actually, can we just send sigRequest over?

lziest:
LGTM

PR #132, opened 2015-03-17
Get missing info from CSR fields

dr107:
If there is info missing in the API request, supplement it with
information pulled out of the CSR pem.

TBD: add unit tests. I'm putting this PR up early to give the CFSSL devs a chance to get started on code review.


lziest:
sorry, somehow I ends up commenting on the commit. And please add some test cases.

dr107:
How would I generate a CSR with different kinds of Subjects? The openSSL process seems to only allow one entry for each kind of name, but the CFSSL/Go code allows for multiple. This is what's making writing tests so difficult.

lziest:
Let's hold on this. I do feel this PR is not what we want.

lziest:
Now I get a better understanding on this now. I would like the work to be done the other way around. Let's modify the way the subject is overridden by req.Subject at signer.ParseCertificateRequest(). Right now if req.Subject is not nil, CSR subject is completely overwritten. Instead, selectively overwritten fields that are present in the req.Subject. If subject.CN != nil, overwrite common name, etc. Does that make sense?

dr107:
Got it. I'll move the work to signer.ParseCertificateRequest

signer/signer.go
```diff
@@ -65,6 +65,33 @@ func (s *Subject) Name() pkix.Name {
...
+func replaceSliceIfEmpty(replaced, newContents *[]string) {
+	if len(*replaced) == 0 {
+		*replaced = *newContents
+	}
+}
+
+// populateSubjectFromCSR has functionality similar to Name, except
+// it fills the fields of the resulting pkix.Name with req's if the
+// subject's corresponding fields are empty
+func (s *Subject) populateSubjectFromCSR(req pkix.Name) pkix.Name {
```
lziest: I think this can be exported.

signer/signer_test.go
```diff
@@ -1 +1,89 @@
...
+
+import (
+	"crypto/x509/pkix"
+	"reflect"
+	"testing"
+
+	"github.com/cloudflare/cfssl/csr"
+)
+
+func TestReplaceSliceIfEmpty(t *testing.T) {
```
lziest: Thanks. But I think it is not needed to test a helper function explicitly.

signer/signer.go
```diff
@@ -65,6 +65,33 @@ func (s *Subject) Name() pkix.Name {
...
+func replaceSliceIfEmpty(replaced, newContents *[]string) {
+	if len(*replaced) == 0 {
+		*replaced = *newContents
+	}
+}
+
+// populateSubjectFromCSR has functionality similar to Name, except
+// it fills the fields of the resulting pkix.Name with req's if the
+// subject's corresponding fields are empty
+func (s *Subject) PopulateSubjectFromCSR(req pkix.Name) pkix.Name {
```
lziest: golint complaint: this method should be declared as a independent helper func.

signer/signer_test.go
```diff
@@ -1 +1,89 @@
...
+
+	replaceSliceIfEmpty(nonEmpty, contents)
+
+	if reflect.DeepEqual(*nonEmpty, *contents) {
+		t.Fatal("Replaced non-empty slice")
+	}
+
+}
+
+func TestPopulateSubjectFromCSR(t *testing.T) {
```
lziest: And consider duplicating this test logic at signer/local/local_test.go TestOverrideSubject(). Let's use tests to specify the expected behavior of signers, otherwise bugs may be introduced later without being noticed.

lziest:
try to sign a csr with customized subject, and inspect the subject of the outcome certificate. You can use an existing csr in the testdata directory.

dr107:
And should I do that instead of, or in addition to the tests I already had.

lziest:
in addition to :)

dr107:
Ok. Sorry to keep bugging you, but is there any way to create a signer from the signer package? I keep getting circular dependency errors.

lziest:
you can create a signer by signer/local package. Are you trying to test against API? The API tests need a bit effort to avoid circular dependencies.

dr107:
Yeah. I try to use local.NewSignerFromFile and it obviously creates a circular dependency. How else can I create a signer?

lziest:
Can you simply piggyback tests here?https://github.com/cloudflare/cfssl/blob/master/signer/local/local_test.go#L479

dr107:
Will do.

grittygrease:
Dan, I think your code makes more sense living in the signer/local package.
That should resolve your dependency issues.

dr107:
My code or just these tests?

grittygrease:
Your code should be in local signer, and the tests. The remote signer does
not introspect certificates or CSRs, so this is not something that needs to
be used by both types of signers.

Nick

dr107:
Ok. I'll have the code and tests in by EOD tomorrow.

dr107:
Is this what you guys were talking about? This issue has been kicked around all over the place so I want to make sure.

lziest:
Let's make a bigger and more complete change. Step back, you can see signer.ParseCertificateRequest is doing 3 things: parse and verify CSR, override subject and override hosts. And we are adding a new SubjectOverride function for the new logic. It's very patchy.
Can you reduce ParseCertificateRequest to do CSR parsing only. And move subject override and hosts override to individual functions in signer/local. This way, the override logic is confined to signer/local, and it is easier to test.

dr107:
That makes a lot of sense.

signer/signer.go
```diff
@@ -118,7 +117,7 @@ func DefaultSigAlgo(priv crypto.Signer) x509.SignatureAlgorithm {
 // builds a certificate template from it. If not nil, the subject
 // information from subject will be used in place of the information in
 // the CSR. The same is true for the list of hosts.
-func ParseCertificateRequest(s Signer, csrBytes []byte, sub *Subject, hosts []string) (template *x509.Certificate, err error) {
+func ParseCertificateRequest(s Signer, csrBytes []byte, sub *Subject) (template *x509.Certificate, err error) {
```
lziest: Looks like we can also remove sub.

signer/local/local_test.go
```diff
@@ -506,10 +569,31 @@ func TestOverrideSubject(t *testing.T) {
 		t.Fatalf("%v", err)
 	}
 
+	block, _ := pem.Decode(csrPEM)
+	template, err := signer.ParseCertificateRequest(s, block.Bytes, request.Subject)
```
lziest: how about use csr, err := x509.ParseCertificateRequest(block.Bytes) directly?

lziest:
Other than those minor issues, LGTM. @kisom please take a look.

kisom:
There's one style nit, but other than that, LGTM.

signer/local/local_test.go
```diff
@@ -476,8 +478,67 @@ func TestCAIssuing(t *testing.T) {
...
+	noOU.Names[0].OU = ""
+	name = PopulateSubjectFromCSR(&noOU, fullName)
+	if !reflect.DeepEqual(name.OrganizationalUnit, fullName.OrganizationalUnit) {
+		t.Fatal("Failed to replace empty organizational unit")
+	}
+
+}
 func TestOverrideSubject(t *testing.T) {
-	csrPEM, err := ioutil.ReadFile(testCSR)
+	csrPEM, err := ioutil.ReadFile("testdata/test.csr")
```
lziest: just a nit: can you move it to the const section and give it a name, e.g. fullsubjectCSR?

PR #133, opened 2015-03-18
Disabling authentication for 'info'

bifurcation:
`remote.Certificate()` is broken right now when you use an authentication provider.  The API client tries to send an authenticated request to `/api/v1/cfssl/authinfo`, for which there's no corresponding API server.  This PR disables authentication for info requests (which aren't sensitive anyway), until such time as there's an API server to respond to authenticated requests.


lziest:
LGTM

PR #134, opened 2015-03-19
bug fix in certificate signing.

lziest:
- when signing a certificiate, hosts overridden shoulde be decoupled
  from subject overridden.


kisom:
LGTM.

Issue #117, opened 2015-02-27
Add support for creating dummy certificates, servers, etc. when testing


hwh33:
Currently, all testing seems to be done with static certificate chains which were created for that test from the command line.  It would be more streamlined to be able to create your own instances in the test code itself.  Issue #96 references a similar problem and suggests the ability to create a test server.  Ideally, one should be able to create a dummy version of anything they might need in a test.  If you need something besides certificate chains and servers, please say so in a comment.

jacobhaven:
The ability to cleanly spawn test TLS servers and clients with arbitrary configs will certainly be useful, both for Bundler (#96) and other use cases. I look forward to the PR!

Issue #96, opened 2015-02-11
Move away from testing against live sites in bundler/bundle_from_remote_test.go


lziest:
The initial batch of bundle_from_remote tests relies on "correct" TLS handshake behaviors of real websites to succeed. The SNI tests has been broken for a while because the SNI test site has been down. We should have setup a test TLS web server locally and test against it.

kisom:
I definitely agree with this, and I'd love to see this in.

ZenGround0:
I am getting started on this!

PR #141, opened 2015-03-25
Harry/helpertestfunctions

hwh33:
This is to address Issue #96 and Issue #117.

Adds a package helpers/testsuite which includes functions for generation of certificates and certificate chains in the code (so no more need for pre-made test data).  Also adds the capability to start up a cfssl server running on your local machine.  

These functions all use the CLI in one way or another, so any updates to the way these things are done with the command-line tools will carry-through to this package.

Test coverage for this package is currently at 74%.  All of the uncovered code is very easy to hit, so I will work on bringing it up to 100%.


helpers/testsuite/testing_helpers_test.go
```diff
@@ -0,0 +1,454 @@
...
+		t.Log("The following fields were unequal:")
+		for _, field := range unequalFields {
+			t.Log(field)
+		}
+		t.Fatal("Certificates unequal.")
+	}
+
+}
+
+/* Compare two x509 certificate chains. We only compare relevant data to
```
kisom: A style nit: we prefer the line commentary style over the C block-comment style.

kisom:
I think the C block-style commentary in `helpers/testsuite/testing_helpers_test.go` should be changed to the C++ line-style commentary (e.g. each line starting with `//`). Other than that, LGTM.

hwh33:
@kisom Ah okay, I will change that then.  Thanks!

lziest:
please fix build failure. Note it is not a coverage faiure.

grittygrease:
Here are two errors that are reported:
api/bundle/bundle.go:55: not enough arguments in call to h.bundler.BundleFromRemote
cli/bundle/bundle.go:78: not enough arguments in call to b.BundleFromRemote

Once these are fixed, please squash and rebase.

grittygrease:
Tests are passing. Please squash and rebase.

PR #145, opened 2015-03-30
fix coverall env setup in .travis.yml

lziest:
this is an attempt to fix the coverall failures for forks.


lziest:
well, it doesn't work.

PR #149, opened 2015-04-01
cfssl scan: Initial API endpoint

jacobhaven:
New API endpoint for cfssl scan.

### DIFF
--- a/api/scan/scan.go
+++ b/api/scan/scan.go
@@ -2,10 +2,7 @@ package scan
 
 import (
 	"encoding/json"
-	"io/ioutil"
-	"net"
 	"net/http"
-	"regexp"
 
 	"github.com/cloudflare/cfssl/api"
 	"github.com/cloudflare/cfssl/errors"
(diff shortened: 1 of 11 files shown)


api/scan/scan.go
```diff
@@ -0,0 +1,97 @@
...
+		return errors.NewBadRequest(err)
+	}
+
+	responses := make(map[string][]FamilyResponse)
+	for _, host := range req.Hosts {
+		if _, _, err := net.SplitHostPort(host); err != nil {
+			host = net.JoinHostPort(host, "443")
+		}
+
+		for _, family := range scan.AllFamilies {
```
grittygrease: This has too many levels of depth for a main function. Can you extract the regular expression logic into its own function and add test cases?

doc/api.txt
```diff
@@ -314,6 +314,37 @@ Example:
...
+Example:
+
+    scan.json:
+    {
+        "hosts": [
+            "example.com",
+            "cloudflare.com"
+            "localhost:44433"
+        ]
+        family: "Connectivity"
```
grittygrease: Can you add the list of families, or a way to list families here?

api/scan/scan.go
jacobhaven: Yeah, I'd agree. It's was also duplicated in the cli package. I merged them into one scan/output loop (though this breaks realtime cli output...) I might eventually make it spawn a goroutine and return response structs (easily JSON marshalable) one at a time. Having realtime results (non-buffered or flushing http response) using net/http ResponseWriter is more complicated, but possible.

doc/api.txt
jacobhaven: I added a scaninfo endpoint that is just a static list of the families/scanners, but every result natively contains a list of all the families/scanners (and their descriptions) that were run.

jacobhaven:
I set up a test server [here](http://cfssl.jhaven.me:8888/api/v1/cfssl/scan?host=cloudflare.com)

[This site](https://www.jsoneditoronline.org/?url=http://cfssl.jhaven.me:8888/api/v1/cfssl/scan?host=cloudflare.com) is pretty nice for inspecting the JSON.

grittygrease:
Looks good.

lziest:
LGTM

kisom:
LGTM.

PR #150, opened 2015-04-01
Add info endpoint to return public keys.

kisom:
Right now, clients do not have a means to access the public certificates for the signers in the multiroot CA. This PR adds an endpoint that returns a json encoded list of API certificates to the client.

### DIFF
--- a/cmd/multirootca/ca.go
+++ b/cmd/multirootca/ca.go
@@ -3,6 +3,8 @@ package main
 import (
 	"crypto/ecdsa"
 	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"net/http"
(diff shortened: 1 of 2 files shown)


cmd/multirootca/ca.go
```diff
@@ -53,12 +64,13 @@ func main() {
 	}
 
 	for label, root := range roots {
-		s, err := parseSigner(root)
+		s, c, err := parseSigner(root)
 		if err != nil {
 			log.Criticalf("%v", err)
 		}
 		signers[label] = s
 		log.Info("loaded signer ", label)
```
lziest: move this line below?

lziest:
LGTM.

grittygrease:
Why create a new endpoint instead of using a info handler from api/info?

e.g.:
import "github.com/cloudflare/cfssl/api/info"

<define signer s>
infoHandler, err := info.NewHandler(s)
http.Handle("/api/v1/cfssl/info", infoHandler)

grittygrease:
This remote info handler also does not take the incoming InfoReq structure, making it incompatible with the api/client's remote info call.

grittygrease:
In general, I think this PR needs to be re-thought. Calling "info" on a CFSSL with a remote parameter that points to a multiroot CA should "just work" with this API. Furthermore, I think listing all available labels with an API may even be a privacy issue.

PR #152, opened 2015-04-07
Add info endpoint to return public keys.

kisom:
This PR adds the info endpoint to the multiroot CA in cmd/multiroot. The info endpoint uses the label from the client info request to select a signer (falling back to the default label if no label is provided), passing this label and profile to the signer for that label, if it exists.


lziest:
I put a test on signer/local to test the previous behavior of no-op on info

grittygrease:
The current Travis CI is failing on the following test from local_test.go. Since the behavior of this function has changed, this test should also be changed.

local_test.go:274:
_, err = s.Certificate("any", "")
	if err == nil {
		t.Fatal("Certificate() currently shouldn't support label")
	}

PR #165, opened 2015-04-16
Add policy-level CSR whitelist.

jsha:
For public CA use, Let's Encrypt / Boulder would like to be able to explicitly
distrust certain fields from the untrusted input CSR. More specifically, we'd
like a whitelist of fields that can be copied from the CSR, so that any fields
added in the future will be excluded rather than included.

This change builds on / replaces @kisom's earlier change that provides a whitelist on the
Subject field, passed by the API call.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,16 +18,16 @@ import (
 	"github.com/cloudflare/cfssl/log"
 )
 
-// A CsrWhitelist stores booleans for fields in the CSR. If a CsrWhitelist is
+// A CSRWhitelist stores booleans for fields in the CSR. If a CSRWhitelist is
 // not present in a SigningProfile, all of these fields may be copied from the
-// CSR into the signed certificate. If a CsrWhitelist *is* present in a
-// SigningProfile, only those fields with a `true` value in the CsrWhitelist may
+// CSR into the signed certificate. If a CSRWhitelist *is* present in a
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -18,6 +18,20 @@ import (
...
+// A CsrWhitelist stores booleans for fields in the CSR. If a CsrWhitelist is
+// not present in a SigningProfile, all of these fields may be copied from the
+// CSR into the signed certificate. If a CsrWhitelist *is* present in a
+// SigningProfile, only those fields with a `true` value in the CsrWhitelist may
+// be copied from the CSR to the signed certificate. Note that some of these
+// fields, like Subject, can be provided or partially provided through the API.
+// Since API clients are expected to be trusted, but CSRs are not, fields
+// provided through the API are not subject to whitelisting through this
+// mechanism.
+type CsrWhitelist struct {
```
grittygrease: Go style dictates that this should be CSRWhitelist

jsha:
Pushed a new version with the fixed name. BTW, I'm not sure why Travis fails: It appears to be complaining about a missing REPOTOKEN?

kisom:
@jsha It's a known problem with coveralls.

kisom:
LGTM.

grittygrease:
LGTM

PR #170, opened 2015-04-17
Update installation instructions

dr107:
Update installation instructions to inform users about Red Hat's
removal of elliptic curve cryptography from their official go
packages.


lziest:
LGTM

PR #171, opened 2015-04-17
Add an OSCP responder

grittygrease:
This PR adds a very basic OCSP responder to CFSSL. It comes in few major
parts:

A generic interface for an OCSP "source", which maps requests to
responses
An instance of this interface that reads responses from a flat file
An OCSP Responder struct that can respond for any source
A CLI command that assembles the whole thing together. (file -> source
-> responder -> http)


kisom:
LGTM.

PR #177, opened 2015-04-23
Serve a simple HTML site with Scan UI

jacobhaven:
You can see a live example [running here](https://cfssl.jhaven.me/scan).

*Please* comment on my JavaScript/CSS/etc. I have not idea what I'm doing!


jacobhaven:
Things that need work: 
* Handling of CFSSL API errors
* Inclusion of UIs for API endpoints other that scan
* Descriptions on pages to explain what they are
* Actual templating shared between pages
* Non-hacked-together, buggy JavaScript

terinjokes:
Without doing anything fancy like pulling in a whole framework (or asking me nicely), you can make this more maintainable by using your functional programming skills to build up jQuery objects.

An example snippet:

```javascript
var rows = $.map(output, function(index, value) {
  var row = $('tr');

  if (output_type !== 'array') {
    row.append($('td').text(index);
  } else {
    row.append($('td').text(JSON.stringify(value));
  }

  return row;
});
var table = $('table').addClass('table table-bordered table-striped').append(rows);
```

cli/serve/static/scan.html
```diff
@@ -0,0 +1,106 @@
...
+                result_row += '<td><span class="glyphicon glyphicon-remove-sign"><\/span> Bad<\/td>';
+              } else {
+                result_row += '<td><span class="glyphicon glyphicon-question-sign"><\/span><\/td>';
+              }
+              result_row += '<\/tr>';
+              var output = result['output'];
+              if (output !== undefined) {
+                result_row += '<td>';
+                var output_type = $.type(output);
+                if (output_type == "object" || output_type == "array") {
```
terinjokes: You should consistently use single quotes (except for HTML attributes inside of a string).

terinjokes: Ditto for triple equals.

cli/serve/serve.go
```diff
@@ -95,6 +95,14 @@ func registerHandlers(c cli.Config) error {
...
 	http.Handle("/api/v1/cfssl/scaninfo", scan.NewInfoHandler())
 
+	log.Info("Serving static files")
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		http.ServeFile(w, r, "cli/serve/static/index.html")
+	})
+	http.HandleFunc("/scan", func(w http.ResponseWriter, r *http.Request) {
+		http.ServeFile(w, r, "cli/serve/static/scan.html")
+	})
+
```
grittygrease: This implies that the file will be stored locally from "./cli/serve/static/scan.html". It would be better to load this file at compile time.

jacobhaven: Fixed.

jacobhaven:
We should get some better description text and maybe branding so we can put this up on https://cfssl.org

grittygrease:
Looking good. Let's save the branding for a later commit.

grittygrease:
Travis is failing, please rebase to fix.

grittygrease:
Looks good enough to use for now.

PR #179, opened 2015-04-27
Fix missing certificate PEM for API newcert

slantview:
When using the API, the certificate is missing due to some leftover refactor code I narrowed down to be this PR: https://github.com/cloudflare/cfssl/commit/ea6c2ed8b0884063ef450d74c34b44f148a84f5f

This fixes it by removing certPEM and using certBytes which is what is returned from the signer.


kisom:
LGTM (failure is due to known goveralls issue). Thanks!

slantview:
Sweet, thanks!

PR #182, opened 2015-04-28
Add CSR Error

dr107:
Add a CSRError class, and make sure it is used in the places there
is a CSR problem, rather than a certificate problem.


grittygrease:
LGTM

kisom:
LGTM.

Issue #111, opened 2015-02-21
Add info command


grittygrease:
The info API endpoint returns information about the certificate used for a CA. With the latest changes from issue #98, info can return information about remote signers. There should be a command line option to return information about remote signers.

Options to take: remote, label, profile, config

dr107:
I'm a little unsure what this command is trying to do. What would be an example usage?

kisom:
It would look something like

```
$ cfssl -remote ca.example.net -profile www-server 
{"success":true,"result":{"certificate":"-----BEGIN CERTIFICATE-----\n...n-----END CERTIFICATE-----"},"errors":[],"messages":[]}
```

I think it'll mostly just be a CLI interface to the `api/client.Info` function.

dr107:
I think I get it. I'll keep poking around.

grittygrease:
All merged.

PR #184, opened 2015-05-01
Add info command

dr107:
Address issue #111

### DIFF
--- a/cli/info/info.go
+++ b/cli/info/info.go
@@ -1,0 +1,64 @@
+package info
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/cloudflare/cfssl/api"
+	"github.com/cloudflare/cfssl/api/client"
+	"github.com/cloudflare/cfssl/cli"
+	"github.com/cloudflare/cfssl/errors"
+
+	goerr "errors"
+)
+
+var infoUsageTxt = `cfssl info -- get info about a remote signer
+
+Usage:
+
+Get info about a remote signer:
+cfssl info -remote remote_host [-label label] [-profile profile] [-label label] 
+
+Flags:
+`
+
+var infoFlags = []string{"remote", "label", "profile", "config"}
+
+func infoMain(args []string, c cli.Config) (err error) {
+	if len(args) > 0 {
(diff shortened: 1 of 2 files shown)


cli/info/info.go
```diff
@@ -0,0 +1,52 @@
...
+var infoUsageTxt = `cfssl info -- get info about a remote signer`
+
+var infoFlags = []string{"remote", "label", "profile", "config"}
+
+func bundlerMain(args []string, c cli.Config) (err error) {
+	if len(args) > 0 {
+		return goerr.New("argument is provided but not defined; please refer to the usage by flag -h.")
+	}
+
+	serv := client.NewServer(c.Remote)
```
kisom: It would probably be useful to check whether `c.Remote` is empty, and let the user know if it is.

cli/info/info.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/cloudflare/cfssl/errors"
+
+	goerr "errors"
+)
+
+var infoUsageTxt = `cfssl info -- get info about a remote signer`
+
+var infoFlags = []string{"remote", "label", "profile", "config"}
+
+func bundlerMain(args []string, c cli.Config) (err error) {
```
grittygrease: This should not be called bundlerMain

grittygrease:
This can be been made more comprehensive by also handing the case where the remote is hidden in a profile by, creating signers and extracting the info using the signer.Info function. But for now this is enough.

dr107:
I have no problem doing more with this now, but I don't understand the problem. What does it mean for the remote to be hidden in a profile?

grittygrease:
This PR covers extracting info from a remote CFSSL server by directly by setting the remote in the command line. Another way CFSSL can be set up locally is by having a configuration file that defines a mix of profiles, some pointing to different remote servers. This setup allows one local server to sign certificates with multiple remote servers, with the config file serving as something like an air traffic control for request and directing calls to the right remote signer via the "profile" field. What I'm suggesting is adding a way to use the configuration file to direct the remote info file instead of directly taking the remote from the command line.

grittygrease:
Add the parameters into the usage text and this is good to merge.

lziest:
LGTM

PR #185, opened 2015-05-01
Rewrite/refactor SHA1 deprecation logic and fix tests.

lziest:
This is to fix the test issue blocking other pull requests. The initial static tests on bundler are showing their age and need to be improve/rewritten. This is the first step of this. And more will come.

Summary of changes:
- Revert the initial design of SHA1 deprecation policy being part of
  platform metadata because it makes metadata bloated. Instead, the
  platform metadata should only contain info on CA trust root stores
  and crypto support.
- New design of SHA1 deprecation consideration
	- It's only about browsers.
	- It's currently only for warning purpose.
	- The list of SHA1 deprecated browsers are much more static than
	  root certificate stores.
- Create a new type called SHA1Deprecated to handle SHA1
  deprecation warning.
- Make it explicit that ubiquity scoring doesn't take SHA1 deprecation
  into account. Right now the weights of Chrome browsers in the original
  metadata file is encoded as 0. So this is a safe change. And the scoring
  can be changed later, in a more flexible way.
- Hardcode the list of Chrome versions that would conditionally reject
  SHA1 certificate chain. And we can certainly add more in the future.
- Rename function DeprecatedSHA1Platforms to DeprecatedSHA1Browsers
- Make sure the tests about SHA1 deprecation warning is the same in
  bundler.
- Switch from static tests of SHA1 deprecation warning to dynamic ones
  in bundler package.
- improve test coverage in ubiquity package


lziest:
will increase test coverage, don't merge.

lziest:
now ready

grittygrease:
Zi, the logic about Chrome <42 was relevant and helpful when this was written, but less relevant now that Chrome 42 has been released to mainline Chrome users. You can reduce complexity even more by updating the logic to check for SHA-1 after Dec 31 2015 only for Chrome 42.

Also, It might be worth looking into Firefox's plan to deprecate SHA-1 and add that logic.

lziest:
Add support of Mozilla SHA1 deprecation policy:
- After 01-01-2016, reject newly issued SHA-1 leaf certificate.
- After 01-01-2017, reject any valid SHA-1 cert chain.
- Currently, any SHA-1 cert chain valid beyond 01-01-2017 will have a warning in developer console.

lziest:
Microsoft's policy is the same as the 2nd policy of Mozilla:
- After 01-01-2017, reject any valid SHA-1 cert chain.

ubiquity/testdata/ca.pem.metadata
```diff
@@ -0,0 +1,27 @@
+[
+{
+	"name":"Chrome Browser M39",
```
grittygrease: do you want to test the current "real" metadata? The Firefox logic is slightly different from Chrome (chain vs. leaf) and is worth testing. The chain creation code in helpers should be able to help you dynamically create the test cases.

lziest:
Close this. Going to split this into two PRs.

PR #189, opened 2015-05-04
fix bundler tests

lziest:
- replace existing static tests with dynamic ones on SHA-1 deprecation warnings.
- remove inaccurate tests on SHA2 homogenity. Will add a dynamic tests
  to cover it.


kisom:
LGTM.

PR #190, opened 2015-05-05
Zi/refactor sha1 deprecation

lziest:
Rewrite/refactor SHA1 deprecation logic and improve test coverage.
- Revert the initial design of SHA1 deprecation policy being part of
  platform metadata because it makes metadata bloated. Instead, the
  platform metadata should only contain info on CA trust root stores
  and crypto support.
- Add a new type called SHA1Deprecated to handle SHA1
  deprecation warning.
- Add support to Firefox SHA1 deprecation policy in SHA1Deprecated.
- Make it explicit that ubiquity scoring doesn't take SHA1
  deprecation into account. Right now the weights of Chrome browsers
  in the metadata file is encoded as 0. So this is a safe change.
- Hardcode the list of platforms that would conditionally reject
  SHA1 certificate chain.
- Increase test coverage of package ubiquity
- Add a dynamic bundle test on prioritizing SHA-2 intermediates when the leaf cert is a SHA-2 cert.

### DIFF
--- a/bundler/bundler.go
+++ b/bundler/bundler.go
@@ -47,12 +47,11 @@ const (
 )
 
 const (
-	sha2Warning              = "The bundle contains certs signed with advanced hash functions such as SHA2,  which are problematic at certain operating systems, e.g. Windows XP SP2."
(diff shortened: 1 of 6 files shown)


grittygrease:
This looks great and ready to merge.

kisom:
LGTM.

PR #192, opened 2015-05-06
info endpoint now returns more usage and expiry string

timestruck:


api/info/info.go
```diff
@@ -50,8 +51,22 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
...
+	}
+
+	if profile == nil && policy != nil {
+		profile = policy.Default
+	}
+
 	resp := client.InfoResp{
-		Certificate: bundler.PemBlockToString(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw}),
+		Certificate:  bundler.PemBlockToString(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw}),
+		Usage:        profile.Usage,
```
lziest: there is possibility that profile is nil at this point. Dereferencing nil would be bad. Consider declare 'usage' and 'expiryString' explicitly and extract them from the above policy, and if the policy is nil, or policy.Profile[req.Profile] is nil, those values will be naturally nil.

timestruck: Is this what you mean?

        policy := h.sign.Policy()
if policy != nil && policy.Profiles != nil && req.Profile != "" {
profile = policy.Profiles[req.Profile]
}

if profile == nil && policy != nil {
profile = policy.Default
}

        var usage []string
        var expiry string

        if policy == nil {
                 usage = nil
                 expiry = ""
        } else {
                 usage = profile.Usage
                 expiry = profile.ExpiryString
        }

resp := client.InfoResp{
Certificate:  bundler.PemBlockToString(&pem.Block{Type: "CERTIFICATE",
Bytes: cert.Raw}),
Usage:        usage,
ExpiryString: expiry,
}

On Wed, May 6, 2015 at 1:23 PM, Zi Lin <notifications@github.com> wrote:

> In api/info/info.go
> <https://github.com/cloudflare/cfssl/pull/192#discussion_r29780788>:
>
> >  	resp := client.InfoResp{
> > -		Certificate: bundler.PemBlockToString(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw}),
> > +		Certificate:  bundler.PemBlockToString(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw}),
> > +		Usage:        profile.Usage,
>
> there is possibility that profile is nil at this point. Dereferencing nil
> would be bad. Consider declare 'usage' and 'expiryString' explicitly and
> extract them from the above policy, and if the policy is nil, or
> policy.Profile[req.Profile] is nil, those values will be naturally nil.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cloudflare/cfssl/pull/192/files#r29780788>.
>



-- 
Alice Xia
Cornell University '16

kisom: With profile, it *shouldn't* be nil after the initial setting (though it *could* be); a failure check should be done there. Then you can use the profile without worrying it it's nil.

timestruck: I see, so I should be checking if profile is nil, rather than if policy is
nil?

On Wed, May 6, 2015 at 2:51 PM, Kyle Isom <notifications@github.com> wrote:

> In api/info/info.go
> <https://github.com/cloudflare/cfssl/pull/192#discussion_r29789464>:
>
> >  	resp := client.InfoResp{
> > -		Certificate: bundler.PemBlockToString(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw}),
> > +		Certificate:  bundler.PemBlockToString(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw}),
> > +		Usage:        profile.Usage,
>
> With profile, it *shouldn't* be nil after the initial setting (though it
> *could* be); a failure check should be done there. Then you can use the
> profile without worrying it it's nil.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cloudflare/cfssl/pull/192/files#r29789464>.
>



-- 
Alice Xia
Cornell University '16

api/info/info.go
```diff
@@ -50,8 +51,28 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
...
+	policy := h.sign.Policy()
+	if policy != nil && policy.Profiles != nil && req.Profile != "" {
+		profile = policy.Profiles[req.Profile]
+	}
+	if profile == nil && policy != nil {
+		profile = policy.Default
+	}
+	var usage []string
+	var expiry string
+	if profile == nil {
```
lziest: per @kisom, if the profile is nil, response with error. So ignore what I suggested.

api/info/info.go
```diff
@@ -50,8 +51,21 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
...
+
+	var profile *config.SigningProfile
+	policy := h.sign.Policy()
+	if policy != nil && policy.Profiles != nil && req.Profile != "" {
+		profile = policy.Profiles[req.Profile]
+	}
+	if profile == nil && policy != nil {
+		profile = policy.Default
+	}
+	var usage []string
```
grittygrease: These two variable declarations are not needed.

kisom:
Needs rebase.

lziest:
Looks like this has been merged as part of Dan's PR.

PR #194, opened 2015-05-06
Dan/info additions

dr107:


cli/info/info.go
```diff
@@ -44,9 +47,35 @@ func infoMain(args []string, c cli.Config) (err error) {
 	if err != nil {
 		return err
 	}
+
+	_, err = helpers.ParseCertificatePEM(certPem)
+	if err != nil {
+		return errors.Wrap(errors.CertificateError, errors.DecodeFailed, err)
```
lziest: Thanks for testing this. I think the err is already a cfssl error, so double wrapping it would trigger an panic here.

cli/info/info.go
```diff
@@ -44,9 +47,35 @@ func infoMain(args []string, c cli.Config) (err error) {
...
+	}
+
+	s, err := sign.SignerFromConfig(c)
```
lziest: This is questionable for remote signer. I understand this is confusing because it is the frontier of development.
A remote signer's profile has no actual signing policy, it only has auth key config and a remote address. So getting usage and expiry string from that would just be zero values.
Although the cli command parameter '-config' is included, the purpose is to provide the auth key to generate auth token. But, right now the authentication is not yet turned on. Since the info endpoint is currently just returning a certificate, it is considered safe to reveal anyway. So for now, let's ignore '-config'

The real problem is, if api/info starts to return more information, the api/client need to adopt to the changes, at line 46, the certificate is returned by api/client, and take a look at api/client, we realize the result is actually a map, but api/client.Server captures the value under 'certificate' and return it. So change api/client to actually return the true response of api info, so the usage and expiry string should be just two lookups.

grittygrease: @dr107 There's a bit more plumbing needed to get the full feature.

There are two cases: 1) when -remote is not empty 2) when -remote is empty and -config is not
  
In 1) use the -remote and the api/client package if the -remote flag is there. The api/client package should be updated to return the Usage and ExpiryString that the info API returns. Right now it only returns the certificate.

If the -remote flag is not there, then it's 2). In this case continue to use the config package to get a signer.Signer using sign.SignerFromConfig(c). Update the code in the various signers so that s.Info() returns Usage and ExpiryString along with Certificate, and use those to construct the InfoResp.

dr107: I see. For some reason, the map in Server.Req only had "certificate" as a key before, so I was kind of floundering to get the other data. This should be no problem.

api/client/client.go
```diff
@@ -145,22 +145,47 @@ func (srv *Server) Sign(jsonData []byte) ([]byte, error) {
...
+	}, nil
+}
+
+func (srv *Server) getResultMap(jsonData []byte, target string) (result map[string]interface{}, err error) {
+	url := srv.getURL(target)
+	response, err := srv.post(url, jsonData)
+	if err != nil {
+		return
+	}
+	result = response.Result.(map[string]interface{})
```
kisom: A few things here:

* You should check whether response.Result is nil
* You should make sure it can be coerced to a `map[string]interface{}`. You can do that with something like:

```
result, ok := response.Result.(map[string]interface{})
if !ok {
        // return an error
}
return result, nil
```

api/info/info.go
```diff
@@ -50,8 +51,22 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
...
+
+	policy := h.sign.Policy()
+	if policy != nil && policy.Profiles != nil && req.Profile != "" {
+		profile = policy.Profiles[req.Profile]
+	}
+
+	if profile == nil && policy != nil {
+		profile = policy.Default
+	}
+
```
kisom: Should verify that profile is not nil before using its fields below. It shouldn't be nil in normal operation, so you can fail with an error if it is.

api/client/client.go
```diff
@@ -145,22 +145,47 @@ func (srv *Server) Sign(jsonData []byte) ([]byte, error) {
...
+	res, err := srv.getResultMap(jsonData, "info")
+	if err != nil {
+		return nil, err
+	}
+
+	cert := res["certificate"]
+	usages := res["usages"].([]interface{})
+	exp := res["expiry"]
+
+	usageStrings := make([]string, len(usages), len(usages))
```
kisom: You don't need to set the slice's capacity if you set its length.

cli/info/info.go
```diff
@@ -24,29 +28,86 @@ Flags:
...
+			return
+		}
+
+	} else if c.CFG != nil {
+		resp, err = getInfoFromConfig(c)
+		if err != nil {
+			return
+		}
+	} else {
+		return goerr.New("Either -remote or -config must be given. Do cfssl info -h for usage.")
```
kisom: Instead of 'Do cfssl info -h', I would phrase it "Refer to cfssl info -h" or something along those lines.

kisom:
LGTM.

PR #196, opened 2015-05-08
Harry/scanserver

hwh33:
Implements a test server for the bundler and scan packages (addressing issues #117 and #96).  This test server is flexible, so it should be usable for future packages as well.  Also re-factors the existing CFSSL server.


helpers/testsuite/test_certificates.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, nil, err
+	}
+	tempFile, err := createTempFile(jsonBytes)
+	if err != nil {
+		os.Remove(tempFile)
+		return nil, nil, err
+	}
+
+	// Create the certificate with the CLI tools.
+	command := exec.Command("cfssl", "gencert", "-initca", tempFile)
```
lziest: this is sort of complicated. why not using initca.New(request) directly? That's much easier, right?

helpers/testsuite/test_certificates.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, nil, err
+	}
+	tempJSONFile, err := createTempFile(jsonBytes)
+	if err != nil {
+		os.Remove(tempJSONFile)
+		return nil, nil, err
+	}
+
+	// Create a CSR file with the CLI tools.
+	command := exec.Command("cfssl", "genkey", tempJSONFile)
```
lziest: similarly, we could construct a local signer from the cert and key. Generate CSR bytes and key bytes from request using csr.ParseRequest. The only tricky part is to construct the policy, which I would attach a snippet to help you:

	defaultProfile := &config.SigningProfile{
		Usage:        []string{"cert sign", "server auth"}, // so the signed certificate can be used as intermediate CA and server auth certificate. You can adjust it for your need.
		CA:           true,
		Expiry:       time.Hour,
		ExpiryString: "1h",
	}
	policy := &config.Signing{
		Profiles: map[string]*config.SigningProfile{},
		Default:  defaultProfile,
	}
	s, err := local.NewSigner(key, cert, sigAlgo, policy)

lziest:
Also, it appears helpers/testsuite/test_certificates.go is defining helper functions that are not utilized. Separate it for a second commit or omit it for this pull request to avoid confusion.

helpers/testsuite/testsuite_test.go
```diff
@@ -66,23 +83,123 @@ var (
...
 
+func TestStartTestServer(t *testing.T) {
+	addr := "127.0.0.1"
+	portTLS := "7050"
+	portTCP := "8050"
+
+	certFile := "testdata/cert/cert_chain.crt"
+	keyFile := "testdata/cert/decrypted.ssl.key"
+
+	// caCertFile := "testdata/initCA/ca.pem"
```
lziest: comments or unused code?

helpers/testsuite/test_servers.go
```diff
@@ -0,0 +1,230 @@
+// Servers which can be configured then run locally to test various parts of
```
lziest: name this file "servers.go" instead of "test_servers.go"?

helpers/testsuite/testsuite_helpers.go
```diff
@@ -0,0 +1,70 @@
+// Helper functions used internally in the testsuite package.
```
lziest: ditto. rename the file, "helpers.go" may sound redundant, how about "cli.go"?

helpers/testsuite/testsuite_test.go
```diff
@@ -66,23 +83,123 @@ var (
...
+	err = tlsServer.UseDefaultTLSConfig(cert, key)
+	checkError(err, t)
+	err = tlsServer.Start()
+	checkError(err, t)
+
+	// === TEST: bundle using the test server ============================== //
+
+	newBundler, err := bundler.NewBundler(certFile, certFile)
+	checkError(err, t)
+	_, err = newBundler.BundleFromRemote("harryharpham.me", addr, portTLS, bundler.Optimal)
```
lziest: Thank you! Can you also setup the same test under bundler?

helpers/testsuite/testsuite_test.go
```diff
@@ -66,23 +83,123 @@ var (
...
+					", Grade: " + result.Grade + "]")
+			}
+			if result.Grade != scan.Good.String() {
+				t.Fatal("The following scan failed: [Family: " + family +
+					", Scanner: " + scanner + ", Grade: " + result.Grade + "]")
+			}
+		}
+	}
+
+	// === TEST: test the connectivity scan's response to a TCP-only server //
```
lziest: Also, replicate this cool test to scan package if possible.

lziest:
Good job. Thanks.

lziest:
LGTM

lziest:
The commit message may need some cleanup though.

hwh33:
Ah yes, sorry about that.  How is that?  It's a bit verbose, but I added/changed a few distinct things and had trouble summarizing.  Let me know if there's something you would prefer.

lziest:
@hwh33 you can do multi-line commit message. The first line is the summary. And details can be itemized as "- add test servers", "- redo certificate generation" "- add tests to bundler and scan" etc. in the following lines.

hwh33:
@lziest Oh okay, I see.  I'll go ahead and change that.  Thank you!

helpers/testsuite/servers_test.go
```diff
@@ -0,0 +1,181 @@
...
+	// should get an 'address in use' error.
+	server2 := CFSSLServer{
+		Addr:  server.Addr,
+		Port:  server.Port,
+		CA:    CACert,
+		CAKey: CAKey,
+	}
+	err = server2.Start()
+	if err == nil || !strings.Contains(err.Error(), "Error occurred on server: address") {
+		t.Fatal("Two servers allowed on same address and port.")
```
lziest: I think it is ok not to test this. This is causing race condition in the tests and we probably never start two servers in the functional tests.

hwh33: @lziest Okay, that makes sense.  I removed the test.

grittygrease:
@kisom @lziest What's the status of this PR? Are there issues in this code that are not resolved?

PR #199, opened 2015-05-11
Add extra documentation to sub-packages.

grittygrease:


lziest:
LGTM

kisom:
Style nit: the package docs should be complete sentences with punctuation. Other than that, LGTM.

lziest:
golint is more strict. It requires the doc starts with " Package X" rather than "Package api/X"

api/initca/initca.go
```diff
@@ -1,3 +1,4 @@
+// Package api/initca implements the HTTP handler for the CA initialization command
```
lziest: "api/initca" -> "initca"

PR #201, opened 2015-05-11
Make info -remote command safe

grittygrease:
The info command panics if there is an older cfssl at the other end or if an error is returned.


lziest:
LGTM.

kisom:
LGTM.

Issue #209, opened 2015-05-15
IP Address Alternate Name


TheDeadCode:
Hi! It seems that IP address in the subject alt name is not supported.

Please see [here](http://tools.ietf.org/html/rfc2818) on section 3.1 paragraph 5.

 > In some cases, the URI is specified as an IP address rather than a
   hostname. In this case, the iPAddress subjectAltName must be present
   in the certificate and must exactly match the IP in the URI.

lziest:
Hi @TheDeadCode, we add the support to the signer, which is capable to override the hostname/IP addresses, check [here](https://github.com/cloudflare/cfssl/blob/master/signer/local/local.go#L151)
However, we haven't add the support to the csr package. We will fix it now.

lziest:
Should be solved.

PR #210, opened 2015-05-15
Add support of IP SAN to csr package

lziest:
This addressed #209


lziest:
Yes, cli with comma-separated hostnames/IPs are covered in signer tests already. Also I manually tests with cli signer etc. The cli equivalent of csr is cli/genkey, which doesn't take comma-separated hostnames/IPs.

TheDeadCode:
Is there any ETA on the pull request?

grittygrease:
LGTM

PR #212, opened 2015-05-16
cli bundle genkey test

ch684:


cli/bundle/bundle_test.go
```diff
@@ -1 +1,33 @@
...
+	testdata := path.Join(gopath, "src", "github.com", "cloudflare", "cfssl", "bundler", "testdata")
+	pkg := path.Join(gopath, "src", "github.com", "cfssl_trust")
+	//cfssl bundle [-ca-bundle file] [-int-bundle file] [-key keyfile] [-flavor int] [-metadata file] CERT
+	out1, err1 := exec.Command(cfssl, "bundle",
+		"-cert="+path.Join(testdata, "bunnings.pem"),
+		"-ca-bundle="+path.Join(pkg, "ca-bundle.crt"),
+		"-int-bundle="+path.Join(pkg, "int-bundle.crt"),
+		"-metadata="+path.Join(pkg, "ca-bundle.crt.metadata")).CombinedOutput()
+	fmt.Printf("cfssl bundle [-ca-bundle file] [-int-bundle file] [-metadata file] [-key keyfile] [-flavor int]  CERT\n%s\n", string(out1))
+	fmt.Println("Error: ", err1)
```
lziest: printing out result is not a real test, since most of time stdout/stderr output is omitted. It's expected we do t.Error(), t.Fail() or t.Fatal() at least once.

cli/bundle/bundle_test.go
```diff
@@ -1 +1,33 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"testing"
+)
+
+func TestBundle(t *testing.T) {
```
lziest: If we aim for test coverage, we probably just test individual functions in those packages. The test engine is not smart enough to capture code path covered by exec.Command in a separate process :(

cli/genkey/genkey_test.go
```diff
@@ -1 +1,30 @@
...
+)
+
+func TestGenkey(t *testing.T) {
+	gopath := os.Getenv("GOPATH")
+	cfssl := path.Join(gopath, "bin", "cfssl")
+	testdata := path.Join(gopath, "src", "github.com", "cloudflare", "cfssl", "testdata")
+
+	out1, err1 := exec.Command(cfssl, "genkey", path.Join(testdata, "csr.json")).CombinedOutput()
+	fmt.Printf("cfssl genkey CSRJSON\n%s\n", string(out1))
+	fmt.Println("Error: ", err1)
```
lziest: With the error checking below, this line is not long needed.

cli/genkey/genkey_test.go
```diff
@@ -1 +1,30 @@
...
+	"testing"
+)
+
+func TestGenkey(t *testing.T) {
+	gopath := os.Getenv("GOPATH")
+	cfssl := path.Join(gopath, "bin", "cfssl")
+	testdata := path.Join(gopath, "src", "github.com", "cloudflare", "cfssl", "testdata")
+
+	out1, err1 := exec.Command(cfssl, "genkey", path.Join(testdata, "csr.json")).CombinedOutput()
+	fmt.Printf("cfssl genkey CSRJSON\n%s\n", string(out1))
```
lziest: could you parse the result? it is a JSON object with csr and key, right?

cli/genkey/genkey_test.go
```diff
@@ -1 +1,30 @@
...
+	fmt.Println("Error: ", err1)
+	if err1 != nil {
+		t.Fatal(err1)
+	}
+
+	c := cli.Config{}
```
lziest: Add one more test with C.IsCA is true.

cli/genkey/genkey_test.go
```diff
@@ -1 +1,42 @@
...
+	err2 := genkeyMain([]string{path.Join(testdata, "csr.json")}, c)
+	if err2 != nil {
+		t.Fatal(err2)
+	}
+
+	c.IsCA = true
+
+	err3 := genkeyMain([]string{path.Join(testdata, "csr.json")}, c)
+	if err2 != nil {
+		t.Fatal(err3)
```
lziest: just reuse 'err'. As you can see here, it is problematic and error-prone to use adhoc new variables and copy/paste old logic.

cli/genkey/genkey_test.go
```diff
@@ -1 +1,45 @@
...
+
+func TestGenkey(t *testing.T) {
+	//testing through console
+	gopath := os.Getenv("GOPATH")
+	cfssl := path.Join(gopath, "bin", "cfssl")
+	testdata := path.Join(gopath, "src", "github.com", "cloudflare", "cfssl", "testdata")
+
+	out1, err := exec.Command(cfssl, "genkey", path.Join(testdata, "csr.json")).Output()
+	var response map[string]interface{}
+	json.Unmarshal(out1, &response)
```
lziest: this is not satisfactory. Check the error and make sure the content of response is what you need. What key would response have? What value would be corresponding to each key? etc.

cli/genkey/genkey_test.go
```diff
@@ -1 +1,45 @@
...
+
+	if err != nil {
```
lziest: this should move to be immediately follow where err is assigned, namely at line 19

cli/genkey/genkey_test.go
```diff
@@ -1 +1,45 @@
...
+	if response["key"] == nil {
+		t.Fatal("No key is outputted.")
+	}
+	if response["csr"] == nil {
+		t.Fatal("No csr is outputted.")
+	}
+
+	c := cli.Config{}
+
+	err1 := genkeyMain([]string{path.Join(testdata, "csr.json")}, c)
```
lziest: I guess you have tried to write "err :=" here, which would be caught by compiler as error. ":=" means create a new variable, "=" is a simple assignment. Since err is already created, use "=" here. Again, don't use "err1" or "err2". variable with number suffix is always bad, in practice.

cli/genkey/genkey_test.go
```diff
@@ -1 +1,45 @@
...
+	if err != nil {
+		t.Fatal(err1)
```
lziest: This whole if statement doesn't make sense. I hope you can spot the problem here.

cli/genkey/genkey_test.go
```diff
@@ -1 +1,45 @@
...
+	}
+
+	c.IsCA = true
+
+	err2 := genkeyMain([]string{path.Join(testdata, "csr.json")}, c)
+	if err != nil {
+		t.Fatal(err2)
```
lziest: same problem.

lziest:
let's focus on getting cli/genkey/genkey_test.go work as expected. Would you remove cli/bundle/bundle_test.go from this pull request? Basically as I mentioned before, we are moving away from the static test with files that would expire at certain time, causing tests to fail.

cli/genkey/genkey_test.go
```diff
@@ -1 +1,46 @@
...
+	cfssl := path.Join(gopath, "bin", "cfssl")
+	testdata := path.Join(gopath, "src", "github.com", "cloudflare", "cfssl", "testdata")
+
+	out, err := exec.Command(cfssl, "genkey", path.Join(testdata, "csr.json")).Output()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var response map[string]interface{}
+	json.Unmarshal(out, &response)
```
lziest: almost done. Check the error returned by this statement.

lziest:
cool, LGTM

cli/genkey/genkey_test.go
```diff
@@ -1 +1,49 @@
 package genkey
+
+import (
+	"encoding/json"
+	"github.com/cloudflare/cfssl/cli"
```
lziest: One last thing. Put this import in a separated block. Examples can be found in the code base.

PR #218, opened 2015-05-20
Print the result of the API response in info.

### LINKED ISSUE #217: Info command line consistency
When the info command is run, it returns the data in the API wrapping. However, this is inconsistent with the way the other CLI commands work; namely, `cfssljson` won't work with the `-bare` option.

kisom:
This fixes #217.


lziest:
LGTM

PR #219, opened 2015-05-21
scan: check for revocation in PKI ChainValidation

jacobhaven:
Mainly for use in a "Strict SSL"-compatibility checker. (CFSSL-149)

### DIFF
--- a/scan/pki.go
+++ b/scan/pki.go
@@ -4,11 +4,11 @@ import (
 	"bytes"
 	"crypto/x509"
 	"fmt"
-	"net"
 	"time"
 
 	"github.com/cloudflare/cf-tls/tls"
 	"github.com/cloudflare/cfssl/helpers"
+	"github.com/cloudflare/cfssl/revoke"
 )
 
 // PKI contains scanners for the Public Key Infrastructure.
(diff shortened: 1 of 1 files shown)


grittygrease:
LGTM

scan/pki.go
```diff
@@ -94,6 +95,7 @@ func chainValidation(host string) (grade Grade, output Output, err error) {
 		return
 	}
 
+	// This is probably a redundant check
```
kisom: Style nitpicks: this should be a complete sentence. It would also be useful to explain why it might be redundant and why it's been kept in (even if the answer is just that the comment serves as a marker to go back and look at the code later).

kisom:
One nit about a comment, other than that LGTM.

scan/pki.go
jacobhaven: I didn't dive deep into all the checks [x509#Certificate.VerifyHostname](https://golang.org/pkg/crypto/x509/#Certificate.VerifyHostname) is doing, but I _think_ `cfssl/revoke#VerifyCertificate` is covers a superset of them (obviously adding CRL and OCSP revocation checks).

I think it's probably best to just remove this "quick-fail" method, as it will give less descriptive errors.

PR #222, opened 2015-05-26
Add support for scanning server elliptic curve preferences.

### LINKED ISSUE #221: Scan for server-support of elliptic curves
Requires a few modifications to [cf-tls](https://github.com/cloudflare/cf-tls/blob/master/tls/cfsslscan_common.go)
* Adding [all the RFC 4492 and RFC 7027 curves](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8), storing them similarly to the ciphersuites (with a map of uint16s to various (IANA, OpenSSL, etc) names returned via a String() method for a canonical human name) Type should be [CurveID](https://tip.golang.org/pkg/crypto/tls/#CurveID)
* extension of [SayHello](https://github.com/cloudflare/cf-tls/blob/master/tls/cfsslscan_handshake.go) to add an extra chosen `curveID` return value.

And then in scan, just adding another `curveID` argument to `sayHello`, included in calls inside [cipherSuiteScan](https://github.com/cloudflare/cfssl/blob/master/scan/tls_handshake.go#L108) that loops for every confirmed good EC suite.

jkroll:
This change adds support for scanning server curve choice preferences for the ECDH cipher suites, which is interesting in light of logjam (even though that only affects non-ED DH, it's nice to know what's happening in case something similar becomes interesting in the future). This fixes #221, although it's slightly more involved than @jacobhaven suggested. Specifically, the server curve preference is not declared until the ServerKeyExchange message, meaning that you have to go well past ServerHello to extract it.

The scan is performed using the same algorithm as for cipher suites, where the server's negotiated preference is dropped iteratively from the list of all possible choices until we've enumerated the server's preference list.

Update: the scan results now seem correct (assuming the right return is only `secp256r1` for all ECDH cipher suites CloudFlare supports - I can't tell from the published SSL config which curve(s) are used). The only remaining issue is that the handshake in cf-tls sometimes fails in specific conditions (assuming the scan results are correct).

NB - the buildbot always fails here because this code depends on unmerged changes to cf-tls. Is there a way to specify which branch it should build from?

### DIFF
--- a/scan/tls_handshake.go
+++ b/scan/tls_handshake.go
@@ -10,6 +10,9 @@ import (
 	"github.com/cloudflare/cf-tls/tls"
 )
 
+// Sentinel for failures in sayHello. Should always be caught.
+var errHelloFailed = errors.New("Handshake failed in sayHello")
+
 // TLSHandshake contains scanners testing host cipher suite negotiation
 var TLSHandshake = &Family{
 	Description: "Scans for host's SSL/TLS version and cipher suite negotiation",
@@ -21,7 +24,31 @@ var TLSHandshake = &Family{
 	},
 }
 
-func sayHello(host string, ciphers []uint16, vers uint16) (cipherIndex int, err error) {
+func getCipherIndex(ciphers []uint16, serverCipher uint16) (cipherIndex int, err error) {
+	//func getCipherIndex(ciphers []uint16, serverCipher uint16) (cipherIndex int, err error) {
+	//	fmt.Println(serverCipher, ciphers)
+	var cipherID uint16
+	for cipherIndex, cipherID = range ciphers {
+		if serverCipher == cipherID {
+			return
+		}
+	}
+	err = fmt.Errorf("server negotiated ciphersuite we didn't send: %s", tls.CipherSuites[serverCipher])
+	return
+}
+
+func getCurveIndex(curves []tls.CurveID, serverCurve tls.CurveID) (curveIndex int, err error) {
+	var curveID tls.CurveID
+	for curveIndex, curveID = range curves {
+		if serverCurve == curveID {
+			return
+		}
+	}
+	err = fmt.Errorf("server negotiated elliptic curve we didn't send: %s", tls.Curves[serverCurve])
+	return
+}
+
+func sayHello(host string, ciphers []uint16, curves []tls.CurveID, vers uint16) (cipherIndex, curveIndex int, err error) {
 	tcpConn, err := net.Dial(Network, host)
 	if err != nil {
 		return
@@ -30,10 +57,12 @@ func sayHello(host string, ciphers []uint16, vers uint16) (cipherIndex int, err 
 	config.MinVersion = vers
 	config.MaxVersion = vers
 	config.CipherSuites = ciphers
+	config.CurvePreferences = curves
 	conn := tls.Client(tcpConn, config)
-	serverCipher, serverVersion, err := conn.SayHello()
+	serverCipher, serverCurveType, serverCurve, serverVersion, err := conn.SayHello()
 	conn.Close()
 	if err != nil {
+		err = errHelloFailed
 		return
 	}
 
@@ -42,13 +71,18 @@ func sayHello(host string, ciphers []uint16, vers uint16) (cipherIndex int, err 
 		return
 	}
 
-	var cipherID uint16
-	for cipherIndex, cipherID = range ciphers {
-		if serverCipher == cipherID {
-			return
+	cipherIndex, err = getCipherIndex(ciphers, serverCipher)
+
+	if tls.CipherSuites[serverCipher].EllipticCurve {
+		if curves == nil {
+			curves = allCurvesIDs()
 		}
+		if serverCurveType != 3 {
+			err = fmt.Errorf("server negotiated non-named ECDH parameters; we didn't analyze them. Server curve type: %d", serverCurveType)
+		}
+		curveIndex, err = getCurveIndex(curves, serverCurve)
 	}
-	err = fmt.Errorf("server negotiated ciphersuite we didn't send: %s", tls.CipherSuites[serverCipher])
+
 	return
 }
 
@@ -60,10 +94,29 @@ func allCiphersIDs() []uint16 {
 	return ciphers
 }
 
-// cipherVersions contains lists of host's supported cipher suites based on SSL/TLS Version
+func allCurvesIDs() []tls.CurveID {
+	curves := make([]tls.CurveID, 0, len(tls.Curves))
+	for curveID := range tls.Curves {
+		// No unassigned or explicit curves in the scan, per http://tools.ietf.org/html/rfc4492#section-5.4
+		if curveID == 0 || curveID == 65281 || curveID == 65282 {
+			continue
+		} else {
+			curves = append(curves, curveID)
+		}
+	}
+	return curves
+}
+
+type cipherDatum struct {
+	versionID uint16
+	curves    []tls.CurveID
+}
+
+// cipherVersions contains lists of host's supported cipher suites based on SSL/TLS Version.
+// If a cipher suite uses ECC, also contains a list of supported curves by SSL/TLS Version.
 type cipherVersions struct {
 	cipherID uint16
-	versions []uint16
+	data     []cipherDatum
 }
 
 type cipherVersionList []cipherVersions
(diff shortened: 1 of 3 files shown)


scan/tls_handshake.go
```diff
@@ -94,18 +133,39 @@ func (cvList cipherVersionList) MarshalJSON() ([]byte, error) {
...
+	curves := make([]tls.CurveID, len(allCurves))
+	copy(curves, allCurves)
+	for len(curves) > 0 {
+		cipherIndex, curveIndex, err := sayHello(host, []uint16{cipherID}, curves, vers)
+		if err != nil {
+			break
+		}
+		curveScanCipherID := ciphers[cipherIndex]
+		if (cipherID != curveScanCipherID) {
+			err = fmt.Errorf("server returned a different cipher suite during follow-up curve scan: requested %s, got %s", tls.CipherSuites[cipherID], tls.CipherSuites[curveScanCipherID])
```
jacobhaven: This is already handled as an error from `sayHello`, so this is dead code.

jkroll: Good point. I removed the redundant check here.

grittygrease:
Pending changes to cf-tls.

scan/tls_handshake.go
```diff
@@ -105,21 +188,35 @@ func cipherSuiteScan(host string) (grade Grade, output Output, err error) {
...
-			cipherIndex, err := sayHello(host, ciphers, vers)
+			var cipherIndex int
+			cipherIndex, _, err = sayHello(host, ciphers, nil, vers)
 			if err != nil {
-				break
+				if err == tls.HandshakeFailureErr {
+					err = nil
+					break
+				}
+				return
```
jacobhaven: In non `alertHandshakeFailure` situations, we still treat the `err` from `sayHello` as a "true error". _Any_ TLS alert should just be ignored and treated with a break from the inner loop, not a return.

You're right that we probably shouldn't log these alerts, since they're expected. It's probably simplest to add a sentinel value inside the `scan` package.

scan/tls_handshake.go
```diff
@@ -30,8 +54,9 @@ func sayHello(host string, ciphers []uint16, vers uint16) (cipherIndex int, err
 	config.MinVersion = vers
 	config.MaxVersion = vers
 	config.CipherSuites = ciphers
+	config.CurvePreferences = curves
 	conn := tls.Client(tcpConn, config)
-	serverCipher, serverVersion, err := conn.SayHello()
+	serverCipher, serverCurveType, serverCurve, serverVersion, err := conn.SayHello()
 	conn.Close()
 	if err != nil {
 		return
```
jacobhaven:
```
err = helloFailed
return
```

scan/tls_handshake.go
```diff
@@ -21,7 +21,31 @@ var TLSHandshake = &Family{
...
+	var curveID tls.CurveID
+	for curveIndex, curveID = range curves {
+		if serverCurve == curveID {
+			return
+		}
+	}
+	err = fmt.Errorf("server negotiated elliptic curve we didn't send: %s", tls.Curves[serverCurve])
+	return
+}
+
```
jacobhaven: ```var helloFailed = errors.New("scan: SayHello failed")```

scan/tls_handshake.go
```diff
@@ -105,21 +188,35 @@ func cipherSuiteScan(host string) (grade Grade, output Output, err error) {
 		ciphers := make([]uint16, len(allCiphers))
 		copy(ciphers, allCiphers)
 		for len(ciphers) > 0 {
-			cipherIndex, err := sayHello(host, ciphers, vers)
+			var cipherIndex int
+			cipherIndex, _, err = sayHello(host, ciphers, nil, vers)
 			if err != nil {
```
jacobhaven:
```
		if err == helloFailed {
			err = nil
			break
		} else if err != nil {
			return
		}
```

scan/tls_handshake.go
```diff
@@ -94,6 +155,28 @@ func (cvList cipherVersionList) MarshalJSON() ([]byte, error) {
...
 }
 
+func doCurveScan(host string, vers, cipherID uint16, ciphers []uint16) (supportedCurves []tls.CurveID, err error) {
+	allCurves := allCurvesIDs()
+	curves := make([]tls.CurveID, len(allCurves))
+	copy(curves, allCurves)
+	for len(curves) > 0 {
+		var curveIndex int
+		_, curveIndex, err = sayHello(host, []uint16{cipherID}, curves, vers)
+		if err != nil {
```
jacobhaven: This check will be pretty necessary everywhere `sayHello` is called
```
		if err == helloFailed {
			err = nil
			break
		} else if err != nil {
			return
		}
```

scan/tls_handshake.go
jacobhaven:
```
// helloFailed is a sentinel value to indicate a tls error was received while attempting
// this handshake. It's actual value is arbitrary and should not be displayed.
var helloFailed = errors.New("scan: tls handshake hello failed")
```

scan/tls_handshake.go
jkroll: This will shadow all other possibly legitimate errors possibly returned by conn.SayHello, no? Or are you saying that there aren't any legitimate errors that can happen there (or that the loop in `cipherSuiteScan` should break in that case regardless)? I guess one could put the sentinel value into the cf-tls helper files and use reflect.DeepEqual to see how to set it, such that other errors are still passed up to here.

jacobhaven: Yep, any error from SayHello is "illegitimate", i.e.: it shouldn't be displayed or stop the scan as it's generated by the server after (predictably) invalid Client Hellos. It just needs be be served as a sentinel the show this hello was bad in some way, but the scan should be continued with other hellos.

Another possibility is adding an `ok bool` return value or something. I'm ambivalent.

jkroll:
I've pushed changes here and in cf-tls that use this sentinel approach. I'd tried it before, but abandoned it for lack of a way to set the sentinel properly. Thanks to Jacob for pointing me at `reflect.DeepEqual`.

This code works for me, but you may want to test it locally before merging since I haven't tried to make Travis build from the right branch. Sorry it's taken so long to land - it took me a long time to realize that the scan was being terminated by an alert message and that the fallthrough error returned when computing the cipherIndex is meaningless.

scan/tls_handshake.go
jkroll: So you don't think there's value in logging any of the other errors that can happen during readHandshake? There are, for example, interesting errors that can be returned from  `conn.readRecord` and it might be worth getting them out to the user - shadowing them will make the tool less useful for debugging odd behavior. Or am I overthinking this?

jkroll: In particular, this would shadow the return of outbound connection errors returned from `conn.readHandshake` at https://github.com/cloudflare/cf-tls/blob/master/tls/conn.go#L777. I'll assume as you say that those aren't interesting, since probably such errors end up getting returned by `tcp.Dial`.

jkroll:
At this commit, I use @jacobhaven's sentinel approach with the sentinel entirely in cfssl/scan. This shadows some errors that can be returned from the tls package, but on reflection, I agree with Jacob that they're uninteresting or will be returned elsewhere.

jacobhaven:
LGTM, but `go line` (and thus Travis CI) are unhappy about the error value naming conventions. 
`error var helloFailedErr should have name of the form errFoo`, i.e. rename it to `errHelloFailed`, `git --amend` the commit, force push it and it should be fine (except coveralls probably complaining about a decrease in test coverage)

jkroll:
Done. I thought the whole tao of Go was that `gofmt` was meant to take care of all that stuff for you. Also, that naming convention seems backwards to me with the rest of the usual Go naming conventions, but never mind.

scan/tls_handshake.go
```diff
@@ -71,11 +124,15 @@ type cipherVersionList []cipherVersions
...
 	cvStrings := make([]string, len(cvList))
 	for i, c := range cvList {
-		versStrings := make([]string, len(c.versions))
-		for j, vers := range c.versions {
-			versStrings[j] = tls.Versions[vers]
+		versStrings := make([]string, len(c.data))
+		for j, d := range c.data {
+			curveStrings := make([]string, len(d.curves))
+			for k, c := range d.curves {
+				curveStrings[k] = fmt.Sprintf("%d", tls.Curves[c])
```
jacobhaven: This should just be `curveStrings[k] = tls.Curves[c]`, as it's already a string.

jkroll: Fixed. Somehow I thought it was a `tls.CurveID` but that was dumb of me. I think I am most surprised that `%d` is a valid format string for printf-ing strings.

jacobhaven:
Yay, Travis is happy. And and week later... *PR merged* and up in an ugly UI at https://cfssl.jhaven.me/scan.html

PR #226, opened 2015-05-29
Add a pkcs11key_stub for build tools.

jsha:
In Boulder, we're using godep to vendorize dependencies. I'm working on a change to include the pkcs11 dependencies as well. This requires a patched version of godep to support tags (https://github.com/tools/godep/pull/117).

Unfortunately, godep gets confused and gives the error:

```
godep: no buildable Go source files in /home/jsha/go/packages/src/github.com/cloudflare/cfssl/crypto/pkcs11key
godep: error loading dependencies
```

This error is simply fixed by adding this file.

PR #227, opened 2015-05-29
Add OCSP config and PKCS#11 OCSP support

jsha:


crypto/pkcs11key/pkcs11key_stub.go
```diff
@@ -0,0 +1,5 @@
+// Some Go tools will complain "no buildable Go source files in ..." because
```
kisom: If we're putting this file in here instead of in #226, the comment from that PR should be ported here.

jsha:
Updated to fix go vet issues, passes tests now.

jsha:
Ping on this?

kisom:
LGTM.

PR #229, opened 2015-05-30
Improvement: Adds new cli flag 'int-dir' to bundle cmd

dol:
The bundle command creates a special intermediate folder to store
the downloaded intermediate certificates.
The current implementation creates an 'intermediates' folder
at the current path, where cfssl is executed. This polutes the
file system with many unnecessary 'intermediates' folders.
This commit fixes the problem by adding a new flag 'int-dir' to
customize the intermediate folder path. The same flag is already
used for the HTTP server and inherits it's default value from
the config.

### DIFF
--- a/cli/bundle/bundle.go
+++ b/cli/bundle/bundle.go
@@ -15,18 +15,19 @@ var bundlerUsageText = `cfssl bundle -- create a certificate bundle that contain
 
 Usage of bundle:
 	- Bundle local certificate files
-        cfssl bundle -cert file [-ca-bundle file] [-int-bundle file] [-metadata file] [-key keyfile] [-flavor optimal|ubiquitous|force] [-password password]
+        cfssl bundle -cert file [-ca-bundle file] [-int-bundle file] [-int-dir dir] [-metadata file] [-key keyfile] [-flavor optimal|ubiquitous|force] [-password password]
 	- Bundle certificate from remote server.
-        cfssl bundle -domain domain_name [-ip ip_address] [-ca-bundle file] [-int-bundle file] [-metadata file]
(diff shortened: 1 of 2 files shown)


grittygrease:
LGTM

cli/bundle/bundle.go
```diff
@@ -15,18 +15,19 @@ var bundlerUsageText = `cfssl bundle -- create a certificate bundle that contain
...
 Flags:
 `
 
 // flags used by 'cfssl bundle'
-var bundlerFlags = []string{"cert", "key", "ca-bundle", "int-bundle", "flavor", "metadata", "domain", "ip", "password"}
+var bundlerFlags = []string{"cert", "key", "ca-bundle", "int-bundle", "flavor", "int-dir", "metadata", "domain", "ip", "password"}
 
 // bundlerMain is the main CLI of bundler functionality.
 func bundlerMain(args []string, c cli.Config) (err error) {
+	bundler.IntermediateStash = c.IntDir
```
lziest: I didn't see changes to cli.Config to add 'IntDir', is it left out?

dol: No, it was already there. Used primary for the 'serve' command. I took the 'serve' behavior of 'int-dir' and added it to the 'bundle' command.
https://github.com/cloudflare/cfssl/blob/master/cli/config.go#L71
https://github.com/cloudflare/cfssl/blob/master/cli/serve/serve.go#L114

lziest: I see, thanks!

dol:
Thx for the merge.
BTW: Thank you for the superb work on this tool.

PR #233, opened 2015-06-11
serve: Bundle web UI endpoint and additional cleanup of static serve

jacobhaven:
Pretty UI for Bundle (from remote). This doesn't support just uploading/pasting certificates yet.

Cleanup of CSS and Javascript, prettyfying code and pages.

Demo up here: https://cfssl.jhaven.me/bundle.html


grittygrease:
Good first step towards a comprehensive bundle UI.

jacobhaven:
In general, it might make sense to have all the pages (Index, Scan, Bundle, SelfSign) to actually be served as a single HTML file to side-step the lack of templates to avoid duplicated header/etc. code. I'm not sure the CSS/Javascript trickery needed to make this happen in a nice way.

jacobhaven:
Bundle specifically should have another form option of uploading/pasting a certificate (similar to the cloudflare.com UI) to be bundled.

grittygrease:
Can you remove the ".html" from each of the subpages? It seems old fashioned.

grittygrease:
@jacobhaven This seems to be failing in Travis, can you rebase?

jacobhaven:
There was actually a bug in the test where the server wasn't getting shut down (causing havoc with later tests). Everything should be good now.

cli/serve/serve.go
```diff
@@ -91,15 +94,24 @@ var v1Endpoints = map[string]func() (http.Handler, error){
...
+	}
+	return (&url.URL{Path: prefix + endpoint}).String()
+}
+
 // registerHandlers instantiates various handlers and associate them to corresponding endpoints.
 func registerHandlers() {
 	for path, getHandler := range v1Endpoints {
 		if path != "/" {
-			path = fmt.Sprintf("/api/v1/cfssl/%s", path)
+			path = "/api/v1/cfssl/" + path
```
kisom: Should this use the `v1APIPath` function defined above?

jacobhaven: Yes. I'll update that in some future PR...

PR #238, opened 2015-06-16
Add support for certificate policy qualifiers.

jsha:
Previously we could only add certificate policies, but not their related
qualifiers.

Fixes #230.

PR #239, opened 2015-06-17
Fix marshaling of OID

jsha:
In #238 I accidentally used %v with our custom oid, which lacks the correct String method. This PR casts to asn1.ObjectIdentifier to get the correct String method.


kisom:
That's my bad, I used `asn1.ObjectIdentifier` in the test code.

PR #240, opened 2015-06-19
scan: add scan to check if host is orange-clouded

jacobhaven:
An additional DNS scan that checks that all A and AAAA records point to CloudFlare IP addresses.

### DIFF
--- a/scan/connectivity.go
+++ b/scan/connectivity.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"net"
-	"net/http"
 	"strings"
 
 	"github.com/cloudflare/cf-tls/tls"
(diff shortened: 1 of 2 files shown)


scan/connectivity.go
```diff
@@ -50,10 +79,37 @@ func dnsLookupScan(host string) (grade Grade, output Output, err error) {
...
+		ip := net.ParseIP(addr)
+		for _, cfNet := range cloudflareNets {
+			if cfNet.Contains(ip) {
+				cfStatus[addr] = true
+				break
+			}
+		}
+		if !cfStatus[addr] {
+			cfStatus[addr] = false
+			grade = Bad
```
lziest: :+1:

lziest:
LGTM

jacobhaven:
What do we think? Do we want to download the IP ranges (from https://www.cloudflare.com/ips-v4 /v6) at init-time, or hardcode them, see the diff here: https://github.com/cloudflare/cfssl/commit/f6dae78e025b7077de69ec404de3ceeabfef9f88

grittygrease:
If the cloudflare.com/ips endpoint is unavailable, this test should be skipped or marked as N/A.

jacobhaven:
@grittygrease That seems the most reasonable. I made that change and also moved the setup outside of the global `init()` function so it's only called when the test is run.

scan/connectivity.go
```diff
@@ -50,10 +50,78 @@ func dnsLookupScan(host string) (grade Grade, output Output, err error) {
...
 	}
+
 	grade, output = Good, addrs
 	return
 }
 
+var cloudflareNets []*net.IPNet
+
+func initOnCloudFlareScan() error {
+	// Don't re-download ranges if we already have them.
```
jacobhaven: Note that we do retry if there had previously been an error while downloading.

scan/connectivity.go
```diff
@@ -50,10 +50,78 @@ func dnsLookupScan(host string) (grade Grade, output Output, err error) {
...
+	}
+	defer v6resp.Body.Close()
+
+	scanner := bufio.NewScanner(io.MultiReader(v4resp.Body, strings.NewReader("\n"), v6resp.Body))
+	for scanner.Scan() {
+		_, ipnet, err := net.ParseCIDR(scanner.Text())
+		if err != nil {
+			return fmt.Errorf("Couldn't parse CIDR range: %v", err)
+		}
+		cloudflareNets = append(cloudflareNets, ipnet)
```
jacobhaven: We can also end up with a partial stored cloudflareNets if an error is encountered after only some of the ranges have been added.

jacobhaven:
I've addressed those two concerns, and I'm happy with this PR.

PR #241, opened 2015-06-22
Emit a CSR when gencert creates new CA from key.

kisom:
Currently, when running `cfssl gencert -initca`, it doesn't emit a CSR with the certificate; this adds support for returning a CSR with the certificate. This allows a CA to later be signed by an offline CA.


grittygrease:
LGTM

jacobhaven:
LGTM

PR #245, opened 2015-06-24
Add IP whitelists to multiroot config files

kisom:
This pull request allows support for comma-separated lists of IP networks in the multiroot configuration file under the "nets" key. If a whitelist exists, signature requests will be checked against the
list, and only those that are whitelisted will be permitted to continue.


whitelist/example/example_whitelist.go~
```diff
@@ -0,0 +1,75 @@
+package main
```
grittygrease: This looks like a temp file.

grittygrease:
Can you include some valid/invalid IPv6 addresses and ranges in the testcode?

grittygrease:
Looks good.

PR #255, opened 2015-06-30
Set pkcs11 user pin in environment

oliof:
To hide the user pin from the process list, set it in the environment. That way, only users with access to `/proc/$(pidof cfssl)/environ` (typically root) can read the user pin. You can store the pin in a file accessible by root only and set the pin as follows:

    USER_PIN=$(cat /etc/cfssl/pin)
    $GOPATH/bin/cfssl ...

If USER_PIN is unset, the behavior is identical to the current one (i.e. you need to provide a pin on the command line).

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"flag"
+	"os"
 
 	"github.com/cloudflare/cfssl/config"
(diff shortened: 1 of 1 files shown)


kisom:
I like this idea, thanks for the PR.

PR #258, opened 2015-07-01
Add support for multiple remotes in API client

kisom:
This PR enables support for groups of CFSSL remote instances in the API client. Groups of servers are indistinguishable from the client's perspective from a single server. Groups are constructed from a specific strategy using the `NewGroup` function, or by passing a comma-separated list to `NewServer` to construct an ordered list of remotes. The change to `NewServer` provides a non-invasive mechanism for supporting a group of remotes using the strategy that we intend to use first.

The only strategy supported now is an ordered list, in which the list of hosts are attempted in sequence until one succeeds or none do with the goal of improving availability. Potentially as a part of CFSSL v1.2, a round-robin strategy should be implemented to provide load-balancing.

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/cloudflare/cfssl/api"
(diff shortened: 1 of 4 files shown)


jacobhaven:
It looks like making `AuthReq` and `Req` private has [broken signer/remote](https://github.com/cloudflare/cfssl/blob/master/signer/remote/remote.go#L86).

kisom:
The latest commit decouples `signer/remote` from the now-internal functions in the client, using only the `Remote` interface specified in the `api/client` package.

grittygrease:
LGTM

PR #260, opened 2015-07-02
Remove old PolicyStrings field.

jsha:
This was superseded by CertificatePolicies.

Looks like this was accidentally restored in a42b0f13. It causes a conflict when loading JSON because there are two fields named `policies`.


kisom:
Thanks!

grittygrease:
LGTM

PR #261, opened 2015-07-06
Return with exit code 1 on error.

kisom:
Right now, if `cfssl` or `cfssljson` encounters an error, they don't return an error code. This ensures that they do.


grittygrease:
Looks good.

lziest:
LGTM

PR #267, opened 2015-07-17
Improve the build process

ekristen:
This makes it very simple and easy to cross compile cfssl and the other cli tools for darwin, linux, and windows for i386 and amd64 and even arm on linux. 

Borrowed the concept from https://github.com/docker/machine

Also added BUILDING instructions file.


ekristen:
@kisom please take a look when you get a chance.

ekristen:
The first time your run the build script it'll take a bit longer because it builds the proper environment in the docker image, subsequent runs will almost immediately start the builds.

kisom:
For cross-compiled builds, I've been using [goxc](https://github.com/laher/goxc):

```
go get github.com/cloudflare/cfssl
go get github.com/laher/goxc
goxc -t
goxc github.com/cloudflare/cfssl/cmd/...
```

ekristen:
To each his own, but `goxc github.com/cloudflare/cfssl/cmd/...` requires that the code be in the github repo instead of making changes locally and building for testing. I'm sure there is a way to build goxc locally, just it's not documented anyway.  I just attempted to do it locally, and `goxc -t` ran for 10 minutes and I have a pretty beefy system and still didn't finish the build toolchain, so I stopped it.

With docker it takes a few minutes (depending on network connection) to download the golang docker image, to build the docker image the builds cfssl it takes about 21 seconds on my machine. Then another 20 seconds to build 7 different platforms, can be faster if the parallel builds is increased. Subsequent builds, do not need to download the docker image or build the cfssl-builder image, so its only ~20 seconds to build all the CLI tools. 

The best part is your workstation stays clean of all the deps.

At the very least I'd recommend someone update the documentation for building, I had to figure it out on my own, as I'm not a go expert.

kisom:
Forgot to note, once I get into work (where my work machine has Docker installed), I can test this out.

kisom:
LGTM.

PR #273, opened 2015-07-24
Default behaviour is to build with PKCS #11.

### LINKED ISSUE #262: Remove pkcs11 build tag?
Right now the pkcs11 code is hidden behind the `pkcs11` build tag, so it doesn't get built unless you pass `-tags pkcs11` when you build. This is causing some toolchain issues for Boulder: In particular, the stock Godeps doesn't understand build tags, so we need to make sure everyone who updates our Godeps uses a patched copy.

Is there a strong reason for using the build tag? If not, I'd like to make it so pkcs11 support is unconditionally built.

kisom:
- Instead of a `pkcs11` tag, there is now a `nopkcs11` tag; the
  default behaviour is now to build in PKCS #11.
- Update the Docker cross-compilation to build without PKCS #11
  support.
- Update documentation to note changes to default behaviour and
  ensure users can build without this support.
- Closes #262.


kisom:
Build failures are due to coveralls.io failure.

jsha:
Thanks so much for implementing this! It looks good to me.

lziest:
LGTM

lziest:
always tricked by the 'close' button. sorry.

PR #278, opened 2015-07-25
Make scans run concurrently

### LINKED ISSUE #236: Add concurrency to scan on multiple sites
When scanning a list multiple sites, scan cycles through the target sites sequentially. Any one scan that has network issues will hold up the rest of the scan. This should be done in parallel to speed up large scans.

taravancil:
Running scans sequentially takes too long, especially if any one scan
runs into network issues.

This change runs each scan in a separate goroutine, and returns
results/errors whenever an individual scan finishes.

Resolves #236

### DIFF
--- a/cli/scan/scan.go
+++ b/cli/scan/scan.go
@@ -3,9 +3,9 @@ package scan
 import (
 	"encoding/json"
 	"fmt"
-
 	"github.com/cloudflare/cfssl/cli"
 	"github.com/cloudflare/cfssl/scan"
+	"time"
 )
 
 var scanUsageText = `cfssl scan -- scan a host for issues
@@ -28,6 +28,9 @@ func scanMain(args []string, c cli.Config) (err error) {
 	if c.List {
 		printJSON(scan.Default)
 	} else {
+		hosts := 0
+		done := make(chan bool)
+
 		// Execute for each HOST argument given
 		for len(args) > 0 {
 			var host string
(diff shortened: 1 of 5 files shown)


cli/scan/scan.go
```diff
@@ -35,17 +43,29 @@ func scanMain(args []string, c cli.Config) (err error) {
 			if err != nil {
 				return
 			}
+			hosts++
 
-			fmt.Printf("Scanning %s...\n", host)
```
jacobhaven: I think it's still worth printing out the hosts being scanned, even they're just all printed at once.

scan/scan_common.go
```diff
@@ -132,39 +132,36 @@ type ScannerResult struct {
...
 
-	go func() {
 		for familyName, family := range fs {
 			if familyRegexp.MatchString(familyName) {
 				scannerResults := make(map[string]ScannerResult)
 
 				for scannerName, scanner := range family.Scanners {
 					if scannerRegexp.MatchString(scannerName) {
+
 						grade, output, err := scanner.Scan(host)
```
jacobhaven: I think the biggest win we can get is by running individual scans (or maybe Families) in parallel and streaming back the results.

It's much more common to be waiting on multiple scans on a single host than scans across multiple hosts (especially from API calls).

taravancil: I'll work on this tonight.

jacobhaven:
I tried to clarify my concerns in a comment on the issue: https://github.com/cloudflare/cfssl/issues/236#issuecomment-126084244

This host-concurrent scan is fine. But full Scan (or Family)-level concurrency would be much better.

grittygrease:
@taravancil @jacobhaven What's the status of this PR? The tests are currently failing.

scan/scan_common.go
```diff
@@ -132,71 +132,68 @@ type ScannerResult struct {
...
+				}(scannerName, scanner, scannerResults)
+			}
+
+			res := PackagedFamilyResult{
+				Result:     scannerResults,
+				FamilyName: familyName,
 			}
+
+			resChan <- res
+			<-done
```
jacobhaven: It's not clear to me what this `done` is doing. I think a [sync.WaitGroup](https://golang.org/pkg/sync/#WaitGroup) is more likely what you want.

jacobhaven:
@taravancil I've rebased this and refactored a lot of the concurrency code. If you give me commit access to [your fork](https://github.com/taravancil/cfssl/tree/concurrent-scan) I'll push the changes there to update this PR. Otherwise, I'll just need to make a new PR

jacobhaven:
You can see the changes here: https://github.com/cloudflare/cfssl/commit/3ec3b898c608273dd38a5f3c4cebc05cad8a333d

jacobhaven:
I'm moving this PR to #388 continue discussion there...

PR #283, opened 2015-07-27
scan: Fix supportedSKXSignatureAlgorithms not being reset to default.

jacobhaven:
This will remove the `unknown hash function` seen after having set sigAlgs that we don't actually support.


lziest:
LGTM, pending CI

PR #285, opened 2015-07-29
Add LogFatal level of logging that also calls `os.Exit(1)`

jacobhaven:
I also updated instances where `log.Critical` was immediately followed by `os.Exit` (there may be other more complex instances that could use an update.)


lziest:
LGTM

PR #286, opened 2015-07-29
Expiry validation

taravancil:
I've included changes to check that a cert is not valid for too long. 

How ought this be integrated into the ubiquity package? My thought is that in [CompareExpiryUbiquity](https://github.com/cloudflare/cfssl/blob/master/ubiquity/ubiquity_crypto.go#L137-152) we could first check `ValidExpiry()` for each cert, and return whenever one but not the other cert is valid for too long. Otherwise continue to compare based on which cert lasts longer. 

```
if helpers.ValidExpiry(t1) && !helpers.ValidExpiry(t2) {
    return 1
}
if !helpers.ValidExpiry(t1) helpers.ValidExpriy(t2) {
    return -1
}
// If both are valid or invalid, continue ranking by expiration date
```
Thoughts?

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -55,6 +55,44 @@ func ExpiryTime(chain []*x509.Certificate) *time.Time {
 	return &notAfter
 }
 
+// MonthsValid returns the number of months for which a certificate is valid.
+func MonthsValid(c *x509.Certificate) int {
+	issued := c.NotBefore
+	expiry := c.NotAfter
+	years := (expiry.Year() - issued.Year())
+	months := years*12 + int(expiry.Month()) - int(issued.Month())
+
+	// Round up if valid for less than a full month
+	if expiry.Day() > issued.Day() {
+		months++
+	}
+	return months
+}
+
+// ValidExpiry determines if a certificate is valid for an acceptable
+// length of time per the CA/Browser Forum baseline requirements.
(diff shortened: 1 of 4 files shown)


helpers/helpers.go
```diff
@@ -55,6 +55,44 @@ func ExpiryTime(chain []*x509.Certificate) *time.Time {
...
+	}
+	return months
+}
+
+// ValidExpiry determines if a certificate is valid for an acceptable
+// length of time per the CA/Browser Forum baseline requirements.
+// See https://cabforum.org/wp-content/uploads/CAB-Forum-BR-1.3.0.pdf
+func ValidExpiry(c *x509.Certificate) bool {
+	issued := c.NotBefore
+	jul2012 := time.Date(2012, time.July, 01, 0, 0, 0, 0, time.UTC)
```
kisom: If you subtract one nanosecond from these, you can elide the call to `Equal` below.

kisom: If these are moved outside the function, the function won't have to compute the values each time.

kisom:
I like this idea; as an implementation note, it seems to me that the check should run whenever the flavour isn't force, though.

helpers/helpers.go
taravancil: Good call.

taravancil:
Noted. I'll work on integrating it into the ubiquity check later.

kisom:
@taravancil I went ahead and fixed the build issues so we could get this in for the 1.1 release.

PR #287, opened 2015-08-03
Add API docs for authsign, sign, init_ca, info, newkey and newcert

lziest:
Plus, remove the redundant 'hostname' parameter for newcert endpoint


kisom:
Awesome, thanks!

PR #289, opened 2015-08-03
scan: clean up output from scan loop to remove skipped scans/families.

jacobhaven:


scan/scan_common.go
```diff
@@ -190,7 +193,7 @@ func (fs FamilySet) RunScans(host, family, scanner string, dur time.Duration) (m
 	select {
 	case <-timeout:
 		log.Warningf("scan: %s timed out after %v", host, dur)
-		return nil, nil
+		return familyResults, nil
```
jacobhaven: We might as well return the partially completed family results here.

jacobhaven:
@taravancil You might want to look at this and make sure it doesn't break any of your concurrent-scan work.

taravancil:
@jacobhaven did you see the [changes](https://github.com/cloudflare/cfssl/pull/278/files) I pushed this morning? Individual scans run concurrently now and the results are sent on a channel as they finish. The updates aren't compatible with your changes, so lmk what you think about my approach.

Also, the build is failing for some reason, so I need to look into that...

jacobhaven:
@taravancil That will probably be the more general solution. This was just to remove kruft from the current output.

taravancil:
LGTM!

kisom:
LGTM.

PR #291, opened 2015-08-04
Add print-defaults commands.

kisom:
There are only two configs here, but it's a starting point for adding any other default configuration strings desired.


lziest:
LGTM

PR #298, opened 2015-08-07
Fix multiscan output for SessionResumeScan

jacobhaven:
There was a bug where `grade` was being set rather than `g` inside the `multiscan` func. I also refactored, and fixed a bit of the error handling.


lziest:
LGTM

PR #301, opened 2015-08-10
Introduce a smaller docker image based on Alpine

dhiltgen:
This adds a Dockerfile based on Alpine linux, and collapses most of
the build steps into one RUN command that cleans up after itself.
The resulting image is ~38 MB compared to >500 MB for the primary
dev image.


kisom:
I like this. It'll take a day or so before I can test it, though.

kisom:
@dhiltgen I like the much smaller image size, thanks for providing this.

kisom:
LGTM.

PR #306, opened 2015-08-14
Remove useless -uselocal and add -nostatic to disable static hosting

### LINKED ISSUE #304: Make "static" serving optional
Recently there was a change that now requires you have a `static` box in order to run `cfssl serve` otherwise it panics with:
```
panic: could not locate box "static"
```
Can you make that optional? I don't want to serve any static files I just want to serve is `/newcert` and I have no need for any static files. Older versions worked perfectly for this. It makes it hard to distribute a single static binary built with `go build` since there's now a new requirement that you have a static folder of assets.

### LINKED ISSUE #305: -uselocal does nothing now
I have a `static` directory in the working directory but when I run cfssl with `-uselocal` it still panics with:
```
panic: could not locate box "static"
```
I grep'd the repo and it doesn't look like `config.uselocal` is used anymore. So we have to bundle the static files now?

fastest963:
Fixes #305
Fixes #304

### DIFF
--- a/cli/serve/serve.go
+++ b/cli/serve/serve.go
@@ -31,17 +31,18 @@ var serverUsageText = `cfssl serve -- set up a HTTP server handles CF SSL reques
 Usage of serve:
         cfssl serve [-address address] [-ca cert] [-ca-bundle bundle] \
                     [-ca-key key] [-int-bundle bundle] [-int-dir dir] [-port port] \
-                    [-metadata file] [-remote remote_host] [-config config] [-uselocal]
+                    [-metadata file] [-remote remote_host] [-config config]
 
 Flags:
 `
 
 // Flags used by 'cfssl serve'
-var serverFlags = []string{"address", "port", "ca", "ca-key", "ca-bundle", "int-bundle", "int-dir", "metadata", "remote", "config", "uselocal"}
+var serverFlags = []string{"address", "port", "ca", "ca-key", "ca-bundle", "int-bundle", "int-dir", "metadata", "remote", "config"}
 
 var (
-	conf cli.Config
-	s    signer.Signer
+	conf      cli.Config
+	s         signer.Signer
+	staticDir = "static"
 )
 
 var errBadSigner = errors.New("signer not initialized")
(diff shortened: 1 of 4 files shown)


cli/serve/serve.go
```diff
@@ -94,9 +95,23 @@ var v1Endpoints = map[string]func() (http.Handler, error){
...
+		order := []rice.LocateMethod{
+			rice.LocateEmbedded,
+			rice.LocateAppended,
+			rice.LocateFS,
+			rice.LocateWorkingDirectory,
+		}
+		c := &rice.Config{LocateOrder: order}
+		box, err := c.FindBox(staticDir)
+		if err != nil {
+			panic(err)
```
jacobhaven: This shouldn't panic. It should just return `nil, err`. This alone fixes the #304 issue, without the need for an extra argument.

fastest963: I agree I just didn't want to change the existing behavior. I will make it not panic.

cli/serve/serve.go
```diff
@@ -94,9 +95,23 @@ var v1Endpoints = map[string]func() (http.Handler, error){
 	"scaninfo": func() (http.Handler, error) {
 		return scan.NewInfoHandler(), nil
 	},
+}
 
+var staticEndpoints = map[string]func() (http.Handler, error){
 	"/": func() (http.Handler, error) {
-		return http.FileServer(rice.MustFindBox("static").HTTPBox()), nil
+		//the default rice order doesn't include working directory
```
jacobhaven: This wasn't an intended use case (if the cfssl source has been moved, it would really be ideal to [build the binary with any static file embedded](https://github.com/cloudflare/cfssl/blob/master/README.md#static-builds)). But, it does seem fine to add the working directory as a final fallback.

jacobhaven:
I agree `-uselocal` should have been removed when we switched to rice (it was just an oversight). I don't think we need a new argument though. The way the handlers are structured, we can gracefully not serve static files if they're not found, and we should just do that.

cli/serve/serve.go
jacobhaven: That's fair. `MustFindBox` wasn't really the correct behavior, so we should fix that. :)

fastest963:
@jacobhaven thanks for the quick review. I only added the new argument since I didn't know if the panic was intended and so I wanted to add an explicit way to opt out of the panic but still keep the *possibly* intended default case of panic. I've made the changes and removed both the flags. Thanks again.

fastest963:
Woops good catch on the docs. I updated again and removed the flags from the docs.

jacobhaven:
Alright, LGTM

PR #312, opened 2015-08-20
Truncate rather than round OCSP time.

jsha:
Without this, we generate OCSP responses that are not yet valid half the time.


kisom:
LGTM.

Issue #259, opened 2015-07-01
-pkcs11-token is misleading and ambiguous


oliof:
https://github.com/cloudflare/cfssl/blob/master/signer/pkcs11/pkcs11.go#L46
	`priv, err := pkcs11key.New(cfg.Module, cfg.Token, cfg.PIN, cfg.Label)`
calls https://github.com/cloudflare/cfssl/blob/master/crypto/pkcs11key/pkcs11key.go#L52 
   ` func New(module, slot, pin, privLabel string) (ps *PKCS11Key, err error)`

The command line flag `-pkcs11-token` is used to initialize `cfg.Token`. pkcs11key uses this value to look for a Slot Descriptor, and selects a slot accordingly.

There are several issues with this:

1. A token is not a slot; the command line flag is misnamed
2. Using a SlotDescriptor to select a token is not ideal, because SlotDescriptors are not unique. In testing, this led to CFSSL picking the wrong slot, as the current code simply uses the first slot that matches the descriptor. Instead of a slot descriptor, a slot id should be used.
3. The key is chosen by label. Different keys can have identical labels  

Enhancement suggestion:

cfssl/crypto/cfsslkey.New() should identify the key by combination of token label, key id, key label (in softhsm 1.3 though, neither key label nor key id are unique, so the risk of getting another key than intended when new keys are generated still persists. I'll try to find out whether this is a bug in softhsm or acccording to the PKCS#11 spec).

grittygrease:
@jsha Is this still applicable?

jsha:
Unfortunately, yes. Looking at the code, it looks like I forgot to update the CLI flags to be compatible with the config JSON renames I did. I think this doesn't show up as a build error because the flags get put into a `map[string]string` instead of a JSON struct. :-(

I can take responsibility for fixing this. What sort of timeline would you like it on?

jacobhaven:
`pkcs11-token` has been removed from CFSSL

PR #319, opened 2015-09-12
use slot ids in the cfssl pkcs11 api

jmhodges:
It was using TokenLabels solely to select slots but those can have duplicates on the same HSM. Instead, use slot IDs with them.

Found in https://github.com/letsencrypt/boulder/pull/784

### DIFF
--- a/crypto/pkcs11key/pkcs11key.go
+++ b/crypto/pkcs11key/pkcs11key.go
@@ -8,6 +8,7 @@ import (
 	"crypto"
 	"crypto/rsa"
 	"errors"
+	"fmt"
 	"io"
 	"math/big"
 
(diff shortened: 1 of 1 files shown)


jmhodges:
This is an API break, but a required one.

jmhodges:
Argh, sorry. I thought this patch had already been tested against your codebase instead of just ours.

Fixing.

kisom:
I think this will address #259, also.

jmhodges:
Okay, that test failure seems to be an artifact of the travis config.

Specially, since my code is coming from jmhodges/cfssl, instead of cloudflare/cfssl, the `go get  github.com/cloudflare/cfssl/...` can't work correctly.

jmhodges:
If someone could push my branch into the cloudflare version of this and check the build, that would be a great kindness.

kisom:
@jmhodges Pushed.

jmhodges:
The build error remains but I cannot reproduce it by running the tests locally with `go test ./...` in the cfssl repo.

jmhodges:
Ah, ha. I got repro with a `go install`. Okay.

Thank you for the help.

PR #321, opened 2015-09-15
Look up PKCS#11 keys by key label, regardless of slot

bifurcation:
In some PKCS#11 modules, slots are not distinguished by name, but it is possible to assign distinct names to private keys.  This PR updates the PKCS#11 key object handle this situation better:

* Each time the key is used, we search for it across all slots
* The slot description is used to filter slots in this search

### DIFF
--- a/crypto/pkcs11key/pkcs11key.go
+++ b/crypto/pkcs11key/pkcs11key.go
@@ -129,11 +129,14 @@ func (ps *PKCS11Key) Destroy() {
 }
 
 // Look up the token that contains the desired private key
-func (ps *PKCS11Key) openSession() (session pkcs11.SessionHandle, keyHandle pkcs11.ObjectHandle, err error) {
+func (ps *PKCS11Key) openSession() (pkcs11.SessionHandle, pkcs11.ObjectHandle, error) {
+	var emptySession pkcs11.SessionHandle
+	var emptyHandle pkcs11.ObjectHandle
+
 	// Find slot by description
 	slots, err := ps.module.GetSlotList(true)
 	if err != nil {
-		return
+		return emptySession, emptyHandle, err
 	}
 	for _, slot := range slots {
 		// If ps.slotDescription is provided, we will only check matching slots
@@ -147,7 +150,7 @@ func (ps *PKCS11Key) openSession() (session pkcs11.SessionHandle, keyHandle pkcs
 		}
 
 		// Open session
-		session, err = ps.module.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION)
+		session, err := ps.module.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION)
 		if err != nil {
 			continue
 		}
@@ -179,14 +182,11 @@ func (ps *PKCS11Key) openSession() (session pkcs11.SessionHandle, keyHandle pkcs
 		}
 
 		if len(objs) > 0 {
-			// session and err have already been set above
-			keyHandle = objs[0]
-			return
+			return session, objs[0], nil
 		}
 	}
 
-	err = errors.New("slot not found")
-	return
+	return emptySession, emptyHandle, errors.New("slot not found")
 }
 
 func (ps *PKCS11Key) closeSession(session pkcs11.SessionHandle) {
--- a/crypto/pkcs11key/pkcs11key.go
+++ b/crypto/pkcs11key/pkcs11key.go
@@ -51,7 +51,15 @@ type PKCS11Key struct {
 }
 
 // New instantiates a new handle to a PKCS #11-backed key.
+// The slotDescription and tokenLabel parameters are optional (i.e., if
+// they are set to "", they will be ignored).  However, at least one of
+// them must be set.
 func New(module, slotDescription, tokenLabel, pin, privLabel string) (ps *PKCS11Key, err error) {
+	if slotDescription == "" && tokenLabel == "" {
+		err = errors.New("either slotDescription or tokenLabel must be specified")
+		return
+	}
+
 	// Set up a new pkcs11 object and initialize it
 	p := pkcs11.New(module)
 	if p == nil {
@@ -146,7 +154,7 @@ func (ps *PKCS11Key) openSession() (pkcs11.SessionHandle, pkcs11.ObjectHandle, e
 		if ps.slotDescription != "" {
 			slotInfo, err := ps.module.GetSlotInfo(slot)
 			if err != nil {
-				continue
+				return emptySession, emptyHandle, err
 			}
 			if slotInfo.SlotDescription != ps.slotDescription {
 				continue
@@ -157,7 +165,7 @@ func (ps *PKCS11Key) openSession() (pkcs11.SessionHandle, pkcs11.ObjectHandle, e
 		if ps.tokenLabel != "" {
 			tokenInfo, err := ps.module.GetTokenInfo(slot)
 			if err != nil {
-				continue
+				return emptySession, emptyHandle, err
 			}
 			if tokenInfo.Label != ps.tokenLabel {
 				continue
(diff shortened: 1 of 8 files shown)


kisom:
@bifurcation Looks like go vet failed; `%u` on line 144 of `crypto/pkcs11key/pkcs11key.go` isn't recognised.

kisom:
Actually a number of places where `%u` is used:

```
2.24s$ go vet github.com/cloudflare/cfssl/...
crypto/pkcs11key/pkcs11key.go:144: unrecognized printf verb 'u'
crypto/pkcs11key/pkcs11key.go:148: unrecognized printf verb 'u'
crypto/pkcs11key/pkcs11key.go:155: unrecognized printf verb 'u'
crypto/pkcs11key/pkcs11key.go:161: unrecognized printf verb 'u'
crypto/pkcs11key/pkcs11key.go:172: unrecognized printf verb 'u'
crypto/pkcs11key/pkcs11key.go:179: unrecognized printf verb 'u'
crypto/pkcs11key/pkcs11key.go:184: unrecognized printf verb 'u'
```

crypto/pkcs11key/pkcs11key.go
```diff
@@ -158,31 +129,67 @@ func (ps *PKCS11Key) Destroy() {
 	}
 }
 
-func (ps *PKCS11Key) openSession() (session pkcs11.SessionHandle, err error) {
+// Look up the token that contains the desired private key
+func (ps *PKCS11Key) openSession() (session pkcs11.SessionHandle, keyHandle pkcs11.ObjectHandle, err error) {
```
jsha: This function is now too long to used named returns, please use positional returns instead for clarity, per Go style.

bifurcation:
Thanks, @kisom, forgot to remove some printfs from debugging.  `go vet` looks good locally now.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -158,31 +129,67 @@ func (ps *PKCS11Key) Destroy() {
...
 	// Find slot by description
 	slots, err := ps.module.GetSlotList(true)
 	if err != nil {
 		return
 	}
 	for _, slot := range slots {
-		slotInfo, err := ps.module.GetSlotInfo(slot)
+		// If ps.slotDescription is provided, we will only check matching slots
+		slotInfo, err2 := ps.module.GetSlotInfo(slot)
```
jsha: Why err2 here?

crypto/pkcs11key/pkcs11key.go
```diff
@@ -158,31 +129,67 @@ func (ps *PKCS11Key) Destroy() {
...
+		if err2 != nil {
+			err = err2
+			fmt.Printf(">>> [%u] Couldn't get slot description\n", slot)
```
jsha: Remove these debugging statements.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -158,31 +129,67 @@ func (ps *PKCS11Key) Destroy() {
...
-			// Open session
-			session, err = ps.module.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION)
-			if err != nil {
-				return session, err
-			}
+		// Login
+		if err = ps.module.Login(session, pkcs11.CKU_USER, ps.pin); err != nil {
+			fmt.Printf(">>> [%u] Couldn't log in with pin [%s] [%v]\n", slot, ps.pin, err)
+			ps.closeSession(session)
+			continue
```
jsha: I think when we get a failure to log in, we shouldn't proceed to try the next slot. If someone has multiple slots with the same description, this would log into each slot until it finds the right one. This would very quickly lock out the earliest slots for PIN failures.

I think this is an argument for why it's useful to add the token label as an additional selector. If we use private key label as a selector, we have to log in before we can check for a match. But if we also use token label, we can avoid spurious logins to tokens we didn't want.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -158,31 +129,67 @@ func (ps *PKCS11Key) Destroy() {
...
-				return session, err
-			}
+		// See if the private key is present
+		template := []*pkcs11.Attribute{
+			pkcs11.NewAttribute(pkcs11.CKA_CLASS, pkcs11.CKO_PRIVATE_KEY),
+			pkcs11.NewAttribute(pkcs11.CKA_LABEL, ps.privateKeyLabel),
+		}
+		if err = ps.module.FindObjectsInit(session, template); err != nil {
+			fmt.Printf(">>> [%u] Couldn't FindObjectsInit\n", slot)
+			ps.closeSession(session)
```
jsha: The repetition of `ps.closeSession()` suggests a refactor: If you put the inner part of this loop in its own function, you can follow `openSession` with `defer ps.closeSession()`, which will be shorter and more reliable.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -158,31 +128,60 @@ func (ps *PKCS11Key) Destroy() {
...
+		// See if the private key is present
+		template := []*pkcs11.Attribute{
+			pkcs11.NewAttribute(pkcs11.CKA_CLASS, pkcs11.CKO_PRIVATE_KEY),
+			pkcs11.NewAttribute(pkcs11.CKA_LABEL, ps.privateKeyLabel),
+		}
+		if err = ps.module.FindObjectsInit(session, template); err != nil {
+			ps.closeSession(session)
+			continue
+		}
+		objs, _, err2 := ps.module.FindObjects(session, 2)
```
jsha: Why err2 here?

kisom:
I'll be able to actually review this in a few hours. Thanks @jsha for the review, too.

crypto/pkcs11key/pkcs11key.go
bifurcation: Shadowing, but now not necessary.

crypto/pkcs11key/pkcs11key.go
jsha: To expand on this a little: I think whatever approach we take should (a) attempt `Login` at most once per `openSession`, and (b) call `openSession` once during `New` and immediately return error if it fails. That way we ensure that we fail fast on any `Login`-related error and minimize the chance of lockouts.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -34,22 +34,24 @@ type PKCS11Key struct {
...
-	// We will automatically search for this in the slot list.
-	slotDescription string
-
 	// The PIN to be used to log in to the device
 	pin string
 
+	// The name of the slot to be used, or "" to use any slot
+	slotDescription string
+
+	// The label of the token to be used, or "" to use any token
```
jsha: I'm fine with slotDescription being optional, but I think at least tokenLabel should be mandatory. Otherwise it's easy to accidentally leave out the token label, attempt to log into the wrong token, fail repeatedly, and lock that incorrect token.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -158,36 +131,76 @@ func (ps *PKCS11Key) Destroy() {
...
 	}
 	for _, slot := range slots {
-		slotInfo, err := ps.module.GetSlotInfo(slot)
-		if err != nil {
-			continue
+		// If ps.slotDescription is provided, only check matching slots
+		if ps.slotDescription != "" {
+			slotInfo, err := ps.module.GetSlotInfo(slot)
+			if err != nil {
+				continue
```
jsha: There are a number of places in this loop where you continue if there's an error, but I think it would be more appropriate in each case to exit with that error.

signer/pkcs11/pkcs11.go
```diff
@@ -18,6 +18,7 @@ import (
 // #11 key.
 type Config struct {
 	Module string
+	Slot   string
```
jsha: So long as we are breaking the API, let's take the opportunity to give more descriptive names to these fields. Suggestions: SlotDescription, TokenLabel, and Label -> PrivateKeyLabel.

bifurcation:
@jsha Now I'm confused.  Is this being pursued here or in your branch?  Happy to pass this to you if you want to get it finished.

jsha:
@bifurcation: I wound up pulling this into a larger branch I was working on for concurrency problems. Sorry I neglected to let you know! I'll take it over from here, thanks for getting it this far.

bifurcation:
@jsha No problem.  I consider the ball passed.

Issue #323, opened 2015-09-15
Scan should support both domain and IP as inputs


grittygrease:
In order to scan hosts that do not have DNS entries but expect SNI, both domain and ip should be inputs for scan.

If 'ip' is missing, the IP should be obtained via looking up the domain with DNS. If both 'ip' and 'domain' are present, the IP should be used directly with the Domain being sent as SNI.

This mirrors the behavior in bundle, i.e.:
cfssl bundle -domain domain_name [-ip ip_address]

PR #324, opened 2015-09-15
Add optional IP parameter to scan CLI and API

jacobhaven:
This splits the `host` argument into `addr` (with optional fixed IP and Port) and `hostname`.

Note that none of these scans send HTTP requests, so there is no code to set the Host header (It will simply be `req.Host = hostname`).

The scans that don't make sense with a fixed address (i.e. DNS lookup) just use hostname


jacobhaven:
Addresses Issue #323 and #324

kisom:
LGTM.

PR #327, opened 2015-09-16
Minor cfssl improvements

kayrus:
- Fixed CentOS package name in README.md and BUILDING.md.
- If we have defaults at `cli/printdefault/defaults.go` - let's use them. Set default algo to rsa.
- Renamed `www` profile name to `server`. I suppose this is more proper name.
- Basically you can use `"hosts":[""]` json and it will pass validation. But if you would like to set hosts through `-hostname` option only when you omit `hosts` - this validation harms.

Don't have much experience with go. But I had an idea to validate "hosts" and "hostname" in proper way. Right now it is possible to pass `"hosts":["","",""]` or `-hostname=",,,,"`. Probably we have to fix `SplitHosts` and `OverrideHosts` functions. I've tried to implement something like this:

``` go
regexp.MustCompile("[ ,]+").Split(strings.TrimSpace(hostList),-1)
```

but it didn't work with ",     1 , 2 , 4  ,   5  ,   , dfdfdsfs  ,   , , ,  " strings

### DIFF
--- a/csr/csr.go
+++ b/csr/csr.go
@@ -49,7 +49,7 @@ type BasicKeyRequest struct {
 
 // NewBasicKeyRequest returns a default BasicKeyRequest.
 func NewBasicKeyRequest() *BasicKeyRequest {
-	return &BasicKeyRequest{"ecdsa", curveP256}
+	return &BasicKeyRequest{"rsa", 2048}
 }
 
 // Algo returns the requested key algorithm represented as a string.
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -31,7 +31,7 @@ The requirements to build without Docker are:
 3. A properly configured GOPATH
 4. The default behaviour is to build with PKCS #11, which  requires the
    `gcc` compiler and the libtool development library and header files. On
-   Ubuntu, this is `libltdl-dev`.
+   Ubuntu, this is `libltdl-dev`. On Centos/RHEL, this is `libtool-ltdl-devel`.
 
 To build with PKCS #11 support, run:
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Installation requires a
 properly set `GOPATH`.  The default behaviour is to build with PKCS
 \#11, which requires the `gcc` compiler and the libtool development
 library and header files. On Ubuntu, this is
-`libltdl-dev`. On Centos/RHEL, this is 'libtool' and 'libtool-ltdl'.
+`libltdl-dev`. On Centos/RHEL, this is `libtool-ltdl-devel`.
 If these are not installed, you can pass `-tags nopkcs11` to the below
 go get commands.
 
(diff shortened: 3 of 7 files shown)


csr/csr.go
```diff
@@ -49,7 +49,7 @@ type BasicKeyRequest struct {
 
 // NewBasicKeyRequest returns a default BasicKeyRequest.
 func NewBasicKeyRequest() *BasicKeyRequest {
-	return &BasicKeyRequest{"ecdsa", curveP256}
+	return &BasicKeyRequest{"rsa", 2048}
```
kisom: I'd still prefer to keep the ECDSA default; the RSA is provided in print-defaults to make it clear how to override the algorithm.

kayrus: @kisom That should be mentioned somehow. Don't know where exactly, but it should be accessible in cfssl binary.

As for the "hosts" and "hostname" validation, can you suggest one-liner? Or should I use simple loop?

kisom: @kayrus I think a loop that appends if the trimmed string length is non-zero would probably work best.

As for the ECDSA default, I'll add a new documentation file discussing certificate requests in CFSSL. Do you think that would be a good place for them?

kayrus: I still think that if you have "dafaults" command, it should display real defaults

kayrus:
Another bug I've just discovered. cfssl doesn't use "expiry" which is defined in "default" section of ca-config.json. For example:
```json
{
    "signing": {
        "default": {
            "expiry": "43800h"
        },
        "profiles": {
            "server": {
                "usages": [
                    "signing",
                    "key encipherment",
                    "server auth"
                ]
            },
            "client": {
                "usages": [
                    "signing",
                    "key encipherment",
                    "client auth"
                ]
            }
        }
    }
}
```

cfssl will complain: `Failed to load config file: {"code":5200,"message":"empty expiry string"}Failed to parse input: unexpected end of JSON input`

I suppose cfssl should read default values if it can not find them in profiles.

grittygrease:
Waiting for @kisom to return from vacation to resolve.

lziest:
orphaned PR

devisions:
Related to @kayrus aforementioned issue, I can confirm that even in the current version the default `expiry` is not considered. Therefore, what still works is indeed to add it to all profiles. Below the working example (letf it also in the `default` section although it's not relevant for now).

By _current version_ I mean:
- installed using `go get github.com/cloudflare/cfssl/cmd/cfssl@v1.5.0`
- and the reported version is `dev`:
  ```shell
  $ cfssl version
  Version: dev
  Runtime: go1.15.5
  ```

`ca-config.json` file
```json
{
  "signing": {
    "default": {
      "expiry": "8760h"
    },
    "profiles": {
      "server": {
        "expiry": "8760h",
        "usages": ["signing", "key encipherment", "server auth"]
      },
      "client": {
        "expiry": "8760h",
        "usages": ["signing", "key encipherment", "client auth"]
      }
    }
  }
}
```

PR #328, opened 2015-09-16
Allow bundled opaque key to be part of bundle output

lziest:
as long as the opaque key implements fmt.Stringer interface


grittygrease:
LGTM

Issue #329, opened 2015-09-17
pkcs11key.Sign is not thread safe


jsha:
Currently, pkcs11key opens a new session for each `Sign` call, does a `Login, SignInit, Sign,` and `Logout` on the token. This is inefficient (each login requires a key unwrap), but it's also a problem if `Sign` is ever called from two threads simultaneously.

According to ftp://ftp.rsasecurity.com/pub/pkcs/pkcs-11/v2-30/pkcs-11v2-30b-d6.pdf, `Login` and `Logout` affect the whole application, not just specific sessions. So if two threads call `Sign`, it's possible that they both `Login` succesfully, but then thread 1 calls `Logout` before thread 2 calls `SignInit`. The result: by the time thread 2 calls `SignInit` the application is already logged out of the token, so `SignInit` fails with `CKR_OBJECT_HANDLE_INVALID`.

We're working on a patch for this, along with a few other concurrency and performance fixes.

Issue #325, opened 2015-09-15
pkcs11key.New modifies contexts that are global


jmhodges:
`pkcs11key.New` calls `github.com/miekg/pkcs11.New` which is a wrapper around a `C.New` cgo function defined in that same package which use `lt_dlopen`. In `pkcs11key.New`, there is also a `pkcs11.Initialize` function used that wraps around a  `C.C_Initialize` function also defined in the pkcs11 package. These functions are meant to only be called once for a given module string (that is, the path to a shared library) and will throw errors if called multiple times.

If you have multiple keys you want to operate on, or are just testing things, you'll get Bad Behavior when calling pkcs11.New because it's operating, unknowingly, on global state.

Similarly, the `*PKCS11Key.Destroy` method causes the process state to go bad because it unloads the shared library entirely from the process.

Instead, `pkcs11key.New` ought to take a `*pkcs11.Ctx` object as its first argument instead of the module string, and the `Destroy` method on `PKCS11Key` needs to disappear.

Issue #322, opened 2015-09-15
Performance: multiple, persistent session support for pkcs11key


jsha:
Right now, pkcs11key opens a session and logs in for every Sign operation. For best performance, it's better to keep some number of sessions alive long-term. Per discussion at https://github.com/letsencrypt/boulder/pull/795, common practice is to have a thread pool with one session per thread.

I think, using channels and goroutines, we can achieve this without changing the Sign API to be asynchronous. We would have Sign send a message on a channel that is consumed by one of the goroutines holding an open session, then block on receiving a response on another channel. Possibly the `chan chan` pattern is useful for this: https://tleyden.github.io/blog/2013/11/23/understanding-chan-chans-in-go/.

I'm planning to do this work along with @jmhodges for Let's Encrypt, but I wanted to file an issue here for tracking, and to make sure this sounds like a good plan to you.

jacobhaven:
@jsha It's a bit problematic introducing a blocking backend to the synchronous Sign call. If it times out, the sign will just fail.

We have been looking into support asynchronous signers ([sketched out here](https://github.com/cloudflare/cfssl/commit/495b3ea848e31a780e7d36b976bb6c35d82e8088)), but this will require a bit more work: Generation of unique asynchronous callback tokens, a thread-safe map from these tokens to `RetryError`, and an API that returns these tokens and another that accepts them to retry the sign operation (calling `RetryError#Retry()`)

jsha:
@jacobhaven: To be clear, this would not be a remote backend, but would be multiple threads of the same process, communicating with Go channels. I don't think there's a risk of timeouts there. There is a risk that the processing thread might have an error but fail to report that error on the response channel, but I think we can avoid that with careful programming and use of `defer`.

Out of curiosity, what is your use case for Asynchronous Signers?

jacobhaven:
@jsha That seems reasonable then, as long as you can rely on your PKCS11 signer to return in a reasonable time.

Asynchronous signer's would be largely useful as a primitive in communicating with external CAs (which have natively asynchronous APIs).

PR #330, opened 2015-09-17
Revamp pkcs11key

jsha:
Fixes https://github.com/cloudflare/cfssl/issues/329
Fixes https://github.com/cloudflare/cfssl/issues/325
Fixes https://github.com/cloudflare/cfssl/issues/322
Incorporates https://github.com/cloudflare/cfssl/pull/321

pkcs11key used to log in and log out for each Sign request. This created a
performance problem, because logins are slow. It also created a concurrency
problem.

Logged in / logged out state is a property of a PKCS#11 application, not a
session. So it was possible for two Sign operations on separate threads to log
in at the same time, and for one Sign operation to finish and call Logout before
the other Sign operation called SignInit. This would result in a
CKR_OBJECT_HANDLE_INVALID error.

To fix both of these issues, I made the session a persistent property of a
PKCS11Key, and protected it with a mutex. Now each PKCS11Key maintains a single
session throughout its lifetime, increasing signing performance.

Also, some of the field names for PKCS11Key were incorrect, and there was no
provision to select by token label, only slot label. In general, token label is
more unique, and the better choice. This change incorporates the elements of
https://github.com/cloudflare/cfssl/pull/321 fixing that. That is, it renames
Token to SlotDescription, Label to PrivateKeyLabel, and adds a new field
TokenLabel. Note for people updating configs: In your configs, don't rename
Token to TokenLabel. The value you currently have in Token is actually a
SlotDescription, and should be named such. You should add a field named
TokenLabel, whose value you can find with pkcs11-tool --list-slots.

I also changed where the PKCS11Key Config object is defined, to make it more
straightforward to share between different components, and added object label
support to pkcs11uri, so private keys can be selected by label.

### DIFF
--- a/crypto/pkcs11key/key_pool.go
+++ b/crypto/pkcs11key/key_pool.go
@@ -55,6 +55,9 @@ func (p *PKCS11KeyPool) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts)
 	return instance.Sign(rand, msg, opts)
 }
 
+// Public returns the public key of any one of the signers in the pool. Since
+// they were all created with the same arguments, the public key should be the
+// same for each one.
 func (p *PKCS11KeyPool) Public() crypto.PublicKey {
 	instance := p.get()
 	defer p.put(instance)
--- a/crypto/pkcs11key/key_pool.go
+++ b/crypto/pkcs11key/key_pool.go
@@ -7,17 +7,17 @@ import (
 	"fmt"
 )
 
-// PKCS11KeyPool is a pool of PKCS11Keys suitable for high performance parallel
+// Pool is a pool of PKCS11Keys suitable for high performance parallel
 // work. PKCS11Key on its own is suitable for multi-threaded use because it has
 // built-in locking, but one PKCS11Key can have at most one operation inflight
 // at a time. If you are using an HSM that supports multiple sessions, you may
-// want to use a PKCS11KeyPool instead, which contains multiple signers.
-// PKCS11KeyPool satisfies the Signer interface just as PKCS11Key does, and
+// want to use a Pool instead, which contains multiple signers.
+// Pool satisfies the Signer interface just as PKCS11Key does, and
 // farms out work to multiple sessions under the hood. This assumes you are
 // calling Sign from multiple goroutines (as would be common in an RPC or HTTP
 // environment). If you only call Sign from a single goroutine, you will only
 // ever get single-session performance.
-type PKCS11KeyPool struct {
+type Pool struct {
 	// This slice acts more or less like a concurrent stack. Keys are popped off
 	// the top for use, and then pushed back on when they are no longer in use.
 	signers    []*PKCS11Key
@@ -28,7 +28,7 @@ type PKCS11KeyPool struct {
 	cond    *sync.Cond
 }
 
-func (p *PKCS11KeyPool) get() *PKCS11Key {
+func (p *Pool) get() *PKCS11Key {
 	p.cond.L.Lock()
 	for len(p.signers) == 0 {
 		p.cond.Wait()
@@ -40,7 +40,7 @@ func (p *PKCS11KeyPool) get() *PKCS11Key {
 	return instance
 }
 
-func (p *PKCS11KeyPool) put(instance *PKCS11Key) {
+func (p *Pool) put(instance *PKCS11Key) {
 	p.cond.L.Lock()
 	p.signers = append(p.signers, instance)
 	p.cond.Signal()
@@ -49,7 +49,7 @@ func (p *PKCS11KeyPool) put(instance *PKCS11Key) {
 
 // Sign performs a signature using an available PKCS #11 key. If there is no key
 // available, it blocks until there is.
-func (p *PKCS11KeyPool) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) ([]byte, error) {
+func (p *Pool) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) ([]byte, error) {
 	instance := p.get()
 	defer p.put(instance)
 	return instance.Sign(rand, msg, opts)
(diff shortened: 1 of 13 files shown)


crypto/pkcs11key/pkcs11key.go
```diff
@@ -128,71 +208,78 @@ func New(module, slot, pin, privLabel string) (ps *PKCS11Key, err error) {
...
+			slotInfo, err := ps.module.GetSlotInfo(slot)
 			if err != nil {
-				return session, err
+				return noSession, err
 			}
-
-			// Login
-			if err = ps.module.Login(session, pkcs11.CKU_USER, ps.pin); err != nil {
-				return session, err
+			if slotInfo.SlotDescription != ps.slotDescription {
```
jmhodges: Isn't slot description just some made up stuff from the manufacturer and not ever guaranteed to be unique? This seems like an unnecessary complication for something that won't be useful for folks. They'll just think they need to have it when they don't.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -128,71 +208,78 @@ func New(module, slot, pin, privLabel string) (ps *PKCS11Key, err error) {
...
-// The idiomatic way to do this (assuming no need for a long-lived
-// signer) is as follows:
-//
-//   ps, err := NewPKCS11Signer(...)
-//   if err != nil { ... }
-//   defer ps.Destroy()
+
+// Destroy tears down a PKCS11Key by closing the session. It should be
+// called before the key gets GC'ed, to avoid leaving dangling sessions.
+// NOTE: We do not want to call module.Logout here. module.Logout applies
```
jmhodges: Maybe put this not in the doc comments, but inside the function for implementers coming later. It's a little confusing.

crypto/pkcs11key/pkcs11key.go
```diff
@@ -128,71 +204,67 @@ func New(module, slot, pin, privLabel string) (ps *PKCS11Key, err error) {
...
-	}
 
-	err = errors.New("slot not found")
-	return
-}
+		// Login
+		// Note: Logged-in status is application-wide, not per session. But in
+		// practice it appears to be okay to login to a token multiple times with the same
+		// credentials.
+		if err = ps.module.Login(session, pkcs11.CKU_USER, ps.pin); err != nil {
```
jcjones: I've done this same thing before with other PKCS11 modules and it's worked fine.

jcjones:
I agree with the decision to remove SlotDescription, for what it's worth.

I'm only qualified to review the changes to `crypto/pkcs11key/pkcs11key.go`, but it looks like you gents have done a good job on it. The new design's mutex layout should even work alright for Encrypt operations, should that ever become useful to CFSSL.

crypto/pkcs11key/key_pool.go
```diff
@@ -0,0 +1,97 @@
...
+	signers := make([]*PKCS11Key, n)
+	for i := 0; i < n; i++ {
+		signers[i], err = New(modulePath, tokenLabel, pin, privateKeyLabel)
+		// If any of the signers fail, exit early. This could be, e.g., a bad PIN,
+		// and we want to make sure not to lock the token.
+		if err != nil {
+			for j := 0; j < i; j++ {
+				signers[j].Destroy()
+			}
+			return nil, fmt.Errorf("Problem making PKCS11Key: %s", err)
```
kisom: As a style thing, this error should be "pkcs11key: problem making PKCS11Key".

crypto/pkcs11key/pkcs11key.go
```diff
@@ -30,13 +34,29 @@ var hashPrefixes = map[crypto.Hash][]byte{
...
+// Each one will have its own session and can be used concurrently. Note that
+// some smartcards like the Yubikey Neo do not support multiple simultaneous
+// sessions and will error out on creation of the second PKCS11Key object.
+//
+// Note: For parallel usage, it is *highly* recommended that you create all your
+// PKCS11Key objects serially, on your main thread, checking for errors each
+// time, and then farm them out for use by different goroutines. If you fail to
+// do this, your application may attempt to login repeatedly with an incorrect
+// PIN, locking the PKCS#11 token.
 type PKCS11Key struct {
```
kisom: While we're changing this, would it make sense to rename this to just `Key`?

jsha: Yes, good idea! One of my worries with this change was that we were changing semantics (slot description vs token label) without creating a corresponding syntax change that would break builds and warn people to take a second look.

kisom: One of the common things we've heard about the PKCS #11 support is that
it's a confusing; anything we can do now to make it easier to understand
I'll support. I believe Let's Encrypt is the major PKCS #11 user, so if
these changes are productive for you all, I think that they will work
well for other people.

kisom:
@jsha It would be useful to have some PKCS11 docs in the top-level docs dir. Is that something that could be included with this PR? It's not strictly necessary, but I think it would be useful for people to get started.

jsha:
@kisom: Sure, will do.

jsha:
Okay, added docs.

kisom:
Tested locally with a yubikey neo. LGTM.

jsha:
@kisom: Thanks! This works great with my SoftHSM instance. I had some trouble with my Yubikey, which I tracked to the CKA_ALWAYS_AUTHENTICATE attribute on my private key object. Can you run this command and tell me if the Private Key Object has `Access: always authenticate` ?

```
pkcs11-tool --module /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so  --login --list-objects    --pin <pin>
```

The problem I was having was that each session would succeed at signing once, and then fail on the second sign attempt. I can see now why the original version of this code did a login and logout on each sign attempt! It appears to be necessary for Yubikeys.

I just pushed an update to this branch that checks for the CKA_ALWAYS_AUTHENTICATE attribute and fails if it's present. However, this probably breaks usage with most Yubikeys. Try running the pkcs11_bench_test against your Yubikey and let me know what the result is?

My next steps is to see if I can clear the CKA_ALWAYS_AUTHENTICATE attribute, in which case we wouldn't need any special casing for Yubikeys. If I can't clear the attribute, and support for Yubikeys is important, we'll have to add some special case code that will do a logout/login after each signature when CKA_ALWAYS_AUTHENTICATE is present. I'd rather avoid the special casing because it will be messy: we would also have to make sure we disallow creation of multiple pkcs11key.Key's, because their behavior would be undefined. So let me know if you consider Yubikey support optional and it'll save me some work!

jsha:
Actually, I went ahead and implemented the login/logout when CKA_ALWAYS_AUTHENTICATE is true, so Yubikeys should work properly.

kisom:
@jsha I'll have some time to do more extensive testing over the weekend. Thanks again for this PR!

jsha:
@kisom: Thanks! Let me know what you find.

kisom:
After jumping through a number of hoops figuring out how to stand up a new CA on a Neo-S, and a much more pleasant (in relative PKCS11 terms) time with SoftHSM, this LGTM.

jsha:
Excellent, thanks for the review & merge!

PR #331, opened 2015-09-18
Make Travis work better for external committers.

jsha:
The way Go handles multiple packages within the same repo makes it challenging
to submit branches as an external committer and rely on one's own Travis
instance for checking. This is a fix we've been using in Boulder.


lziest:
cool thanks!

PR #332, opened 2015-09-18
Additional CSR functionality

kisom:
The `csr` package currently only operates on self-contained
CertificateRequest strucutres. There are times, however, where it is
useful to generate a CSR with an existing key or to re-sign a CSR with
an existing private key. This is mostly useful for client tools that
want to make use of the CFSSL packages.

This pull request provides two new functions:

+ Regenerate, which re-signs a CSR with a private key, and

+ Generate, which uses a CertificateRequest to build a CSR that is
  signed with a provided private key.    

To support this, some additional functions were added:

+ helpers.ParseCSR can parse a CSR. It checks the signature on the
  CSR.    

+ helpers.SignerAlgo takes a crypto.Signer and crypto.Hash and
  returns the appropriate signature algorithm.

+ The `signer.CheckSignature` function has been moved into the `helpers`
   package, making it more widely available to other packages.


lziest:
LGTM

PR #336, opened 2015-09-19
Zi/bundler cli api rethink

lziest:
Rethink the way a bundler is set up:
1. intermediate CA pool, populated through 'int-bundle' parameter, should be optional. It is not a failure to have a empty intermediate CA pool.
2. root CA pool, populated through 'ca-bundle' param, should be optional. It is not a failure for a bundler to have a empty root CA pool, since such bundler can use system roots instead.
3. CA metadata should be optional. But specifying a path that leads to I/O error should be a failure, for cli or api.
4. IntermediateStash, specified through 'int-dir' parameter, should be optional. But specifying a path for it means cfssl will make sure the path exists.
5. fix a few tests that would behave differently between go1.4 and go1.5


lziest:
hold on, I will put one more change to it.

lziest:
Now ready to be reviewed.

kisom:
I don't think force should be the only option: if a `crypto/tls.Config` has a `nil` `RootCAs` field, it will use the operating system roots. I think there's a case to be made for this being valid behaviour.

lziest:
@kisom yes, I should read the document of VerifyOptions. I just assume a nil Root will be bad. Turns out to be OK. I have modified the code and tests accordingly.

kisom:
LGTM.

PR #340, opened 2015-09-21
Save bytes on OCSP responses.

jsha:
When OCSP responder cert is the same as OCSP issuer, don't include the
certificate in the OCSP response, as it is unnecessary.


kisom:
LGTM.

PR #344, opened 2015-09-22
Replace SerialSeq with ClientProvidesSerialNumbers.

jsha:
For Boulder, we're moving to a new serial number generation scheme with an 8-bit datacenter prefix and 136 bits of randomness: https://github.com/letsencrypt/boulder/pull/823. So we don't need the `SerialSeq` feature anymore (and I strongly suspect no-one else uses it). But we do need the ability to specify a serial number when we call into CFSSL.

### DIFF
--- a/api/sign/sign.go
+++ b/api/sign/sign.go
@@ -4,6 +4,7 @@ package sign
 import (
 	"encoding/json"
 	"io/ioutil"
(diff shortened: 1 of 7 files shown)


kisom:
LGTM.

PR #347, opened 2015-09-24
Don't error out on CKR_ATTRIBUTE_TYPE_INVALID.

jsha:
Some HSMs return this error when trying to check for the CKA_ALWAYS_AUTHENTICATE
attribute.

### DIFF
--- a/crypto/pkcs11key/key.go
+++ b/crypto/pkcs11key/key.go
@@ -30,6 +30,24 @@ var hashPrefixes = map[crypto.Hash][]byte{
 	crypto.RIPEMD160: {0x30, 0x20, 0x30, 0x08, 0x06, 0x06, 0x28, 0xcf, 0x06, 0x03, 0x00, 0x31, 0x04, 0x14},
 }
 
+// Ctx defines the subset of pkcs11.Ctx's methods that we use, so we can inject
+// a different Ctx for testing.
+type Ctx interface {
+  CloseSession(sh pkcs11.SessionHandle) error
+	FindObjectsFinal(sh pkcs11.SessionHandle) error
+  FindObjectsInit(sh pkcs11.SessionHandle, temp []*pkcs11.Attribute) error
+  FindObjects(sh pkcs11.SessionHandle, max int) ([]pkcs11.ObjectHandle, bool, error)
+  GetAttributeValue(sh pkcs11.SessionHandle, o pkcs11.ObjectHandle, a []*pkcs11.Attribute) ([]*pkcs11.Attribute, error)
+  GetSlotList(tokenPresent bool) ([]uint, error)
+  GetTokenInfo(slotID uint) (pkcs11.TokenInfo, error)
+	Initialize() error
+  Login(sh pkcs11.SessionHandle, userType uint, pin string) error
+  Logout(sh pkcs11.SessionHandle) error
+  OpenSession(slotID uint, flags uint) (pkcs11.SessionHandle, error)
+  SignInit(sh pkcs11.SessionHandle, m []*pkcs11.Mechanism, o pkcs11.ObjectHandle) error
+  Sign(sh pkcs11.SessionHandle, message []byte) ([]byte, error)
+}
+
 // Key is an implementation of the crypto.Signer interface using a key stored
 // in a PKCS#11 hardware token.  This enables the use of PKCS#11 tokens with
 // the Go x509 library's methods for signing certificates.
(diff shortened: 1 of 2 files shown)


crypto/pkcs11key/key.go
```diff
@@ -30,6 +30,24 @@ var hashPrefixes = map[crypto.Hash][]byte{
 	crypto.RIPEMD160: {0x30, 0x20, 0x30, 0x08, 0x06, 0x06, 0x28, 0xcf, 0x06, 0x03, 0x00, 0x31, 0x04, 0x14},
 }
 
+// Ctx defines the subset of pkcs11.Ctx's methods that we use, so we can inject
+// a different Ctx for testing.
+type Ctx interface {
```
jmhodges: If we're going to do this, can we consider just passing one of these into New and avoid all of that map and mutex code?

I mean, it's a public interface but only the package's code can interoperate with it.

jsha: Making this a public interface was another example of me forgetting the Go capitalization idiom. I meant to make it private.

However, I think that's orthogonal to your point: by leaving this public and taking it as an arg, we can simplify the code a bit.

One argument against: each place that calls this is going to have to implement the same logic to load and initialize the pkcs11 module. That's a certain amount of duplicated code, but additionally it increases the likelihood that the caller will not bother to keep track of whether they've already initialized the module once in their program lifetime, and introduce bugs accidentally. Implementing it here makes it a small bit easier to use correctly.

On the other hand, if we provide guidance in the docs that the caller should just put the naive, lockless version of this code in their package's init method, maybe it's fine?

Cloudflare folks, do you have a strong opinion either way?

jsha: I thought of another argument against telling clients to load the library in `init()` and pass a pointer. If we load the library in `init()`, the best we can do is panic. However, if we do it later during the application, we can return an error. That allows the app to log however it likes, and choose between a clean shutdown or continuing to run with some features disabled (if PKCS#11 support is not a core part of the application).

Also, in general the module path will be a runtime config option, so it's nice to give the app time to load config files before initializing the module.

If we think that in general clients will want to initialize the module outside of `init()`, they start having the same problem of ensuring that the module only loads once, which means reimplementing locking code similar to what we have here.

kisom: We'd prefer it if the package didn't panic during init --- we're not yet using PKCS #11, and it introduces other areas where things could go wrong. Returning errors allows the CFSSL frontend to make choices about how to continue.

jsha: I think we're on the same page, but to be clear: We're not talking about having pkcs11key implement an `init()` function, but about making it the caller's responsibility to import `miekg/pkcs11` and call `pkcs11.New()` and `ctx.Initialize()`, then pass the result to `pkcs11key`. The alternative, the way it currently works, is to leave responsibility for loading the module DLL in the pkcs11key library.

jmhodges: @jsha it would happen during main(), not init(), where you can trivially handle errors, and any libraries built on this (which I would be surprised to find) would just take a Ctx too. 

You can do whatever in this branch, but I don't think arguments about init() are super convincing.

jmhodges: By "you can do whatever", I meant that my qualms were at like 60% and I'm not super worried about it.

jmhodges:
LGTM

kisom:
One final thought: would it make sense it provide some kind of default `Ctx` option (e.g. BasicCtx) that covers common cases?

jsha:
Right now, the client is not actually required to provide a `Ctx` at all. Internally, we get a `*pkcs11.Ctx`, but we treat it as a `Ctx`. The reason is so that we can more easily mock out its methods. It's a relatively common pattern in Go. I'm not sure the best reference, but [this](http://relistan.com/writing-testable-apps-in-go/) seems pretty good.

The rough notion is that if you use type X that implements 50 methods, but you only use 12 of them, you can define an interface Y that consists of exactly those 12 methods. Because Go uses duck typing for interfaces (more or less- I'm sure @jmhodges will correct me), any instance of X automatically implements the Y interface. However, if you want to define a mockX (like my mockCtx), you only have to implement 12 mock methods rather than all 50.

jsha:
BTW, hold off on merging this a sec: I'm going to rename `Ctx` to `ctx` so that it's a private interface rather than a public one.

jsha:
Ok, uploaded a diff.

kisom:
LGTM.

PR #349, opened 2015-09-25
Improve tests.

jsha:
Pull out a test.sh that includes golint and go vet.
Don't `go get` cfssl: It should already be checked out as part of Travis' setup.
This step was overwriting user-specific repos.


kisom:
LGTM, thanks!

PR #351, opened 2015-09-27
Add optional cache control headers to the OCSP responder

### LINKED ISSUE #232: OCSP Responder should return cache-control headers
[responder.go:130](https://github.com/cloudflare/cfssl/blob/master/ocsp/responder.go#L138) should set a cache control header. Something like:
`Cache-Control: public, max-age=[seconds remaining of validity]`

rolandshoemaker:
Changes `ocsp.Source` to return a struct holding the response DER and a `*time.Duration` (instead of just a byte slice) indicating how long the cache control headers should be set for the response. It also adds two bool flags to the `Responder` to gate returning `no-cache` headers on bad or not found requests. Since the `InMemorySource` is static I've also added a argument to `NewSourceFromFile` that allows users to set a static cache lifetime that will be returned for all found responses.

(Side effect: adds tests increasing responder coverage)

Fixes #232.


rolandshoemaker:
I am very confused about the failure, can't repro locally, and it seems like travis is sending a different OCSP request? (based on the b64)

I'm going to blame this on me doing something stupid because it's late and I'll try to fix tomorrow.

jsha:
FWIW, I think it would be better to not add a bunch of configuration options, but to do this by default:

 - Don't cache 404
 - Don't cache bad requests
 - For good requests, parse the Response and look at NextUpdate. Set a Cache-Control header with max-age=&lt;number of seconds between now and NextUpdate&gt;

rolandshoemaker:
I can see a few reasons against doing this

1. There are multiple cases where you may not want to send `<number of seconds between now and NextUpdate>`, for instance if this period is extremely short (in which case you may want to just not cache) or if the response if for a expired/revoked certificate (in which case you may want a much longer/no cache respectively)
2. Re-parsing the response for every request seems like a waste of operations. Both the `InMemorySource`, and the database source we are using in Boulder, either directly parse the response during construction or have access (in Boulders case) to further metadata about the response in order to correctly choose the cache life.

PR #352, opened 2015-09-28
Add AuthRemote.

kisom:
Previously, the AuthKeyName field in a signing profile acted as both the
authentication key that clients would use, and the authentication key
that the profile would use when sending a signing request to a remote
CFSSL. There wasn't any way to specify separate authentication keys.

This PR adds a field to the signing profile: AuthRemote. This pairs an
auth key name and a remote name; the names come from the same auth keys
and remote section of the config, and represent the configuration the
user will use.

As an internal implementation note, the RemoteServer field will now be
set if either the RemoteName or AuthRemote.RemoteName fields are
set.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -53,31 +53,38 @@ type CertificatePolicyQualifier struct {
 	Value string
 }
 
+// AuthRemote is an authenticated remote signer.
+type AuthRemote struct {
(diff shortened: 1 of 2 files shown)


grittygrease:
LGTM

lziest:
LGTM

PR #355, opened 2015-10-01
fix a bundler test

lziest:
A bundler test on reporting ECDSA compatibility was using a expired certificate. We switch to a non-expiring certificate.


kisom:
LGTM.

PR #358, opened 2015-10-01
Add a transport library.

kisom:
The transport library provides tools for building TLS-secured client
and server connections. It is designed to minimise the number of knobs
and switches that are presented to end users, and supports features
such as auto-updating of certificates.


transport/ca/cfssl_provider.go
```diff
@@ -0,0 +1,296 @@
...
+		return nil, errors.New("transport: invalid PEM-encoded certificate signing request")
+	}
+
+	csr, err := x509.ParseCertificateRequest(p.Bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	hosts := make([]string, 0, len(csr.DNSNames)+len(csr.IPAddresses))
+	for i := range csr.DNSNames {
```
jacobhaven: `copy(hosts, csr.DNSNames)`

jacobhaven:
It's possible this should be under `cfssl/helpers/transport`

transport/kp/key_provider_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func TestGenerate(t *testing.T) {
+	err := kp.Load()
+	if err == nil {
+		t.Fatal("key provider shouldn't have a key yet")
+	}
+
+	err = kp.Generate("ecdsa", 256)
```
j-delaney: For better test coverage you could add a test for `Generate` with "rsa" as well

transport/ca/cfssl_provider.go
```diff
@@ -0,0 +1,294 @@
...
+	return "", false
+}
+
+// The client's remote should be set as follows:
+//
+// 1. If the remote has been explicitly set, honour that.
+//
+// 2. If the config specifies a remote, honour that.
+//
+// 3. Use CFKS.
```
grittygrease: This comment is no longer applicable.

transport/client.go
```diff
@@ -0,0 +1,239 @@
...
+	// Identity contains information about the entity that will be
+	// used to construct certificates.
+	Identity *core.Identity
+}
+
+// NewTransport builds a new transport from the default
+func NewTransport(before time.Duration, identity *core.Identity) (*Transport, error) {
+	var tr = &Transport{
+		Before:   before,
+		Roots:    core.SystemRoots,
```
grittygrease: Something is missing here. This doesn't seem to work in the use case where two services, backed by two different CFSSL CAs attempt to talk with each other. Unless the CFSSL CAs were previously added to the system roots, the CFSSL-signed certificates will not be trusted by the client.

When you build a transport, you should be able to choose a set of CA providers to trust rather than trust the system roots. By providing a list of CAs to trust, the transport package could use the CACertificate() function to obtain the trusted roots of those certificates and create a very specialized trust store.

The use case I'm envisioning here is represented by these images:
https://blog.cloudflare.com/content/images/2015/06/provisioning-app-certificates.png https://blog.cloudflare.com/content/images/2015/06/service-to-service-communication.png

transport/core/config.go
```diff
@@ -0,0 +1,45 @@
...
+
+// TLSClientAuthServerConfig returns a new client authentication TLS
+// configuration for servers expecting mutually authenticated
+// clients. The clientAuth parameter should contain the root pool used
+// to authenticate clients.
+func TLSClientAuthServerConfig(cert tls.Certificate, clientAuth *x509.CertPool) *tls.Config {
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		RootCAs:      SystemRoots,
+		ClientCAs:    clientAuth,
```
grittygrease: It would be nice for this list to be automatically created from a list of certificate providers that are known to issue certificates for trusted clients.

transport/core/config.go
```diff
@@ -0,0 +1,45 @@
...
+	"crypto/x509"
+)
+
+// TLSClientAuthClientConfig returns a new client authentication TLS
+// configuration that can be used for a client using client auth
+// connecting to the named host.
+func TLSClientAuthClientConfig(cert tls.Certificate, host string) *tls.Config {
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		RootCAs:      SystemRoots,
```
grittygrease: Again, it would be nice to be able to replace these system roots with a short list comprising of CA providers that are known to be trusted.

transport/transport_test.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+	if !core.SystemRoots.AppendCertsFromPEM(caCert) {
+		t.Fatal("no certificates could be added to system roots")
+	}
+
+	core.SystemRoots.AddCert(tr.Provider.Certificate())
+
+	l, err = Listen("127.0.0.1:8765", trl)
```
grittygrease: This is pretty slick, you can easily set up an HTTPS server with s.Serve(*l).

It would be nice to add an instruction manual for people who have set up their HTTP(S) endpoints with ListenAndServe(TLS) to switch to the new transport.

transport/core/defs.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Identity is used to store information about a particular transport.
+type Identity struct {
+	// Request contains metadata for constructing certificate requests.
+	Request *csr.CertificateRequest `json:"request"`
+
+	// Profiles contains a dictionary of names to dictionaries;
+	// this is intended to allow flexibility in supporting
+	// multiple configurations.
+	Profiles map[string]map[string]string `json:"profiles"`
```
grittygrease: Can this support remotes that have an auth provider?

transport/ca/cfssl_provider.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+var cfsslConfigDirs = []string{
+	"/usr/local/cfssl",
+	"/etc/cfssl",
+	"/state/etc/cfssl",
+}
+
+// The CFKS standard is to have a configuration file for a label as
+// <config>.label.
```
lziest: <label>.json?

jacobhaven:
Rebase off master to fix CI issues...

transport/client.go
```diff
@@ -0,0 +1,308 @@
...
+	if err != nil {
+		log.Debugf("couldn't generate an X.509 keypair: %v", err)
+	}
+
+	return
+}
+
+// Dial initiates a TLS connection to an outbound server. It returns a
+// TLS connection to the server.
+func Dial(address string, tr *Transport) (*tls.Conn, error) {
```
jacobhaven: Why not:
```Go
func (tr *Transport) Dial(address string) (*tls.Conn, error)
```

kisom: The `crypto/tls` library does it this way: it's not

```
func (cfg *Config) Dial...
```

I just tried to keep that style to cut down on having to remember what
to use where.

transport/listener.go
```diff
@@ -0,0 +1,145 @@
...
+	listener net.Listener
+}
+
+// PollInterval is how often to check whether a new certificate has
+// been found.
+var PollInterval = 30 * time.Second
+
+// Listen sets up a new server. If an error is returned, it means
+// the server isn't ready to begin listening.
+func Listen(address string, tr *Transport) (*Listener, error) {
```
jacobhaven: Why not:
```Go
func (tr *Transport) Listen(address string) (*Listener, error)
```

transport/listener.go
```diff
@@ -0,0 +1,145 @@
...
+		log.Debug("listener isn't listening")
+		return nil, errors.New("transport: listener isn't active")
+	}
+
+	conn, err := l.listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+
+	conn = tls.Server(conn, l.config)
```
jacobhaven: I think your problem is here. You shouldn't wrap `tls.Server` around the already created `*tls.Conn`.

transport/core/defs.go
kisom: It can --- c.f. 73155f59754fd98574831a155ffe90ba28b8c34f.

transport/ca/cert_provider.go
```diff
@@ -0,0 +1,16 @@
+// Package ca provides the CertificateAuthority interface for the
+// transport package, which provide an interface to get a CSR signed
```
grittygrease: *provides

transport/ca/cfssl_provider.go
```diff
@@ -0,0 +1,285 @@
...
+
+	return v4Loopback.Contains(ip)
+}
+
+// The only time a client should be doing unauthenticated requests is
+// when a local CFSSL is being used.
+func (cap *CFSSL) validateAuth() error {
+	// The client is using some form of authentication, and the best way
+	// to figure out that the auth is invalid is when it's used. Therefore,
+	// we'll elide checking the credentials until that time.
```
grittygrease: I prefer "delay" to "elide"

transport/ca/cfssl_provider.go
```diff
@@ -0,0 +1,285 @@
...
+	return nil
+}
+
+var cfsslConfigDirs = []string{
+	"/usr/local/cfssl",
+	"/etc/cfssl",
+	"/state/etc/cfssl",
+}
+
+// The CFKS standard is to have a configuration file for a label as
```
grittygrease: CFKS is not a well-defined term, I would use CFSSL here.

transport/example/maserver/server.json
```diff
@@ -0,0 +1,26 @@
...
+	    "profile": "server",
+	    "remote": "127.0.0.1:8888"
+	}
+    },
+    "roots": [{
+	    "type": "system"
+    }],
+    "client_roots": [{
+	    "type": "cfssl",
+            "metadata": {
```
grittygrease: alignment is off

transport/core/config.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+)
+
+// TLSClientAuthClientConfig returns a new client authentication TLS
+// configuration that can be used for a client using client auth
+// connecting to the named host.
+func TLSClientAuthClientConfig(cert tls.Certificate, host string) *tls.Config {
```
grittygrease: Not sure how these fit in.

kisom: I've since moved these over the the `transport` package and `git rm`'d this file (collecting comments before a commit and push).

transport/ca/cfssl_provider.go
```diff
@@ -0,0 +1,285 @@
...
+	return f(ak, ad)
+}
+
+// ErrNoAuth is returned when a client is talking to a CFSSL remote
+// that is not on a loopback address and doesn't have an
+// authentication provider set.
+var ErrNoAuth = errors.New("transport: authentication is required for non-local remotes")
+
+var v4Loopback = net.IPNet{
+	IP:   net.IP{127, 0, 0, 1},
```
lmb: Threw me for a second. Maybe make it `net.IP{127, 0, 0, 0}`?

transport/client.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	return tls.Dial("tcp", address, cfg)
+}
+
+// AutoUpdate will automatically update the listener. If a non-nil
+// certUpdates chan is provided, it will receive timestamps for
+// reissued certificates. If errChan is non-nil, any errors that occur
+// in the updater will be passed along.
+func (tr *Transport) AutoUpdate(certUpdates chan time.Time, errChan chan error) {
```
lmb: Explicitly use read/write-only chans? Also, do we want to crash the running go process if this goroutine panics?

transport/client.go
```diff
@@ -0,0 +1,316 @@
...
+		Certificates: []tls.Certificate{cert},
+		CipherSuites: core.CipherSuites,
+		MinVersion:   tls.VersionTLS12,
+	}, nil
+}
+
+// New builds a new transport from an identity and a before time. The
+// before time tells the transport how long before the certificate
+// expires to start attempting to update when auto-updating.
+func New(before time.Duration, identity *core.Identity) (*Transport, error) {
```
lmb: What happens if before > lifetime of a freshly issued cert?

transport/doc.go
```diff
@@ -0,0 +1,34 @@
...
+//                      },
+//              },
+//     }
+//
+//
+//
+// The NewTransport function will return a transport built using the
+// NewKeyProvider and NewCA functions. These functions may be changed
+// by other packages to provide common key provider and CA
+// configurations. Clients can then use Rekey (or launch AutoUpdate in
```
lmb: Rekey doesn't exist?

transport/example/config_auth.json
```diff
@@ -0,0 +1,37 @@
+{
+    "auth_keys": {
+        "client": {
+            "type": "standard",
+            "key": "52abb3ac91971bb72bce17e7a289cd04476490b19e0d8eb7810dc42d4ac16c41"
```
lmb: Noob question, config_auth.json and similar are not supposed to be committed to git I assume?

lmb:
Questions on certificate rollover, all in the context of very long running connections:
- Does a server have to react to an updated cert by resetting connections?
- Does a client have to AutoUpdate?
- If yes, should a client restart connections as well on a new certificate?

transport/client.go
kisom: If the auto updater is running, every time it checks the time (e.g. every `PollInterval` seconds) it will request a new certificate.

transport/example/config_auth.json
kisom: For the example, they should be. The goal is to provide everything to the user to run the examples (except the certificate and key, which are generated with the provided script).

grittygrease:
Re: certificate rollover questions, the certificate is only used at the establishment of a connection so long-running connections can have the certificate swapped out from under them with no danger. Clients should auto-update when their certificate is close to expiration. The same logic applies, no restart needed.

kisom:
@lmb Answered these questions in the latest commit (so that they're also in the documentation now), but the gist is:

* If `AutoUpdate` is running, no additional actions have to be taken. Existing connections are fine, new connections will use the new certificate.
* Clients don't need `AutoUpdate` if they're only making a single connection, but otherwise they should run it.
* If before > the issuance life time of a certificate, every time an update check is done, the package will request a new certificate.

grittygrease:
Looks good to merge.

lziest:
LGTM

transport/client.go
```diff
@@ -0,0 +1,318 @@
...
+			if err == nil {
+				break
+			}
+
+			log.Debug("failed to update certificate, will try again in 5 minutes")
+			if errChan != nil {
+				errChan <- err
+			}
+
+			<-time.After(5 * time.Minute)
```
ashishgandhi: Exponential back-off?

kisom: Added in [`transport/core/backoff.go`](https://github.com/cloudflare/cfssl/blob/51e1ac76731f0fe62a9dad99551820578bff177b/transport/core/backoff.go)

ashishgandhi:
:+1:

PR #359, opened 2015-10-05
Update scan to verify against the CA Bundle argument. 

jacobhaven:
TLSDial returns a warning on bad verify.


jacobhaven:
@ejcx This should make it so you only need to check that TLSDial scan succeeds with a `Good` grade.

grittygrease:
LGTM

PR #361, opened 2015-10-08
Fix broken build and gofmt pkcs11key

grittygrease:


grittygrease:
@jsha The build was failing when `-tags nopkcs11` was set. Can you take a look at this fix?

jsha:
Will take a look!

test.sh
```diff
@@ -16,3 +16,6 @@ fgt golint ./...
 go test ./...
 go list -f '{{if len .TestGoFiles}}"go test -coverprofile={{.Dir}}/.coverprofile {{.ImportPath}}"{{end}}' ./... | xargs -i sh -c {}
 gover . coverprofile.txt
+go get -d ./...
+go build -tags nopkcs11 cmd/...
```
jsha: Why only build cmd with the nopkcs11 tag?

Also FYI: depending on how long this takes, you could speed up the total time your tests take with travis' matrix feature. E.g. you'd replace `go test ./...` with `go test $BUILD_FLAGS ./...` and put in your .travis.yml:

```
env:
  matrix:
    - BUILD_FLAGS=
    - BUILD_FLAGS="-tags nopkcs11"
```

test.sh
```diff
@@ -16,3 +16,6 @@ fgt golint ./...
 go test ./...
 go list -f '{{if len .TestGoFiles}}"go test -coverprofile={{.Dir}}/.coverprofile {{.ImportPath}}"{{end}}' ./... | xargs -i sh -c {}
 gover . coverprofile.txt
+go get -d ./...
```
jsha: What's the go get for?

grittygrease: This was to be consistent with the BUILDING.sh

jsha: Aha, I see. It's probably worth diverging from the BUILDING.md file in Travis, because (a) the get is redundant, and (b) it looks like this go get is breaking the Travis build because go disagrees with Travis about the name of the repo (.../cfssl.git vs .../cfssl).

jsha:
I'd be happy to take over this branch and get the Travis tests passing if you'd like, since I caused the breakage. Let me know if you'd like that!

.travis.yml
```diff
@@ -33,5 +34,8 @@ notifications:
     on_failure: change
 env:
   - secure: "OmaaZ3jhU9VQ/0SYpenUJEfnmKy/MwExkefFRpDbkRSu/hTQpxxALAZV5WEHo7gxLRMRI0pytLo7w+lAd2FlX1CNcyY62MUicta/8P2twsxp+lR3v1bJ7dwk6qsDbO7Nvv3BKPCDQCHUkggbAEJaHEQGdLk4ursNEB1aGimuCEc="
+  matrix:
```
jsha: I think this either needs a `-` in front, or the above `- secure:` needs to become `secure:`

grittygrease:
I updated the Travis YAML file, just waiting for CI to finish. If it is still failing after this, feel free to take this over.

jsha:
Ok, I see a failure: I'm going to take over this branch and make it work. Sorry for the breakage!

grittygrease:
Looks like it's fixed now. Are you ok with the changes?

test.sh
```diff
@@ -13,6 +13,8 @@ if ! which golint > /dev/null ; then
   exit 1
 fi
 fgt golint ./...
-go test ./...
+go test $BUILD_FLAGS ./...
 go list -f '{{if len .TestGoFiles}}"go test -coverprofile={{.Dir}}/.coverprofile {{.ImportPath}}"{{end}}' ./... | xargs -i sh -c {}
 gover . coverprofile.txt
+go get $BUILD_FLAGS -d ./...
```
jsha: These two commands (the go get and go build) are now redundant. For Go projects, Travis already runs `go get -t -v ./...` during its `install` phase. And the `go test $BUILD_FLAGS ./...` above automatically does a go build for that directory. I'd recommend removing this line and the one below.

grittygrease: done

jsha:
LGTM

PR #362, opened 2015-10-08
Run Travis for PRs only.

jsha:
Based on my feedback in https://github.com/cloudflare/cfssl/pull/361, CFSSL now
has four entries in its build matrix: BUILD_FLAGS=""/"nopkcs11" and
(preexisting) Go 1.4 / Go 1.5. For public projects Travis has a limit of five
simultaneous builds, and each entry in the build matrix consumes one.

So we're still under the limit of five, except that Travis builds all branches
plus pull requests. So generally when someone pushes a branch and then makes a
PR based off of it, Travis will be trying to run two builds simultaneously, with
four build matrix entries in each. This means that one of the two builds is
blocked on the other completing, which makes things much slower.

In boulder we've solved this issue by disabling builds for branches except for
master, and any branch named "test-*".


jacobhaven:
LGTM

PR #363, opened 2015-10-09
Fix ValidExpiry test that wasn't being run. Add .coverprofile to .gitignore

andrewbuss:
HasValidExpiry was not run because it didn't start with Test. 
test.sh creates a bunch of .coverprofile files which should be ignored.


grittygrease:
LGTM

PR #367, opened 2015-10-09
Add appendIf test

haneric21:
Added test for appendIf func in signer_test.go


lziest:
This is unnecessary. Please aim on exported/public method.

grittygrease:
Tests on private functions are fine, but I agree that focusing on exported methods provides more value.

lziest:
Yes, thanks for contribution. Sorry I should sound more friendly. This test is fine.

PR #368, opened 2015-10-09
Zi/renew ca

lziest:
The aim is to add a method that can takes an expiring root CA certificate and key to generate a new root CA certificate with the same expiry length. The approach is to reuse the initca.NewFromSigner code path as much as possible. So we add a method in package csr to extract CertificateRequest from x509.Certificate.


initca/initca.go
```diff
@@ -179,6 +227,51 @@ func NewFromSigner(req *csr.CertificateRequest, priv crypto.Signer) (cert, csrPE
...
+// RenewFromSigner re-creates a root certificate from the CA cert and crypto.Signer.
+// The resulting root certificate will have ca
+// as the template and have the same expiry length.
+func RenewFromSigner(ca *x509.Certificate, priv crypto.Signer) ([]byte, error) {
+	if !ca.IsCA {
+		return nil, errors.New("input certificate is not a CA cert")
+	}
+
+	// it is not within renew window of expiration, don't re-generate
+	if time.Now().AddDate(0, 0, RenewWindowInDays).Before(ca.NotAfter) {
```
grittygrease: Why do we restrict the time period for which we can renew a CA?

andrewbuss:
For later reference, which commands did you use to generate the test data?

grittygrease:
How the expiration time of the new CA is computed should be spelled out somewhere in the docs.

initca/initca.go
lziest: Trying to only have expiring CA cert be re-generated. I am open to remove it. I guess it is better to let caller decide whether to call Renew.

lziest:
@andrewbuss for generating CA certificate, check the new README I added 
https://github.com/cloudflare/cfssl/commit/2ada58a613d4a4f0f36f1d28d074be32b5b45750

initca/initca.go
lziest: Now removed.

grittygrease:
LGTM

PR #370, opened 2015-10-09
Logtest

anniephan:
WIP


log/log_test.go
```diff
@@ -1 +1,160 @@
...
+	log.SetOutput(buf)
+	outputf(LevelDebug, string1, nil)
+
+	// test 1: outputf correctly prints string
+	if !strings.Contains(buf.String(), string1) {
+		t.Fail()
+	}
+
+	(*bytes.Buffer).Reset(buf)
+	outputf(LevelDebug-1, string1, nil)
```
lziest: can you use the actual Level constant?

log/log_test.go
```diff
@@ -1 +1,160 @@
...
+
+	// test 1: outputf correctly prints string
+	if !strings.Contains(buf.String(), levelPrefix[LevelInfo]) {
+		t.Fail()
+	}
+	return
+}
+
+func TestDebugf(t *testing.T) {
+	const string1 = "asdf123"
```
lziest: can we use an actual format string? :)

log/log_test.go
```diff
@@ -1 +1,160 @@
...
+)
+
+func TestOutputf(t *testing.T) {
+	const string1 = "asdf123"
+	buf := new(bytes.Buffer)
+
+	log.SetOutput(buf)
+	outputf(LevelDebug, string1, nil)
+
+	// test 1: outputf correctly prints string
```
lziest: if you planning to add more test, please do. If not, remove "test 1" here and at other places.

grittygrease:
Please address the comments above and squash into one commit.

log/log_test.go
anniephan: It turns out that there is no Level constant lower than LevelDebug. So it seems like there is no need to include that second test... Thanks! I will fix.

log/log_test.go
```diff
@@ -1 +1,134 @@
 package log
+
+import (
+	"bytes"
+	"log"
+	"strings"
+	"testing"
+)
+
+const string1 = "asdf123"
```
grittygrease: Can you update this name to not have "1" in it?

anniephan: Ok

PR #372, opened 2015-10-09
Add tests on signing profile with AuthRemote

lziest:


grittygrease:
Thanks for the PR. It looks like there are no good example configs or documentation for this feature. That would be a good next step.

Issue #279, opened 2015-07-27
crashers found by `go-fuzz`


dgryski:
cfssl includes a mostly vendored copy of Azure/go-pkcs12.  I just filed  https://github.com/Azure/go-pkcs12/issues/25 that might affect cfssl too.

kisom:
@dgryski Thanks for the heads up. We'll keep an eye out for patches upstream and investigate patching our copy.

dgryski:
Azure/go-pkcs12 has now been merged into /x/crypto: https://github.com/golang/crypto/commit/c8b9e6388ef638d5a8a9d865c634befdc46a6784 , including fixes for these crashers.  Consider moving to that code base instead?

dgryski:
Closing since cfssl now uses /x/crypto/pkc12 which has fixes for these crashes.

PR #375, opened 2015-10-09
Adapt ParseCertificatesDER to use the new golang pkcs12 library

andrewbuss:
We no longer need to maintain crypto/pkcs12 since pkcs12 has been merged with golang. Addresses comment on #279 . The interface is very similar except the new pkcs12.Decode returns only a single certificate pointer rather than a list of them.


grittygrease:
Looks good. Always appreciate when more code is removed than added.

PR #378, opened 2015-10-09
Return error if config file has no "Signing" field

j-delaney:
Running `cfssl gencert` with a config file that lacks the `Signing` field would cause a `runtime error: invalid memory address or nil pointer dereference`. Now, `config.go` returns an error if there is no `Signing` field


lziest:
lgtm

PR #379, opened 2015-10-10
Fixed nil pointer dereference in config if a config file lacks "signing"

andrewbuss:
Added regression test which fails without the change

cosmetic: gofmt also fixed some formatting issues


andrewbuss:
Woops - @j-delaney already fixed this!

config/config.go
```diff
@@ -86,14 +86,15 @@ type SigningProfile struct {
 	CSRWhitelist                *CSRWhitelist
 	NameWhitelist               *regexp.Regexp
 	ClientProvidesSerialNumbers bool
+	CTLog                       string
```
grittygrease: This should be saved for a separate commit.

PR #383, opened 2015-10-10
Initial implementation of Certificate Transparency

andrewbuss:
Add a config field "ct-log" to specify the URL of a CT log server. cfssl sign will submit the cert chain to the specified server after signing the client's CSR, but before returning the signed cert to them.

No efforts are made to validate the returned SCT or incorporate it into a certificate, or do anything with pre-certificates. (next steps)

This works both on the command line and with the API. I tested by setting up a CT log server and installing my testing cert as one of its trusted roots. 

Testing this requires running a certificate transparency log server under our control, since the implementation rejects untrusted certificates. 

I'll work on dockerizing the CT log server so we can pass in a test cfssl-generated CA certificate to be included in the set of certificates that the log server trusts.

An alternate solution would be to run a public CT log server (which should never be trusted by browsers...) patched to trust any root certificate presented to it. Then testing wouldn't require spinning up a container..

### DIFF
--- a/api/certinfo/certinfo.go
+++ b/api/certinfo/certinfo.go
@@ -1,0 +1,50 @@
+// Package certinfo implements the HTTP handler for the certinfo command.
+package certinfo
+
+import (
+	"net/http"
+
+	"github.com/cloudflare/cfssl/api"
+	"github.com/cloudflare/cfssl/certinfo"
+	"github.com/cloudflare/cfssl/log"
+)
+
+// Handler accepts requests for either remote or uploaded
+// certificates to be bundled, and returns a certificate bundle (or
+// error).
+type Handler struct{}
+
+// NewHandler creates a new bundler that uses the root bundle and
+// intermediate bundle in the trust chain.
+func NewHandler() http.Handler {
+	return api.HTTPHandler{Handler: new(Handler), Methods: []string{"POST"}}
+}
+
+// Handle implements an http.Handler interface for the bundle handler.
+func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) (err error) {
+	blob, matched, err := api.ProcessRequestFirstMatchOf(r,
+		[][]string{
+			[]string{"certificate"},
+			[]string{"domain"},
+		})
+	if err != nil {
+		log.Warningf("invalid request: %v", err)
+		return err
+	}
+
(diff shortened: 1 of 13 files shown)


config/config.go
```diff
@@ -76,6 +76,7 @@ type SigningProfile struct {
 	NotAfter            time.Time  `json:"not_after"`
 	NameWhitelistString string     `json:"name_whitelist"`
 	AuthRemote          AuthRemote `json:"auth_remote"`
+	CTLog               string     `json:"ct_log"`
```
grittygrease: This should probably be an array of CT logs.

signer/local/local.go
```diff
@@ -250,7 +253,19 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
-	return s.sign(&safeTemplate, profile)
+	cert, err = s.sign(&safeTemplate, profile)
+
+	if profile.CTLog != "" {
+		var derCert, _ = pem.Decode(cert)
+		var ctclient = client.New(profile.CTLog)
+		chain := []ct.ASN1Cert{derCert.Bytes, s.ca.Raw}
+		var _, err = ctclient.AddChain(chain)
+		if err != nil {
+			return nil, err
```
grittygrease: This should be wrapped with the appropriate CFSSL error. Also, what are the error cases for this API? What happens if the CT server is temporarily unavailable? Should the certificate issuance fail, or should the certificate be put in a queue to resubmit to the server?

config/config.go
andrewbuss: I've fixed this on my own branch - is there a way to update this PR with that commit?
Maybe best to just keep development on a branch and hold off on merging until the feature is more complete.

signer/local/local.go
andrewbuss: I think that the certificate should not be issued to the client until it's been submitted to the list of log servers specified in the config, which implies that it should fail if the submission fails.

grittygrease: Does the server return specific status codes? Maybe a retry is in order in some cases.

andrewbuss: The CT log client actually takes care of that for us - retrying with backoff when appropriate until it succeeds.

grittygrease: In that case, this works.

PR #386, opened 2015-10-11
enable https connection with endpoint CA and key provided

codyli520:
Added tags "-tls-cert" and "-tls-key" for "cfssl serve" command.
With a endpoint user's certificate and key (pem files) , the cfssl can be served on a secure HTTP protocol (HTTPS)

using Go's default cipher suites , please comment if your browser supports other cipher suites thats not listed.

    TLS_RSA_WITH_RC4_128_SHA                
    TLS_RSA_WITH_3DES_EDE_CBC_SHA           
    TLS_RSA_WITH_AES_128_CBC_SHA            
    TLS_RSA_WITH_AES_256_CBC_SHA            
    TLS_ECDHE_ECDSA_WITH_RC4_128_SHA       
    TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA    
    TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA   
    TLS_ECDHE_RSA_WITH_RC4_128_SHA          
    TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA    
    TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA      
    TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
    TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256  
    TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
    TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384   
    TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
@@ -22,8 +23,7 @@ import (
 
 // A server points to a single remote CFSSL instance.
 type server struct {
-	Address string
-	Port    int
+	URL string
 }
 
 // A Remote points to at least one (but possibly multiple) remote
@@ -50,7 +50,11 @@ func NewServer(addr string) Remote {
 	if len(addrs) > 1 {
 		remote, _ = NewGroup(addrs, StrategyOrderedList)
 	} else {
-		srv := newServer(addr)
+		u, err := normalizeURL(addrs[0])
+		if err != nil {
+			return nil
+		}
+		srv, _ := newServer(u)
 		if srv != nil {
 			remote = srv
 		}
@@ -59,34 +63,16 @@ func NewServer(addr string) Remote {
 }
 
 func (srv *server) Hosts() []string {
-	return []string{net.JoinHostPort(srv.Address, fmt.Sprintf("%d", srv.Port))}
+	return []string{srv.URL}
 }
 
-func newServer(addr string) *server {
-	addr = strings.TrimSpace(addr)
-	host, port, err := net.SplitHostPort(addr)
-	if err != nil {
-		host, port, err = net.SplitHostPort(addr + ":8888")
-		if err != nil {
-			return nil
-		}
-	}
-
-	var portno int
-	if port == "" {
-		portno = 8888
-	} else {
-		portno, err = strconv.Atoi(port)
-		if err != nil {
-			return nil
-		}
-	}
-
-	return &server{host, portno}
+func newServer(u *url.URL) (*server, error) {
+	URL := u.String()
+	return &server{URL}, nil
 }
 
 func (srv *server) getURL(endpoint string) string {
-	return fmt.Sprintf("http://%s:%d/api/v1/cfssl/%s", srv.Address, srv.Port, endpoint)
+	return fmt.Sprintf("%s/api/v1/cfssl/%s", srv.URL, endpoint)
 }
 
 // post connects to the remote server and returns a Response struct
@@ -265,4 +251,45 @@ func NewAuthServer(addr string, provider auth.Provider) *AuthRemote {
 // Sign is overloaded to perform an AuthSign request using the default auth provider.
 func (ar *AuthRemote) Sign(req []byte) ([]byte, error) {
 	return ar.AuthSign(req, nil, ar.provider)
+}
+
+// nomalizeURL checks for http/https protocol, appends "http" as default protocol if not defiend in url
+func normalizeURL(addr string) (*url.URL, error) {
+	addr = strings.TrimSpace(addr)
+
+	u, err := url.Parse(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	if u.Opaque != "" {
+		u.Host = net.JoinHostPort(u.Scheme, u.Opaque)
+		u.Opaque = ""
+	} else if u.Path != "" && !strings.Contains(u.Path, ":") {
+		u.Host = net.JoinHostPort(u.Path, "8888")
+		u.Path = ""
+	} else if u.Scheme == "" {
+		u.Host = u.Path
+		u.Path = ""
+	}
+
+	if u.Scheme != "https" {
+		u.Scheme = "http"
+	}
+
+	_, port, err := net.SplitHostPort(u.Host)
+	if err != nil {
+		_, port, err = net.SplitHostPort(u.Host + ":8888")
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if port != "" {
+		_, err = strconv.Atoi(port)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return u, nil
 }
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -1,10 +1,9 @@
 package client
 
 import (
+	"github.com/cloudflare/cfssl/auth"
 	"net"
 	"testing"
-
-	"github.com/cloudflare/cfssl/auth"
 )
 
 var (
@@ -33,8 +32,20 @@ func TestNewServer(t *testing.T) {
 
 	s = NewServer("127.0.0.1:8888")
 	hosts := s.Hosts()
-	if len(hosts) != 1 && hosts[0] != "127.0.0.1:8888" {
-		t.Fatalf("expected [127.0.0.1:8888], but have %v", hosts)
+	if len(hosts) != 1 || hosts[0] != "http://127.0.0.1:8888" {
+		t.Fatalf("expected [http://127.0.0.1:8888], but have %v", hosts)
+	}
+
+	s = NewServer("http://1.1.1.1:9999")
+	hosts = s.Hosts()
+	if len(hosts) != 1 || hosts[0] != "http://1.1.1.1:9999" {
+		t.Fatalf("expected [http://1.1.1.1:9999], but have %v", hosts)
+	}
+
+	s = NewServer("https://1.1.1.1:8080")
+	hosts = s.Hosts()
+	if len(hosts) != 1 || hosts[0] != "https://1.1.1.1:8080" {
+		t.Fatalf("expected [https://1.1.1.1:8080], but have %v", hosts)
 	}
 }
 
@@ -74,7 +85,45 @@ func TestSign(t *testing.T) {
 }
 
 func TestNewServerGroup(t *testing.T) {
-	s := NewServer("cfssl1.local:8888, cfssl2.local:8888")
+	s := NewServer("cfssl1.local:8888, cfssl2.local:8888, http://cfssl3.local:8888, http://cfssl4.local:8888")
+
+	ogl, ok := s.(*orderedListGroup)
+	if !ok {
+		t.Fatalf("expected NewServer to return an ordered group list with a list of servers, instead got a %T = %+v", ogl, ogl)
+	}
+
+	if len(ogl.remotes) != 4 {
+		t.Fatalf("expected the remote to have four servers, but it has %d", len(ogl.remotes))
+	}
+
+	hosts := ogl.Hosts()
+	if len(hosts) != 4 {
+		t.Fatalf("expected 2 hosts in the group, but have %d", len(hosts))
+	}
+
+	if hosts[0] != "http://cfssl1.local:8888" {
+		t.Fatalf("expected to see http://cfssl1.local:8888, but saw %s",
+			hosts[0])
+	}
+
+	if hosts[1] != "http://cfssl2.local:8888" {
+		t.Fatalf("expected to see http://cfssl2.local:8888, but saw %s",
+			hosts[1])
+	}
+
+	if hosts[2] != "http://cfssl3.local:8888" {
+		t.Fatalf("expected to see http://cfssl1.local:8888, but saw %s",
+			hosts[2])
+	}
+
+	if hosts[3] != "http://cfssl4.local:8888" {
+		t.Fatalf("expected to see http://cfssl2.local:8888, but saw %s",
+			hosts[3])
+	}
+}
+
+func TestNewTLSServerGroup(t *testing.T) {
+	s := NewServer("https://cfssl1.local:8888, https://cfssl2.local:8888")
 
 	ogl, ok := s.(*orderedListGroup)
 	if !ok {
@@ -90,13 +139,13 @@ func TestNewServerGroup(t *testing.T) {
 		t.Fatalf("expected 2 hosts in the group, but have %d", len(hosts))
 	}
 
-	if hosts[0] != "cfssl1.local:8888" {
-		t.Fatalf("expected to see cfssl1.local:8888, but saw %s",
+	if hosts[0] != "https://cfssl1.local:8888" {
+		t.Fatalf("expected to see https://cfssl1.local:8888, but saw %s",
 			hosts[0])
 	}
 
-	if hosts[1] != "cfssl2.local:8888" {
-		t.Fatalf("expected to see cfssl2.local:8888, but saw %s",
+	if hosts[1] != "https://cfssl2.local:8888" {
+		t.Fatalf("expected to see https://cfssl2.local:8888, but saw %s",
 			hosts[1])
 	}
 }
--- a/cli/config.go
+++ b/cli/config.go
@@ -18,6 +18,8 @@ type Config struct {
 	CSRFile           string
 	CAFile            string
 	CAKeyFile         string
+	TLSCertFile       string
+	TLSKeyFile        string
 	KeyFile           string
 	IntermediatesFile string
 	CABundleFile      string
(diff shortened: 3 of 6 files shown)


cli/config.go
```diff
@@ -62,8 +64,10 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.StringVar(&c.Hostname, "hostname", "", "Hostname for the cert, could be a comma-separated hostname list")
 	f.StringVar(&c.CertFile, "cert", "", "Client certificate that contains the public key")
 	f.StringVar(&c.CSRFile, "csr", "", "Certificate signature request file for new public key")
-	f.StringVar(&c.CAFile, "ca", "", "CA used to sign the new certificate")
```
lziest: looks like a merge conflict that brings some old code back.

lziest:
LGTM, but please squash the commits into one. Thanks!

grittygrease:
@codyli520 This is only the first step. You should also do the following:
- support TLS in the multirootca
- support the ability to define a remote CA where the remote CFSSL has TLS enabled. For this you will need to provide a list of trusted roots.

codyli520:
Allowing TLS in multirootca, user must specify -epcert -epkey in multirootca command

cli/config.go
```diff
@@ -64,6 +66,8 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.StringVar(&c.CSRFile, "csr", "", "Certificate signature request file for new public key")
 	f.StringVar(&c.CAFile, "ca", "", "CA used to sign the new certificate")
 	f.StringVar(&c.CAKeyFile, "ca-key", "", "CA private key")
+	f.StringVar(&c.EPCAFile, "endpoint-cert", "", "Other endpoint CA used to sign the new certificate")
```
jacobhaven: I think I would call these `tls-cert` and `tls-key`, with `TLSCertFile` and `TLSKeyFile`

codyli520: ok

cli/serve/serve.go
```diff
@@ -192,8 +194,52 @@ func serverMain(args []string, c cli.Config) error {
...
-	return http.ListenAndServe(addr, nil)
+
+	endpointCA := conf.EPCAFile
+	endpointKey := conf.EPCAKeyFile
+
+	if endpointCA == "" || endpointKey == "" {
+		log.Info("Now listening on ", addr)
+		return http.ListenAndServe(addr, nil)
+	}
+	var CipherSuites = []uint16{
```
jacobhaven: How is this different from using the default set of Go ciphersuites (the SHA384 ciphersuites will be included in Go 1.5 binaries and not in Go 1.4)

codyli520: I have found default ciphersuites are not all supported by Safari browser, but I will double check the default set.

cmd/multirootca/ca.go
```diff
@@ -41,6 +42,8 @@ func main() {
 	flagAddr := flag.String("a", ":8888", "listening address")
 	flagRootFile := flag.String("roots", "", "configuration file specifying root keys")
 	flagDefaultLabel := flag.String("l", "", "specify a default label")
+	flagEndpointCert := flag.String("epcert", "", "server certificate")
```
jacobhaven: `tls-cert` and `tls-key` are probably clearer

cli/serve/serve.go
jacobhaven: The problem isn't to have all the cipher suites supported by the browser, it's to have at least one cipher suite supported by all major browsers (that's the one it will use to negotiate)

codyli520: ok, i will check on that

grittygrease:
Great, the main thing missing is setting the /api/client package to use HTTPS if defined in the configuration.

kisom:
One idea to get client support done is consider adding a function `NewTLSServer`.

cli/serve/serve.go
```diff
@@ -192,8 +194,31 @@ func serverMain(args []string, c cli.Config) error {
...
+
+	tlsCA := conf.TLSCertFile
+	tlsKey := conf.TLSKeyFile
+
+	if tlsCA == "" || tlsKey == "" {
+		log.Info("Now listening on ", addr)
+		return http.ListenAndServe(addr, nil)
+	}
+
+	//config struct
```
kisom: This comment is extraneous, but in general, comments should be complete sentences.

cli/serve/serve.go
```diff
@@ -192,8 +194,31 @@ func serverMain(args []string, c cli.Config) error {
...
+	tlsConfig := &tls.Config{
+		CipherSuites: nil,
+	}
+
+	//server struct
```
kisom: Previous comment applies here as well.

cmd/multirootca/ca.go
```diff
@@ -84,6 +87,26 @@ func main() {
...
 	http.Handle("/api/v1/cfssl/metrics", metrics)
-	log.Info("listening on ", *flagAddr)
-	log.Error(http.ListenAndServe(*flagAddr, nil))
+
+	if *flagEndpointCert == "" && *flagEndpointKey == "" {
+		log.Info("listening on ", *flagAddr)
+		log.Error(http.ListenAndServe(*flagAddr, nil))
+	} else {
+
+		//config struct
```
kisom: Previous comment applies here as well.

cmd/multirootca/ca.go
```diff
@@ -84,6 +87,26 @@ func main() {
...
+		tlsConfig := &tls.Config{
+			CipherSuites: nil,
+		}
+
+		//server struct
```
kisom: Previous comment applies here as well.

codyli520:
I've changed up api/client package a little bit, expecting the config file remote address to specify URL with  "http" or "https"  + ://address:port. Then I will take it as a string and create new server using this URL.
And http.POST work with http and https requests.

grittygrease:
Have you tested this with locally-running CFSSLs?

api/client/client.go
```diff
@@ -59,34 +56,16 @@ func NewServer(addr string) Remote {
...
-		portno = 8888
-	} else {
-		portno, err = strconv.Atoi(port)
-		if err != nil {
-			return nil
-		}
-	}
 
-	return &server{host, portno}
+	return &server{addr}
```
grittygrease: Existing clients use hosts that do not contain the protocol version. You should add http:// as a default if none is defined.

api/client/client.go
```diff
@@ -45,11 +44,20 @@ type Remote interface {
 func NewServer(addr string) Remote {
 	addrs := strings.Split(addr, ",")
 
+	for i := range addrs {
```
grittygrease: consider using the "net/url" package here instead of using raw string operations:
https://golang.org/pkg/net/url/#example_URL

api/client/client.go
```diff
@@ -45,11 +44,20 @@ type Remote interface {
...
+			addrs[i] = "http://" + addrs[i]
+		}
+	}
+
 	var remote Remote
 
 	if len(addrs) > 1 {
 		remote, _ = NewGroup(addrs, StrategyOrderedList)
 	} else {
+		if !strings.Contains(addr, "http") {
```
grittygrease: This should probably only be done at one centralized location. You're normalizing a URL.

api/client/client.go
```diff
@@ -59,14 +67,25 @@ func NewServer(addr string) Remote {
...
+	return []string{srv.URL}
 }
 
 func newServer(addr string) *server {
 	addr = strings.TrimSpace(addr)
-	host, port, err := net.SplitHostPort(addr)
+	address := ""
+	protocol := ""
+
+	if strings.Contains(addr, "http://") {
```
grittygrease: again, this could be a good place for https://golang.org/pkg/net/url/#URL

api/client/client.go
codyli520: I think one limitation of net/url is that a url parsed in the format of http://xxxx.com will be parsed as "scheme://[userinfo@]host/path[?query][#fragment]" , where the url.Scheme part will be either http/https.
If the url is without a protocol, ex.127.0.0.1:8888, net/url will parse the url as "scheme:opaque[?query][#fragment]" where url.Scheme is recognized as 127.0.0.1. 
So I think a little string concatenation for url without http/https define here is necessary?

grittygrease: Good point. Maybe check if scheme ends up being http or https, and then append http:// if it is not and parse again? In any case, this normalization logic should be in one place.

kisom: There's two approaches you could take here:

The first is a regular expression approach, e.g. http://play.golang.org/p/Len2gWH_-x

The other uses the strings package: http://play.golang.org/p/Q_SVb8Gpnn

codyli520: put normalization in its own function and thanks for the suggestions @kisom, I took the strings route.

api/client/client.go
```diff
@@ -59,14 +63,16 @@ func NewServer(addr string) Remote {
 }
 
 func (srv *server) Hosts() []string {
-	return []string{net.JoinHostPort(srv.Address, fmt.Sprintf("%d", srv.Port))}
+	return []string{srv.URL}
 }
 
 func newServer(addr string) *server {
```
jacobhaven: `newServer` might make more sense as:
```Go
func newServer(u *url.URL) (*server, error) {
	[...]
}
```
This would avoid parsing the addr, converting it to string, reparsing it, reconverting it to string. And though this isn't your code originally, discarding an `error` value and returning `nil` is a bad practice in Go.

api/client/client.go
```diff
@@ -266,3 +276,21 @@ func NewAuthServer(addr string, provider auth.Provider) *AuthRemote {
...
+
+// nomalizeURL checks for http/https protocol, appends "http" as default protocol if not defiend in url
+func normalizeURL(addr string) string {
+	addr = strings.TrimSpace(addr)
+	u, err := url.Parse(addr)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if !strings.Contains(u.Scheme, "http") {
```
jacobhaven: There's no need to convert to string and then reparse here, you can just check if `u.Scheme==""` and set it to `"https"`, then return the resulting `*url.URL` for later processing.

codyli520: i used string because take for example u = 127.0.0.1:8888, u.Scheme would be "127.0.0.1" not  " ".

jacobhaven: This is not the case. See this example: https://play.golang.org/p/f__Ogobk0i

codyli520: thank you

jacobhaven: Actually, the problem with that example is that URI Scheme's must begin with a letter (so an IP address will never match a Scheme). However, a DNS name can `"ca.example.com:8888"` will parse incorrectly

You can instead check if `u.Opaque` has been set (it never should be for a normal URL) https://play.golang.org/p/3-8Yeu4Ily

cli/serve/serve.go
```diff
@@ -200,8 +202,29 @@ func serverMain(args []string, c cli.Config) error {
...
+	tlsCA := conf.TLSCertFile
+	tlsKey := conf.TLSKeyFile
+
+	if tlsCA == "" || tlsKey == "" {
+		log.Info("Now listening on ", addr)
+		return http.ListenAndServe(addr, nil)
+	}
+
+	tlsConfig := &tls.Config{
+		CipherSuites: nil,
```
jacobhaven: This is a noop. Slices are initialized to `nil` by default in Go.

cli/serve/serve.go
```diff
@@ -200,8 +202,29 @@ func serverMain(args []string, c cli.Config) error {
...
+	}
+
+	s := &http.Server{
+		Addr:      addr,
+		Handler:   nil,
```
jacobhaven: Ditto. No need to manually set this to `nil`

cmd/multirootca/ca.go
```diff
@@ -84,6 +87,24 @@ func main() {
...
+	} else {
+
+		tlsConfig := &tls.Config{
+			CipherSuites: nil,
+		}
+
+		s := &http.Server{
+			Addr:      *flagAddr,
+			Handler:   nil,
+			TLSConfig: tlsConfig,
```
jacobhaven: Since `tlsConfig` is empty, you can leave it as `nil`. 
```Go
s := &http.Server{Addr: *flagAddr}
```

cli/serve/serve.go
```diff
@@ -200,8 +202,29 @@ func serverMain(args []string, c cli.Config) error {
...
+	if tlsCA == "" || tlsKey == "" {
+		log.Info("Now listening on ", addr)
+		return http.ListenAndServe(addr, nil)
+	}
+
+	tlsConfig := &tls.Config{
+		CipherSuites: nil,
+	}
+
+	s := &http.Server{
```
jacobhaven: This can be simplified to just
```Go
s := &http.Server{Addr: addr}
```

cli/serve/serve.go
```diff
@@ -200,8 +202,29 @@ func serverMain(args []string, c cli.Config) error {
 	registerHandlers()
 
 	addr := net.JoinHostPort(conf.Address, strconv.Itoa(conf.Port))
-	log.Info("Now listening on ", addr)
-	return http.ListenAndServe(addr, nil)
+
+	tlsCA := conf.TLSCertFile
```
jacobhaven: I see no reason to create new names for these variables, and `tlsCA` is pretty misleading, given it's a certificate.

api/client/client.go
```diff
@@ -22,8 +23,7 @@ import (
 
 // A server points to a single remote CFSSL instance.
 type server struct {
-	Address string
-	Port    int
+	URL string
```
jacobhaven: In the future, we might consider adding an optional `*tls.Config` parameter and setting an `*http.Client` with that config.
e.g.:
```Go
type server struct {
	URL string
	client *http.Client
}
func newTLSServer(addr string, config *tls.Config) (*server, error) {
	srv, err := newServer(addr)
	if err != nil {
		return nil, err
	}

	srv.client = &http.Client{Transport: &http.Transport{TLSClientConfig: config}}
	return srv, nil
}

NewTLSServer(addr, &tls.Config{RootCAs: rootPool})
```

jacobhaven:
Other than a few comments on things to be cleaned up, this looks good. I'd make those changes, commit and push them, and then `git rebase -i master` and squash all of your commits down to a single commit message (then `git push -f` to update this PR).

grittygrease:
There are still several issues shown by Travis (including a panic), please correct these.

api/client/client.go
```diff
@@ -266,3 +252,39 @@ func NewAuthServer(addr string, provider auth.Provider) *AuthRemote {
...
+		return nil, err
+	}
+
+	if u.Opaque != "" {
+		u.Host = net.JoinHostPort(u.Scheme, u.Opaque)
+		u.Opaque = ""
+	}
+
+	if u.Scheme != "http" {
+		u.Scheme = "http"
```
jacobhaven: I think you missed an `s` here if we want to default to `https`.

This logic currently rewrites `https://` URLs to `http://` which is not correct.

cmd/multirootca/ca.go
```diff
@@ -84,6 +86,14 @@ func main() {
...
 	http.Handle("/api/v1/cfssl/metrics", metrics)
-	log.Info("listening on ", *flagAddr)
-	log.Error(http.ListenAndServe(*flagAddr, nil))
+
+	if *flagEndpointCert == "" && *flagEndpointKey == "" {
+		log.Info("listening on ", *flagAddr)
+		log.Fatal(http.ListenAndServe(*flagAddr, nil))
+	} else {
+
+		log.Info("Now listening on TLS of port ", *flagAddr)
```
jacobhaven: I'd edit this to: `log.Info("Now listening on https://", addr)`

api/client/client.go
codyli520: @jacobhaven , may "http" if protocol not defined?

api/client/client.go
codyli520: working on it

api/client/client_test.go
```diff
@@ -90,13 +90,13 @@ func TestNewServerGroup(t *testing.T) {
...
-	if hosts[0] != "cfssl1.local:8888" {
-		t.Fatalf("expected to see cfssl1.local:8888, but saw %s",
+	if hosts[0] != "http://cfssl1.local:8888" {
+		t.Fatalf("expected to see http://cfssl1.local:8888, but saw %s",
 			hosts[0])
 	}
 
-	if hosts[1] != "cfssl2.local:8888" {
-		t.Fatalf("expected to see cfssl2.local:8888, but saw %s",
+	if hosts[1] != "http://cfssl2.local:8888" {
```
jacobhaven: You should probably add a tests here with explicit schemes in the `NewServer` call expecting `https://[...]` and `http://[...]`

api/client/client.go
jacobhaven: Ah, I see @grittygrease's comment now. You just need to check if the `u.Scheme != "https"` and set it to `http`.

codyli520:
adding more tests.

codyli520:
ready to merge

kisom:
LGTM.

Issue #269, opened 2015-07-20
Return scan results incrementally.


jacobhaven:
Currently, the results from `cfssl scan` are returned either all-together or not at all. In the case of a [timeout](https://github.com/cloudflare/cfssl/issues/235), this also means returning no information about what operations succeeded/didn't complete.

Having more fine-grained output would help debug issues in this case.

taravancil:
I agree.

Do you think the scan should return results incrementally, or alternatively, return whatever results are available if/when the scan times out? Now that scans [run concurrently](https://github.com/cloudflare/cfssl/pull/278), I'm more inclined toward the latter.

Otherwise, individual scan results for different hosts will be returned in sequence, which isn't ideal.

PR #388, opened 2015-10-12
Make scans run concurrently

### LINKED ISSUE #236: Add concurrency to scan on multiple sites

jacobhaven:
This updates PR #278 by @taravancil 

This change runs individual scans from each family concurrently, and
sends the results on a channel when the family scans finish.

Resolves #236 and #269

### DIFF
--- a/cli/scan/scan.go
+++ b/cli/scan/scan.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/cloudflare/cfssl/cli"
+	"github.com/cloudflare/cfssl/log"
 	"github.com/cloudflare/cfssl/scan"
 )
 
@@ -44,21 +45,19 @@ func scanMain(args []string, c cli.Config) (err error) {
 			if err != nil {
 				return
 			}
-			fmt.Printf("Scanning %s...\n", host)
-
-			var resultChan <-chan *scan.Result
-			resultChan, err = scan.Default.RunScans(host, c.IP, c.Family, c.Scanner, c.Timeout)
-			if err != nil {
-				return
-			}
 
 			wg.Add(1)
-			go func(host string, resultChan <-chan *scan.Result) {
-				results := scan.ProcessResults(resultChan)
+			go func(host string) {
+				fmt.Printf("Scanning %s...\n", host)
+				results, err := scan.Default.RunScans(host, c.IP, c.Family, c.Scanner, c.Timeout)
 				fmt.Printf("=== %s ===\n", host)
-				printJSON(results)
+				if err != nil {
+					log.Error(err)
+				} else {
+					printJSON(results)
+				}
 				wg.Done()
-			}(host, resultChan)
+			}(host)
 		}
 		wg.Wait()
 	}
--- a/cli/scan/scan.go
+++ b/cli/scan/scan.go
@@ -29,6 +29,29 @@ func printJSON(v interface{}) {
 	fmt.Printf("%s\n\n", b)
 }
 
+type context struct {
+	sync.WaitGroup
+	c cli.Config
+}
+
+func newContext(c cli.Config, numHosts int) *context {
+	ctx := &context{c: c}
+	ctx.Add(numHosts)
+	return ctx
+}
+
+func (ctx *context) RunScans(host string) {
+	fmt.Printf("Scanning %s...\n", host)
+	results, err := scan.Default.RunScans(host, ctx.c.IP, ctx.c.Family, ctx.c.Scanner, ctx.c.Timeout)
+	fmt.Printf("=== %s ===\n", host)
+	if err != nil {
+		log.Error(err)
+	} else {
+		printJSON(results)
+	}
+	ctx.Done()
+}
+
 func scanMain(args []string, c cli.Config) (err error) {
 	if c.List {
 		printJSON(scan.Default)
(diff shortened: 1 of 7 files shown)


scan/scan_common.go
```diff
@@ -163,57 +170,72 @@ func (fs FamilySet) RunScans(host, ip, family, scanner string, dur time.Duration
...
-		go startTimer(dur, timeout)
-	}
+	results := make(chan *Result)
 
-	familyResults := make(map[string]FamilyResult)
+	var familyWG sync.WaitGroup
+	familyWG.Add(len(fs))
+	go func() {
+		familyWG.Wait()
+		close(results)
```
jacobhaven: Maybe a comment would be good here. This goroutine is spawned to wait for all the scans to finish and then signal this by closing the (already returned) `results` channel

scan/scan_common.go
```diff
@@ -163,57 +170,72 @@ func (fs FamilySet) RunScans(host, ip, family, scanner string, dur time.Duration
...
-
-				if len(scannerResults) > 0 {
-					familyResults[familyName] = scannerResults
+						scannerWG.Done()
+					}(scannerName, scanner)
 				}
 			}
+			go func() {
+				scannerWG.Wait()
+				familyWG.Done()
```
jacobhaven: This Waits for all the Scanners in a Family to finish before marking it as Done in the WaitGroup.

scan/scan_common.go
```diff
@@ -163,57 +170,72 @@ func (fs FamilySet) RunScans(host, ip, family, scanner string, dur time.Duration
...
+			}
+
+			if results[result.Family] == nil {
+				results[result.Family] = make(FamilyResult)
+			}
+
+			results[result.Family][result.Scanner] = result.ScannerResult
+
+		case <-end:
+			return
```
jacobhaven: There's a bit of a hidden bug here. We can't inform the running scanners that we've timed out, so when they will continue to run and block indefinitely on sending to the `results` channel.

(We could close results here and defer/recover on the `send on closed channel` panic inside `RunScans`)

jacobhaven:
I've added a few comments and fixed the channel-close/timeout of running goroutines issue.

cli/scan/scan.go
```diff
@@ -42,17 +46,20 @@ func scanMain(args []string, c cli.Config) (err error) {
 				return
 			}
 
-			fmt.Printf("Scanning %s...\n", host)
+			wg.Add(1)
+			go func(host string) {
+				fmt.Printf("Scanning %s...\n", host)
+				results, err := scan.Default.RunScans(host, c.IP, c.Family, c.Scanner, c.Timeout)
+				if err != nil {
+					log.Fatal(err)
```
jacobhaven: It's a bit annoying to have to `log.Fatal` here instead of just returning the error, but it accomplishes about the same result.

cli/scan/scan.go
```diff
@@ -42,17 +46,20 @@ func scanMain(args []string, c cli.Config) (err error) {
...
-			var results map[string]scan.FamilyResult
-			results, err = scan.Default.RunScans(host, c.IP, c.Family, c.Scanner, c.Timeout)
-			if err != nil {
-				return
-			}
-			if results != nil {
-				printJSON(results)
-			}
+			wg.Add(1)
+			go func(host string) {
```
ashishgandhi: Cloud make it a function instead of using the context from the closure. That way it is clear what it depends on and only things it needs are passed on.

scan/scan_common.go
```diff
@@ -163,57 +170,81 @@ func (fs FamilySet) RunScans(host, ip, family, scanner string, dur time.Duration
...
-				scannerResults := make(map[string]ScannerResult)
+				scannerWG.Add(len(family.Scanners))
 				for scannerName, scanner := range family.Scanners {
-					if scannerRegexp.MatchString(scannerName) {
-						grade, output, err := scanner.Scan(addr, hostname)
-						if grade != Skipped {
-							result := ScannerResult{
-								Grade:  grade.String(),
-								Output: output,
+					go func(scannerName string, scanner *Scanner) {
```
ashishgandhi: Same comment about closure.

jacobhaven: It's just that there are so many constant globals. It's awkward to have to pass around a `*sync.WaitGroup` and `*regexp.Regexp`.

```Go
func (scanner *Scanner) concurrentScan(familyName, scannerName string, scannerRegexp *regexp.Regexp, scannerWG *sync.WaitGroup) {
...
}
```
```Go
go scanner.concurrentScan(familyName, scannerName, scannerRegexp, scannerWG)
````

cli/scan/scan.go
jacobhaven: Same comment about not wanting to pass around a `*sync.WaitGroup` and `&wg`, it's overly-verbose for a constant global.

ashishgandhi: :+1:

jacobhaven:
@ashishgandhi I did a bit more work corralling all the closure data into structs. I named them "contexts", because that seems appropriately generic. What do you think?

ashishgandhi:
LTGM.

jacobhaven:
I might rename a couple things first before merging. (I'm don't think the "contexts" are completely clear)

PR #389, opened 2015-10-14
Allow for email addresses in the 'hosts' list for the SubjectAltName

twitched:
Changes the local signer so that email addresses in the hostname list are properly added to the SubjectAltName as email: email@address.com.  Also adds a to the test.


grittygrease:
Thanks for the PR, it's a nice addition to the package!

jacobhaven:
LGTM

PR #394, opened 2015-10-20
Certificate Transparency support

andrewbuss:
Add a config field "ct-log" to specify the URL of a CT log server

cfssl sign will submit a poisoned certificate to each CT log server 
and embed the resulting SCTs in the certificate before returning 
the certificate to the client. The signing operation will fail if any 
of the CT log servers reject the certificate.

### DIFF
--- a/errors/error.go
+++ b/errors/error.go
@@ -22,38 +22,38 @@ type Category int
 type Reason int
 
 const (
-// Success indicates no error occurred.
+	// Success indicates no error occurred.
 	Success Category = 1000 * iota // 0XXX
 
-// CertificateError indicates a fault in a certificate.
+	// CertificateError indicates a fault in a certificate.
 	CertificateError // 1XXX
 
-// PrivateKeyError indicates a fault in a private key.
+	// PrivateKeyError indicates a fault in a private key.
 	PrivateKeyError // 2XXX
 
-// IntermediatesError indicates a fault in an intermediate.
+	// IntermediatesError indicates a fault in an intermediate.
 	IntermediatesError // 3XXX
 
-// RootError indicates a fault in a root.
+	// RootError indicates a fault in a root.
 	RootError // 4XXX
 
-// PolicyError indicates an error arising from a malformed or
-// non-existent policy, or a breach of policy.
+	// PolicyError indicates an error arising from a malformed or
+	// non-existent policy, or a breach of policy.
 	PolicyError // 5XXX
 
-// DialError indicates a network fault.
+	// DialError indicates a network fault.
 	DialError // 6XXX
 
-// APIClientError indicates a problem with the API client.
+	// APIClientError indicates a problem with the API client.
 	APIClientError // 7XXX
 
-// OCSPError indicates a problem with OCSP signing
+	// OCSPError indicates a problem with OCSP signing
 	OCSPError // 8XXX
 
-// CSRError indicates a problem with CSR parsing
+	// CSRError indicates a problem with CSR parsing
 	CSRError // 9XXX
 
-// CTError indicates a problem with the certificate transparency process
+	// CTError indicates a problem with the certificate transparency process
 	CTError // 10XXX
 )
 
@@ -79,21 +79,21 @@ const (
 // The following represent certificate non-parsing errors, and must be
 // specified along with CertificateError.
 const (
-// SelfSigned indicates that a certificate is self-signed and
-// cannot be used in the manner being attempted.
+	// SelfSigned indicates that a certificate is self-signed and
+	// cannot be used in the manner being attempted.
 	SelfSigned Reason = 100 * (iota + 1) // Code 11XX
 
-// VerifyFailed is an X.509 verification failure. The least two
-// significant digits of 12XX is determined as the actual x509
-// error is examined.
+	// VerifyFailed is an X.509 verification failure. The least two
+	// significant digits of 12XX is determined as the actual x509
+	// error is examined.
 	VerifyFailed // Code 12XX
 
-// BadRequest indicates that the certificate request is invalid.
+	// BadRequest indicates that the certificate request is invalid.
 	BadRequest // Code 13XX
 
-// MissingSerial indicates that the profile specified
-// 'ClientProvidesSerialNumbers', but the SignRequest did not include a serial
-// number.
+	// MissingSerial indicates that the profile specified
+	// 'ClientProvidesSerialNumbers', but the SignRequest did not include a serial
+	// number.
 	MissingSerial // Code 14XX
 )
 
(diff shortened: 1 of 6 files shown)


grittygrease:
Can you run `goimports` on these files to standardize the spacing?

signer/local/local.go
```diff
@@ -254,7 +257,53 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+			return nil, err
+		}
+		sctList = append(sctList, serializedSCT...)
+	}
+
+	var oidSCTList = []int{1, 3, 6, 1, 4, 1, 11129, 2, 4, 2}
+	var SCTListExtension = pkix.Extension{Id: oidSCTList, Critical: false, Value: sctList}
+	certTBS.ExtraExtensions = append(certTBS.ExtraExtensions, SCTListExtension)
+
+	cert, err = s.sign(&certTBS, profile)
```
lziest: can we simply do "return  s.sign(&certTBS, profile)"?

signer/local/local.go
```diff
@@ -254,7 +257,53 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		safeTemplate.SerialNumber = serialNumber
 	}
 
-	return s.sign(&safeTemplate, profile)
+	var certTBS = safeTemplate
+	if len(profile.CTLogServers) == 0 {
+		cert, err = s.sign(&certTBS, profile)
+		return
+	}
+
```
lziest: the below block of dealing with CTLogServers can be included with "if len(profile.CTLogServers) > 0", so we don't need to write the bail-out early branch logic above.

signer/local/local.go
andrewbuss: If the signature fails, we need to wrap the error in a cferr before returning it.

lziest: I thought s.sign should always return a cferr if there is error. And double wrapping actually triggers a intentional panic.

errors/error.go
```diff
@@ -329,6 +332,13 @@ func New(category Category, reason Reason) *Error {
 		default:
 			panic(fmt.Sprintf("Unsupported CF-SSL error reason %d under category APIClientError.", reason))
 		}
+	case CTError:
+		switch reason {
+		case Unknown:
+			msg = "Certificate transparency parsing failed due to unknown error"
+		default:
+			panic(fmt.Sprintf("Unsupported CF-SSL error reason %d under category APIClientError.", reason))
```
lziest: APIClientError -> CTError

errors/error.go
```diff
@@ -22,36 +22,39 @@ type Category int
 type Reason int
 
 const (
-	// Success indicates no error occurred.
```
lziest: gofmt?

errors/error.go
```diff
@@ -22,36 +22,39 @@ type Category int
...
-	// OCSPError indicates a problem with OCSP signing
+// OCSPError indicates a problem with OCSP signing
 	OCSPError // 8XXX
 
-	// CSRError indicates a problem with CSR parsing
+// CSRError indicates a problem with CSR parsing
 	CSRError // 9XXX
+
+// CTError indicates a problem with the certificate transparency process
+	CTError // 10XXX
```
kisom: This should also be added to [`/doc/errorcode.txt`](https://github.com/cloudflare/cfssl/blob/master/doc/errorcode.txt).

signer/local/local.go
```diff
@@ -254,7 +259,66 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		safeTemplate.SerialNumber = serialNumber
 	}
 
-	return s.sign(&safeTemplate, profile)
+	var certTBS = safeTemplate
+
+	if len(profile.CTLogServers) > 0 {
+		// Add a poison extension which prevents validation
+		var oidCTPoison = []int{1, 3, 6, 1, 4, 1, 11129, 2, 4, 3}
```
kisom: This should be a constant or global variable somewhere.

signer/local/local.go
```diff
@@ -254,7 +259,66 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+
+		var sctList []ct.SignedCertificateTimestamp;
+
+		for _, server := range profile.CTLogServers {
+			var ctclient = client.New(server)
+			var resp, err = ctclient.AddPreChain(prechain)
+			if err != nil {
+				return nil, cferr.Wrap(cferr.CTError, cferr.Unknown, err)
+			}
+			log.Info(resp)
```
kisom: It would be useful to add some additional information to this log line: what's being logged and why. Perhaps only log some information, or use formatting to show what the fields are. Otherwise, this should be a `Debug` level log, and shouldn't be shown to users normally.

signer/local/local.go
```diff
@@ -254,7 +259,66 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+			sctList = append(sctList, *resp)
+		}
+
+		var serializedSCTList, err = serializeSCTList(sctList)
```
kisom: Error isn't checked here.

signer/local/local.go
```diff
@@ -254,7 +259,66 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+		serializedSCTList, err = asn1.Marshal(serializedSCTList)
+
```
kisom: Don't need this newline.

signer/local/local.go
```diff
@@ -22,6 +22,11 @@ import (
 	"github.com/cloudflare/cfssl/info"
 	"github.com/cloudflare/cfssl/log"
 	"github.com/cloudflare/cfssl/signer"
+
+	"encoding/asn1"
+	"encoding/binary"
```
lziest: final nits: "encoding/asn1" and "encoding/binary" should be grounded with standard libraries above.

kisom: `go get golang.org/x/tools/cmd/goimports && goimports signer/local/local.go` will fix this.

signer/local/local.go
```diff
@@ -254,7 +259,66 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+		}
+
+		derCert, _ := pem.Decode(cert)
+		prechain := []ct.ASN1Cert{derCert.Bytes, s.ca.Raw}
+		var sctList []ct.SignedCertificateTimestamp
+
+		for _, server := range profile.CTLogServers {
+			log.Infof("submitting poisoned precertificate to %s", server)
+			var ctclient = client.New(server)
+			var resp, err = ctclient.AddPreChain(prechain)
```
kisom: Generally, we prefer not to have multiple variable assignment inside a `var` statement; in this case given that `err` is already defined, means that it is getting shadowed here. The preference is

```
var resp *ct.SignedCertificateTimestamp
resp, err = ctclient.AddPreChain(prechain)
```

signer/local/local.go
```diff
@@ -254,7 +259,66 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+			if err != nil {
+				return nil, cferr.Wrap(cferr.CTError, cferr.PrecertSubmissionFailed, err)
+			}
+			sctList = append(sctList, *resp)
+		}
+
+		var serializedSCTList, err = serializeSCTList(sctList)
```
kisom: Same `var` multiple assignment / shadowing issue.

signer/local/local.go
```diff
@@ -254,7 +259,66 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+		certTBS.ExtraExtensions = append(certTBS.ExtraExtensions, SCTListExtension)
+	}
+	return s.sign(&certTBS, profile)
+}
+
+func serializeSCTList(sctList []ct.SignedCertificateTimestamp) ([]byte, error) {
+	var buf bytes.Buffer
+	var totalSCTLen = 0
+	for _, sct := range sctList {
+		var sct, err = ct.SerializeSCT(sct)
```
kisom: As a style preference, here we'd prefer `sct, err :=` instead of `var sct, err =`.

kisom:
A few style nits, but otherwise LGTM.

andrewbuss:
Thanks all for the comments. I'm wondering how and how much this should be tested. I've set up a CT log server for my own testing, but I don't think it's a great idea to add a test which depends on that. The certificate transparency client in https://github.com/google/certificate-transparency/ will keep retrying until a submission succeeds, which is inconvenient for testing without a network connection. 

To document my local testing:

I validated this functionality locally by signing a certificate with a policy specifying my log server (which has been configured to trust my testing CA certificate). CFSSL submitted the precertificate and embedded the resulting SCT in the final certificate before signing. I verified that the SCT list was parsed correctly by openssl and Chrome, although Chrome didn't identify the log server by name (expected).

image

signer/local/local.go
```diff
@@ -56,7 +62,7 @@ func NewSigner(priv crypto.Signer, cert *x509.Certificate, sigAlgo x509.Signatur
 
 // NewSignerFromFile generates a new local signer from a caFile
 // and a caKey file, both PEM encoded.
-func NewSignerFromFile(caFile, caKeyFile string, policy *config.Signing) (*Signer, error) {
+func  NewSignerFromFile(caFile, caKeyFile string, policy *config.Signing) (*Signer, error) {
```
grittygrease: Can you set up your text editor to run goimports for every save?

grittygrease:
Any comments, @lziest or @kisom ? Once the gofmt issue is fixed, squash the commits and we should be good to go.

lziest:
lgtm!

signer/local/local.go
kisom: This escaped through --- once this is fixed, LGTM.

andrewbuss: Sorry - made that automatic now.

kisom:
LGTM. I'll merge once CI is green.

PR #398, opened 2015-10-23
api test suite WIP 

anniephan:
All endpoints are tested except for Authsign.


api/api_test_suite.go
```diff
@@ -0,0 +1,206 @@
...
+		panic(err)
+	}
+	tempFile, err := createTempFile(jsonBytes)
+	if err != nil {
+		os.Remove(tempFile)
+		panic(err)
+	}
+
+	// Now we make the request and check the output.
+	remoteServerString := "-remote=" + addressToTest + ":" + strconv.Itoa(portToTest)
```
j-delaney: `fmt.Sprintf` can be used to make this a bit cleaner:

    remoteServerSTring := fmt.Sprintf("-remote=%s:%s", addressToTest, strconv.Itoa(portToTest))

grittygrease:
Looks like Travis is having issues.

api/api_test_suite.go
```diff
@@ -0,0 +1,206 @@
+package main
```
kisom: This needs to be `package api`.

grittygrease:
Can you rebase and see if Travis still fails? Also, check with @j-delaney on how he fixed his issues relating to timing.

api/api_test_suite.go
```diff
@@ -0,0 +1,206 @@
...
+	"fmt"
+	"github.com/cloudflare/cfssl/csr"
+	"github.com/cloudflare/cfssl/helpers/testsuite"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+)
```
j-delaney: I think the order for the imports is supposed to be things from Go's standard library and then other libraries. So it should be like:

    import (
        "bytes"
        "encoding/json"
        "fmt"
        "io/ioutil"
        "net/http"
        "os"
        "os/exec"
        "strconv"
        "strings"

        "github.com/cloudflare/cfssl/csr"
        "github.com/cloudflare/cfssl/helpers/testsuite"
    )

api/api_test_suite.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	return nil
+}
+
+// Returns the cleaned up PEM encoding for the item specified (for example,
+// 'cert' or 'key').
+func cleanCLIOutput(CLIOutput []byte, item string) (cleanedOutput []byte, err error) {
+	outputString := string(CLIOutput)
+	// The keyword will be surrounded by quotes.
+	itemString := "\"" + item + "\""
```
j-delaney: If you want to avoid escaping quotations, you can use ` instead

api/api_test_suite.go
anniephan: Noted.

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+var addressToTest = "127.0.0.1"
+var portToTest = 8888
+var postStr = fmt.Sprintf("http://%s:%s", addressToTest, strconv.Itoa(portToTest))
+var hosts = []byte(`{"hosts":["www.example.com"],
+					   "names":[{"C":"US", "ST":"California", "L":"San Francisco", "O":"example.com"}],
+					   "CN": "www.example.com"}`)
+var requestStr = []byte(`{ "request": {"hosts":["www.example.com"], 
+						"names":[{"C":"US", "ST":"California", "L":"San Francisco", "O":"example.com"}], 
+						"CN": "www.example.com"} }`)
+var certRequest = []byte(`{"certificate_request": "-----BEGIN CERTIFICATE REQUEST-----\nMIIBUjCB+QIBADBqMQswCQYDVQQGEwJVUzEUMBIGA1UEChMLZXhhbXBsZS5jb20x\nFjAUBgNVBAcTDVNhbiBGcmFuY2lzY28xEzARBgNVBAgTCkNhbGlmb3JuaWExGDAW\nBgNVBAMTD3d3dy5leGFtcGxlLmNvbTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IA\nBK/CtZaQ4VliKE+DLIVGLwtSxJgtUKRzGvN1EwI3HRgKDQ3l3urBIzHtUcdMq6HZ\nb8jX0O9fXYUOf4XWggrLk1agLTArBgkqhkiG9w0BCQ4xHjAcMBoGA1UdEQQTMBGC\nD3d3dy5leGFtcGxlLmNvbTAKBggqhkjOPQQDAgNIADBFAiAcvfhXnsLtzep2sKSa\n36W7G9PRbHh8zVGlw3Hph8jR1QIhAKfrgplKwXcUctU5grjQ8KXkJV8RxQUo5KKs\ngFnXYtkb\n-----END CERTIFICATE REQUEST-----\n"}`)
```
andrewbuss: I wonder if these long strings could be moved to testdata files and their contents referenced by filename as in https://github.com/cloudflare/cfssl/blob/master/helpers/helpers_test.go#L16

anniephan: Good idea. That could be arranged.

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+			},
+		},
+		Hosts:      []string{"ca.example.com"},
+		KeyRequest: &keyRequest,
+		CA:         &CAConfig,
+	}
+)
+
+/* We will Unmarshal API and CLI outputs into the following structs */
+type Response struct {
```
j-delaney: Is it possible to just use the `Response` struct found in `api.go`?

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+/* Start up the CFSSL local server. */
+func setup() {
+	// Set up a test server using our CA certificate and key.
+	CACert, CAKey, err := testsuite.CreateSelfSignedCert(CARequest)
+	if err != nil {
+		panic(err)
+	}
+	// Set up a test server using our CA certificate and key.
+	serverData := testsuite.CFSSLServerData{CA: CACert, CAKey: CAKey}
+	s, err := testsuite.StartCFSSLServer(addressToTest, portToTest, serverData)
```
j-delaney: We can just do `server, err := ...` and remove the next line

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+	}
+}
+
+/* Kill the CFSSL local server.*/
+func setdown() {
+	err := server.Kill()
+	if err != nil {
+		panic(err)
+	}
+	return
```
j-delaney: I don't think you need a `return` if there is nothing to return

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+	server = s
+	if err != nil {
+		setdown()
+		panic(err)
+	}
+}
+
+/* Kill the CFSSL local server.*/
+func setdown() {
+	err := server.Kill()
```
j-delaney: If you want to make this more concise you could do:

    if err := server.Kill(); err != nil {
        panic(err)
    }

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return
+}
+
+/* Return API output*/
+func apiOut(request string, jsonStr []byte, endpoint string, t *testing.T) []byte {
+	setup()
+	req, err := http.NewRequest(request, endpoint, bytes.NewBuffer(jsonStr))
```
j-delaney: Does this `err` need to be checked? If not you can replace `err` with `_`

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		setdown()
+		t.Fatalf("Error sending request to client, %v", err)
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		setdown()
+		t.Fatalf("Error reading API response, %v", err)
```
j-delaney: Because `setdown()` calls `panic()`, will this `t.Fatalf` ever be run?

cmd/cfssl/cfssl_test.go
```diff
@@ -1 +1,350 @@
...
+	json.Unmarshal(body, &res)
+	return res
+}
+
+/* START OF TESTS*/
+
+/* Testing the Bundle API against CFSSL Bundle CLI */
+func TestBundle(t *testing.T) {
+	var jsonStr = []byte(`{"domain": "example.com"}`)
+	bundleAPIResp := apiFormatBundle(apiOut("POST", jsonStr, postStr+"/api/v1/cfssl/bundle", t))
```
j-delaney: Since it seems like every all to `apiOut` uses `postStr + "/api/v1/cfssl"`, perhaps `apiOut` could put those in automatically to make the code a bit cleaner

grittygrease:
To answer #1, yes there is a structure it's composed here: https://github.com/cloudflare/cfssl/blob/master/scan/scan_common.go#L132

cmd/cfssl/cfssl_test.go
anniephan: For some reason I completely missed that lol. Thanks!

cmd/cfssl/cfssl_test.go
anniephan: That could be arranged, thanks.

cmd/cfssl/cfssl_test.go
anniephan: Though it's not necessarily always that panic() will get called from setdown() right.

cmd/cfssl/cfssl_test.go
anniephan: I just added a check for err, thanks.

kisom:
@anniephan I'll add a more detailed example in a follow on PR, but in regards to question 2, you need to do something like

```
var req signer.SignRequest
// fill in req
out, err := json.Marshal(req)
if err != nil {
        // assuming return []byte, error
        return nil, err
}

// id isn't used currently.
return client.AuthSign(req, nil, provider) 
```

anniephan:
@kisom Thanks! I'll look into that.

cmd/cfssl/cfssl_test.go
anniephan: ```:=``` apparently will both assign and initialize a variable in Go. We'd want to assign an existing variable ```server``` and initialize/assign nonexisting variable ```err``` at the same time. It looks like setting ```server``` equal to the initialized return of ```testsuite.StartCFSSLServer``` is the best option in this case.

cmd/cfssl/cfssl_test.go
anniephan: Noted.

PR #400, opened 2015-10-26
Add AuthRemote with default auth.Provider for simpler AuthSign requests.

jacobhaven:


api/client/client.go
```diff
@@ -245,3 +245,24 @@ func (srv *server) request(jsonData []byte, target string) ([]byte, error) {
...
+// in the same format at NewServer and a default authentication provider to
+// use for DefaultAuthSign requests.
+func NewAuthServer(addr string, provider auth.Provider) *AuthRemote {
+	return &AuthRemote{
+		Remote:   NewServer(addr),
+		provider: provider,
+	}
+}
+
+// DefaultAuthSign performs an AuthSign request using the default auth provider.
```
kisom: You should overload `Sign` instead of providing a new function.

jacobhaven: Agreed, that's much better. *changed*

kisom:
LGTM.

PR #407, opened 2015-10-30
Fix serve UI bug (404 on direct access of '/scan' or '/bundle') 

jacobhaven:
This sets a redirect mapping in the httpBox for those specific endpoints.


lziest:
lgtm

PR #409, opened 2015-10-30
refactor cmd/multirootca/config to rootconfig

lziest:
- Export RawMap type from rootconfig
- Export LoadRoot function from rootconfig
- Enable selective parsing of a root


grittygrease:
Is "rootconfig" the right name?

lziest:
I need suggestions

grittygrease:
@kisom what do you think this package should be called?

kisom:
@lziest @grittygrease since this is concerned with the configuration of labels, it might make sense to name it something like `labelconfig`.

grittygrease:
How about a "multiroot" package with a "config" sub-package.

kisom:
I like that idea.

lziest:
+1

grittygrease:
LGTM

kisom:
LGTM.

PR #308, opened 2015-08-14
Vendor dependencies.

kisom:
This runs [godep](https://github.com/tools/godep) on the repository.

PR #412, opened 2015-11-03
Moved CT log client support into a separate optional package

andrewbuss:
The CT dependency does not build on OS X at present
(https://github.com/google/certificate-transparency/issues/969)

Build with the ctclient tag to support CT log submission


jacobhaven:
We got this bug fixed in upstream, I'm not sure the new build tag will be necessary.

andrewbuss:
Oh cool! I don't have a Mac to check, but if someone can verify that the problem's resolved I'll close this PR.

signer/ctclient/ctclient.go
```diff
@@ -0,0 +1,64 @@
...
+
+package ctclient
+
+import (
+	"bytes"
+	"encoding/asn1"
+	"encoding/binary"
+	"encoding/pem"
+
+	"crypto/x509"
```
j-delaney: Should this line come after `"bytes"`?

signer/ctclient/ctclient.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, cferr.Wrap(cferr.CTError, cferr.Unknown, err)
+	}
+
+	// Serialize again as an octet string before embedding
+	serializedSCTList, err = asn1.Marshal(serializedSCTList)
+	if err != nil {
+		return nil, cferr.Wrap(cferr.CTError, cferr.Unknown, err)
+	}
+
+	return serializedSCTList, nil
```
j-delaney: Because you're using named returns you could also just do `return` and Go would figure out which vars to use.

signer/ctclient/ctclient.go
kisom: Consider running `goimports` on the code before committing --- it provides a canonical ordering for packages, among other things.

andrewbuss: `goimports` ran on this and didn't touch the import ordering

anniephan: Try taking out that newline in line 10.

andrewbuss: Ah; didn't realize that mattered to goimports. I think this whole PR might be moot if the build works on Mac now, but I'll wait for someone to confirm before closing.

andrewbuss:
In any case, fixed nits.

jacobhaven:
There's [currently a bug](https://travis-ci.org/cloudflare/cfssl/builds/89327249) in the test script's use of `xargs`, but the vendored certificate transparency code compiles and tests pass on OS X (#308). Separating them out with build tags shouldn't be necessary.

PR #415, opened 2015-11-05
The certificate-transparency go dependency is required for the build

xynova:
Fix for issue: cannot find package "github.com/google/certificate-transparency/go/client" in any of... when running /script/build


jacobhaven:
I think this will be better handled by vendoring dependencies (PR #308) and then in the Dockerfiles installing godep and running `godep restore`.

jacobhaven:
This should be fixed in all the Dockerfiles by #419

PR #419, opened 2015-11-10
Update Dockerfiles

jacobhaven:
Trying to simplify and standardize all of the Dockerfiles as much as possible.

We've vendored all the cfssl build dependencies, we might want to vendor github.com/GeertJohan/go.rice/rice github.com/mitchellh/gox and github.com/cloudflare/cfssl_trust as well.


cli/serve/serve.go
```diff
@@ -83,7 +83,6 @@ func (hb *httpBox) Open(name string) (http.File, error) {
 
 // staticBox is the box containing all static assets.
 var staticBox = &httpBox{
-	Box: rice.MustFindBox("static"),
```
jacobhaven: Apparently `rice.MustFindBox` can panic if called before/at init-time.

Dockerfile.build
```diff
@@ -1,4 +1,6 @@
-FROM golang:1.4.3-cross
+FROM golang:1.5
+
+ENV USER root
 
 # Install pkcs11 deps
```
jacobhaven: These probably aren't necessary as `script/build` builds with `-tags nopkcs11` (because there's not support across all platforms?)

cli/serve/serve.go
```diff
@@ -153,6 +157,10 @@ var endpoints = map[string]func() (http.Handler, error){
 	},
 
 	"/": func() (http.Handler, error) {
+		if err := staticBox.findBox("static"); err != nil {
```
jacobhaven: Handled this way, API endpoints can still be served even if `static` assets can't be found.

kisom:
LGTM.

PR #422, opened 2015-11-11
SQL database backend support

andrewbuss:
Uses a hardcoded SQLite db.

When a certificate is signed, it is recorded in the cert store.

Later, the revoke tool can be used to revoke a certificate.

cfssl ocspgen will generate a file of concatenated OCSP responses for
all unexpired certificates in the cert store.

I'm opening the PR to gather feedback on this very partial implementation and run CI tests


certstore/certstore.go
```diff
@@ -0,0 +1,120 @@
...
+type Certificate struct {
+	AsPEM            string
+	Serial           string
+	SignedAt         time.Time
+	Expiration       time.Time
+	RevokedAt        *time.Time
+	RevocationReason sql.NullInt64
+}
+
+var certificatesSchema = `
```
lziest: I have a idea of separating DB creation from DB querying.  I will create a PR for DB creation (aka DB migration 'for the first time')

certstore/certstore.go
```diff
@@ -0,0 +1,120 @@
...
+		serial VARCHAR(50) PRIMARY KEY,
+		signedAt TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+		expiration TIMESTAMP,
+		revokedAt TIMESTAMP NULL DEFAULT NULL,
+		revocationReason INTEGER NULL DEFAULT NULL
+	);
+`
+
+func initializeDB() (db *sql.DB, err error) {
+	db, err = sql.Open("sqlite3", "./certstore.db")
```
lziest: let's assume the db passed in is in a usable state.

certstore/certstore.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, cferr.New(cferr.CertStoreError, cferr.DatabaseInitializationFailed)
+	}
+
+	_, err = db.Exec(certificatesSchema)
+	return db, err
+}
+
+// RecordCert records a new signed cert in the certificate store
+func RecordCert(cert x509.Certificate, signedCert []byte) (err error) {
+	db, err := initializeDB()
```
lziest: I think we can skip this, and get db *sql.DB as part of the pass-in arguments. We just assume db works. Keep the logic of connecting to db separated.

certstore/certstore.go
```diff
@@ -0,0 +1,120 @@
...
+		return cferr.New(cferr.CertStoreError, cferr.RecordCertFailed)
+	}
+
+	return
+}
+
+// RevokeCert revokes a cert with a given serial number and reason code
+func RevokeCert(id string, reasonCode int) (err error) {
+	var db *sql.DB
+	db, err = initializeDB()
```
lziest: ditto.

certstore/certstore.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return
+}
+
+// GetUnexpiredCerts returns all certs in the store which have not expired
+// and information about revocation if applicable
+func GetUnexpiredCerts() (certs []Certificate, err error) {
+	var db *sql.DB
+	db, err = initializeDB()
```
lziest: ditto

lziest:
I have my initial batch of db CRUD in https://github.com/cloudflare/cfssl/commit/8810ff8aba1bb426b8dca24fa7f54e0927cc65b6

please criticize. :)

One thing about the db schema is we want to support multiple ca, so we have 'CALabel' with each certificate, and we can use the 'CALabel' to work with the multirootca to get ocsp response signed.

certstore/certstore.go
lziest: my take on the certificate structure
https://github.com/cloudflare/cfssl/commit/8810ff8aba1bb426b8dca24fa7f54e0927cc65b6

andrewbuss:
Thanks for the comments - apologies that I wasn't able to respond to your PM earlier. Your CRUD is certainly cleaner with sqlstruct. The CALabel makes sense.

Where would be best to initialize the database when signing? We'd need to have the signer config in scope at the time. Should we set up the database when we create a local signer from a config file, and then store a reference in the signer struct? (Also, do we want to store certificates when we use a remote signer?)

lziest:
my initial take is have a cmd config for db connections for multirootca. My idea is to separate DB creation/initialization from cfssl. Interaction with DB would be somewhat similar to interact with the OS and file system, failure may happen but cfssl won't address that directly.

remote signer probably doesn't need to store certificate so far. local signer can store certificate with db config. The db config can be as simple as {"driver":"sqlite3", "source": "/path/to/file"}. I think we don't need to add db config to the existing signing config. It can be a standalone config.

andrewbuss:
Could you elaborate on "have a cmd config for db connections for multirootca" ? I'm not sure what that means. Would the standalone config be passed as a second config file to cfssl?

lziest:
yes, a stand alone config be passed to cfssl.

andrewbuss:
The latest commits are broken - I ran into some git and "go get" issues and needed to push to my branch before they were complete. 

The primary change is that I'm storing a sql.DB reference with the local signer struct and initializing it inside SignerFromConfig if db-config is passed as a command line argument. I'm planning to do the same for revoke and ocspgen.

lziest:
thanks!

certdb/pg/migrations/001_CreateCertificates.sql
```diff
@@ -0,0 +1,17 @@
...
+-- SQL in section 'Up' is executed when this migration is applied
+
+CREATE TABLE certificates (
+  id                serial,
+  serial            bytea NOT NULL PRIMARY KEY,
+  ca_label          bytea NOT NULL,
+  status            bytea NOT NULL,
+  reason            int,
+  revoked_at        timestamptz,
+  pem               bytea NOT NULL,
```
lziest: consider at expired_at column

certdb/sqlite/migrations/001_CreateCertificates.sql
```diff
@@ -0,0 +1,18 @@
...
+-- SQL in section 'Up' is executed when this migration is applied
+
+CREATE TABLE certificates (
+  id                serial,
+  serial            bytea NOT NULL PRIMARY KEY,
+  ca_label          bytea NOT NULL,
+  status            bytea NOT NULL,
+  reason            int,
+  revoked_at        timestamptz,
+  pem               bytea NOT NULL,
```
lziest: consider add expired_at column

kisom:
@andrewbuss This will need to be rebased.

api/revoke/revoke_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	err = certdb.InsertCertificate(db, cert)
+	if err != nil {
+		return nil, err
+	}
+
+	return
+}
+
+func testRevokeCert(t *testing.T, db *sql.DB, serial string, reason *string) (resp *http.Response, body []byte) {
```
lziest: just nitpick: I think it is fine to use "reason string", the empty value of string is ""

cli/ocspsign/ocspsign.go
```diff
@@ -44,8 +47,16 @@ func ocspSignerMain(args []string, c cli.Config) (err error) {
 	}
 
 	if c.Status == "revoked" {
-		req.Reason = c.Reason
+		reasonCode, present := ocsp.RevocationReasonCodes[strings.ToLower(c.Reason)]
+		if !present {
+			reasonCode, err = strconv.Atoi(c.Serial)
```
lziest: Is this intended? Doesn't feel right.

api/revoke/revoke.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if len(req.Serial) == 0 {
+		return errors.NewBadRequestString("serial number is required but not provided")
+	}
+
+	if req.Reason == "" {
+		req.Reason = "0"
+	}
+
+	reasonCode, present := ocsp.RevocationReasonCodes[strings.ToLower(req.Reason)]
```
lziest: since ocsp.RevocationReasonCodes[strings.ToLower(req.Reason)] is used for many times, can we make it a exported helper function in ocsp?

andrewbuss:
Addressed comments and rebased

grittygrease:
What do you suggest the sequence of events be to merge this and https://github.com/cloudflare/cfssl/pull/433, @lziest ?

lziest:
#433 first, then I can rebase this on top of it and resolve conflicts.

andrewbuss:
Superseded by #458

PR #424, opened 2015-11-13
rice requires the value specifed to FindBox to be a string literal, not

robdaemon:
a variable. Since the only use of FindBox is to look up the static
folder, make it a static parameter.

### DIFF
--- a/cli/serve/serve.go
+++ b/cli/serve/serve.go
@@ -68,8 +68,8 @@ type httpBox struct {
 	redirects map[string]string
 }
 
-func (hb *httpBox) findBox(name string) (err error) {
-	hb.Box, err = rice.FindBox(name)
+func (hb *httpBox) findStaticBox() (err error) {
+	hb.Box, err = rice.FindBox("static")
 	return
 }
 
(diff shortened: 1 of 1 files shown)


robdaemon:
When a build is performed via `docker build .` or by following the instructions on the repo, you get the following error back from rice:

```
Removing intermediate container a022b81db291
Step 6 : RUN go get github.com/tools/godep && godep restore && 	go get github.com/GeertJohan/go.rice/rice && rice embed-go -i=./cli/serve && 	git clone https://github.com/cloudflare/cfssl_trust.git /etc/cfssl && 	go install ./cmd/...
 ---> Running in 868f5933d2e0
cli/serve/serve.go:72: Error: found call to rice.FindBox, but argument must be a string literal.
```

jacobhaven:
This is very odd, the build worked for me before. But there is some strange undocumented behavior in go.rice that looks at the Go AST for string literals... for some reason? https://github.com/GeertJohan/go.rice/blob/master/rice/find.go#L83

This PR should fix the bug, thanks! LGTM.

PR #425, opened 2015-11-13
Set caching headers in OCSP responder

jsha:
Also, rename response.pem to response.der, put working responses in resp64.pem,
and move some responder tests into responder_test.go


jsha:
Looks like I need to vendorize the new dep, but when I `godep save -r ./...`, I get a bunch of deletions in Godep.json. Is there a particular way you invoke `godep` on this project?

jsha:
Okay, tests fixed by manually adding an entry in Godeps.json.

lziest:
lgtm

lziest:
Is it possible to squash the commit into one?

kisom:
LGTM.

ocsp/responder.go
```diff
@@ -159,7 +171,20 @@ func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Reques
...
+		response.Write(unauthorizedErrorResponse)
+		return
+	}
+
 	// Write OCSP response to response
+	response.Header().Add("Last-Modified", parsedResponse.ProducedAt.Format(time.RFC1123))
+	response.Header().Add("Expires", parsedResponse.NextUpdate.Format(time.RFC1123))
+	maxAge := int64(parsedResponse.NextUpdate.Sub(rs.clk.Now()) / time.Second)
+	if maxAge > 0 {
+		response.Header().Add("Cache-Control", fmt.Sprintf("max-age=%d", maxAge))
```
tbroyer: `Cache-Control` should also include `public,no-transform,must-revalidate`, and the `max-age` should be a bit earlier than `NextUpdate` such that a client observing HTTP cache semantics will refresh before its cached response expires (`NextUpdate`-wise)

See https://tools.ietf.org/html/rfc5019#section-6 for details. Note that it also suggests to include an `ETag` and RECOMMENDS a value for it.

tbroyer: Oh, and `Last-Modified` should be `ThisUpdate` rather than `ProducedAt` per the same spec.

Issue #390, opened 2015-10-15
Accept TLS Feature (aka OCSP Must Staple) in CSRs


jsha:
The spec that began life as OCSP Must Staple has landed now as TLS Feature: https://tools.ietf.org/html/rfc7633. It would be great if CFSSL supported copying this extension from CSRs. Since it's a new extension, it will probably require some custom ASN.1 marshaling, at least until Go adds native support.

grittygrease:
@jsha, is this covered by https://github.com/cloudflare/cfssl/pull/428 ?

jsha:
Yep!

PR #428, opened 2015-11-18
Add support for client-provided extensions

bifurcation:
This PR makes a couple of changes to allow a client to have certain extensions copied from the CSR to the certificate.

* Adds a list of extension OIDs to the `CSRWhitelist` struct to indicate what is safe to copy
* In `signer.ParseCertificateRequest`, move requested extensions to `ExtraExtensions` in the cert
* In `local.Sign`, only allow through the extensions that are whitelisted

The only special consideration is for `subjectAltName`, which is already handled by `x509.ParseCertificate`. A couple of things to check:

* If there is no whitelist, this allows the client to specify any extension in the certificate (besides SAN).  That seems consistent with the existing semantic of a whitelist being absent, though.
* Disallowed extensions are currently silently ignored, rather than raising an error.


bifurcation:
On further consideration, I refactored this to take extensions explicitly in the request, rather than from the CSR.  So instead of the above changes:

* Adds a list of authorized extension OIDs to `config.SigningProfile`
* Adds a list of requested extensions to `signing.SigningRequest`
* In `local.Sign` copies requested, authorized extensions to the template

In addition, the default is now to accept no extensions, and `local.Sign()` returns an error if an unauthorized extension is provided.

signer/local/local.go
```diff
@@ -260,6 +261,47 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+
+			safeTemplate.ExtraExtensions = append(safeTemplate.ExtraExtensions, pkix.Extension{
+				Id:       oid,
+				Critical: ext.Critical,
+				Value:    rawValue,
+			})
+		}
+	}
+
+	/*
```
kisom: Probably should excise this commented out section.

signer/local/local.go
```diff
@@ -260,6 +261,47 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		safeTemplate.SerialNumber = serialNumber
 	}
 
+	if len(profile.AllowedExtensions) > 0 && len(req.Extensions) > 0 {
+		extFilter := map[string]bool{}
+		for _, oid := range profile.AllowedExtensions {
+			extFilter[asn1.ObjectIdentifier(oid).String()] = true
```
kisom: Maybe this should be set up when the profile is loaded, and an "ExtensionAllowed" method added to avoid having to construct this table on each signature.

signer/local/local_test.go
```diff
@@ -885,6 +887,69 @@ func TestNameWhitelistSign(t *testing.T) {
 
 }
 
+func TestExtensionSign(t *testing.T) {
```
kisom: Can you add a test that validates the behaviour where by default, no extra extensions are permitted?

config/config.go
```diff
@@ -77,6 +77,7 @@ type SigningProfile struct {
 	NameWhitelistString string     `json:"name_whitelist"`
 	AuthRemote          AuthRemote `json:"auth_remote"`
 	CTLogServers        []string   `json:"ct_log_servers"`
+	AllowedExtensions   []OID      `json:"allowed_extensions"`
```
kisom: AllowedExtensions should probably go into the CSRWhitelist.

kisom:
This should cover #390.

config/config.go
bifurcation: Actually, that's how I did it in my initial PR, and I think that's *not* the way to go here.  Instead of copying extensions from the CSR, we should have the calling code provide its own extensions.  That gives the calling code a lot more flexibility in what it can add, and avoids taking extension data from an untrusted source.

bifurcation:
@kisom ping

kisom:
On second thought, I agree with the idea of not putting this in the CSR, and leaving it to some policy layer passing the CSR to CFSSL along with any extensions it has extracted from the CSR. Looks like there's a merge conflict now though, once that's cleaned up, LGTM.

signer/signer.go
```diff
@@ -32,15 +32,25 @@ type Subject struct {
 	Names []csr.Name `json:"names"`
 }
 
+// Extension represents a raw extension to be included in the certificate.  The
+// "value" field must be Base64 encoded, using the URLEncoding variant (i.e.,
```
jsha: It looks like CFSSL uses `StdEncoding` throughout, so it would be more consistent here to use that.

Alternately, we could use hex encoding, which might make the DER structure of extensions slightly more accessible if you recognize them. I could flip a coin either way, but I think URLEncoding is wrong.

Also, can you specify the semantics of what happens if an extension is present in both the CSR and in the SignRequest? Does one take precedence? Do they both get appended?

jsha:
It's a bit odd having both the CSRWhitelist and the ExtensionWhitelist. I think Boulder is probably the only user of CSRWhitelist (I added it a few months ago). We could change the Boulder code to extract the extensions we are willing to copy from the CSR, and pass them all in the new Extensions field of SignRequest. That would allow us to do away with the CSRWhitelist code (after a deploy cycle).

signer/signer.go
bifurcation: Re: Encoding, I think hex would actually be nice.  For those schooled in the DER arts, it can actually be intelligible.

Re: CSR vs. SignRequest: The upshot of this change is that extensions in the CSR are ignored except for the ones processed by `ParseCertificateRequest` (primarily subjectAltName).  So the calling code is responsible for moving any extensions it wants added into the SignRequest.  I think that's correct because it means that the extension data that goes into the cert is provided by the calling code, not the entity submitting the CSR.  I'll find a place to add a comment.

bifurcation:
@jsha I don't really think that makes sense.  The data in CSRWhitelist is not all in extensions.  And in any case, seems like a separate PR.

bifurcation:
@kisom ping

kisom:
LGTM.

PR #431, opened 2015-11-20
Omit empty qualifiers in Certificate Policies.

jsha:
The previous code would emit a zero-length `SEQUENCE` for any CertificatePolicy,
which caused Java's keytool to complain (and is incorrect DER encoding).

Also, remove two unused fields.

Relevant support thread: https://community.letsencrypt.org/t/no-data-available-in-policyqualifiers/3975/11


jmhodges:
From a boulder dev perspective, I'm cool with this. I'm sure others here are more comfortable with this code.

kisom:
LGTM.

PR #435, opened 2015-11-27
minor command line improvements

mitchcapper:
specifying multiple hostsnames on the command line (set the cn to the first and the hosts to all for dns), specifying a custom crl url on the command line

### DIFF
--- a/signer/signer.go
+++ b/signer/signer.go
@@ -53,13 +53,14 @@ type Extension struct {
 // Extensions requested in the CSR are ignored, except for those processed by
 // ParseCertificateRequest (mainly subjectAltName).
 type SignRequest struct {
-	Hosts      []string    `json:"hosts"`
-	Request    string      `json:"certificate_request"`
-	Subject    *Subject    `json:"subject,omitempty"`
-	Profile    string      `json:"profile"`
-	Label      string      `json:"label"`
-	Serial     *big.Int    `json:"serial,omitempty"`
-	Extensions []Extension `json:"extensions,omitempty"`
+	Hosts       []string    `json:"hosts"`
+	Request     string      `json:"certificate_request"`
+	Subject     *Subject    `json:"subject,omitempty"`
+	Profile     string      `json:"profile"`
+	CRLOverride string      `json:"crl_override"`
+	Label       string      `json:"label"`
+	Serial      *big.Int    `json:"serial,omitempty"`
+	Extensions  []Extension `json:"extensions,omitempty"`
 }
 
 // appendIf appends to a if s is not an empty string.
--- a/cli/config.go
+++ b/cli/config.go
@@ -55,6 +55,7 @@ type Config struct {
 	MaxHosts          int
 	Responses         string
 	Path              string
+	CRL               string
 	Usage             string
 	PGPPrivate        string
 	Serial            string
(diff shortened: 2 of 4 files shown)


signer/signer.go
```diff
@@ -52,6 +52,7 @@ type SignRequest struct {
 	Request    string      `json:"certificate_request"`
 	Subject    *Subject    `json:"subject,omitempty"`
 	Profile    string      `json:"profile"`
+	CRLOverride	 string	`json:"crl_override"`
```
grittygrease: Run gofmt on this file, please.

mitchcapper:
all set ran gofmt as requested.

kisom:
Setting the CN to the first hostname isn't something we'd like in CFSSL.

The [current CAB Forum baseline requirements](https://cabforum.org/wp-content/uploads/Baseline_Requirements_V1_3_1.pdf) (if you're particularly curious, section 7.1.4.2.2a) note that the common name field is deprecated; while its use is not prohibited, it's discouraged. On a similar note, [RFC 6125](https://tools.ietf.org/html/rfc6125#section-6.4.4) prohibits (MUST NOT) the use of the field for matching a domain name when a DNS SAN is present. As the SAN is required for a BR-compliant certificate, there's no reason to use the commonName field, and we'd prefer to follow the CAB Forum's guidelines in not using the field. For certificates that aren't meant to secure traffic for a browser (such as a client authentication certificate), the signer will still copy over the common name from the CSR.

mitchcapper:
OK, but currently it just sets it to the entire comma separated domain list.   Honestly I am just trying to largely fix the domain bug here.  Documentation officially says comma separate it to do multiple domains ("[-hostname comma,separated,hostnames]") but that doesn't currently work.   Are you asking I modify it to blank the CN field on the certificates?

kisom:
We'd rather just not touch the CN there.

grittygrease:
@mitchcapper exactly. Please fix the comma separated list bug and leave the CN parsing as it currently is. Thanks for your PR!

mitchcapper:
Sorry for the delay, we implemented this awhile ago and I was trying to figure out why we had that line, at first I figured the CN was picking up the entire list.   It turns out we had a desire to set the CN on the command line outside of the json file and there is no command to do so.  

If this would be useful I am happy to submit a PR, with the desired implementation.  I could see either a separate flag specifying the CN or just a flag to use the hostname for the CN (and then I would assume the same splitting would be desired).  If this is not desired then we will just use a private patch for our version.

grittygrease:
I think a CN flag to fix your issue would work. Happy to review a PR for just that.

mitchcapper:
I rebased this onto HEAD and had previously removed the adjusting of the CN field.  I also submitted PR #591 which adds the ability to override the CN.  Anything else needed for this PR?

kisom:
The tests are failing because `gofmt` needed to fix a file. I did some digging around, and it's line 101 of local.go --- if you could run `go fmt` (or better, `goimports -l -w`) on signer/local/local.go, the tests should pass (I don't have push access to your branch, so I can't make the change myself).

mitchcapper:
the space bit me. go fmt run and pushed.

kisom:
LGTM.

PR #442, opened 2015-12-01
api package code coverage

codyli520:
0.3% increase for api/generator
20% increase for api/initca
3.5% increase for api/scan


kisom:
LGTM.

PR #443, opened 2015-12-01
cli package code coverage

codyli520:
5.5 % increase for cli.go
100% increase for cli/version
85.2% increase for cli/scan
60% increase for cli/sign


kisom:
LGTM.

PR #453, opened 2015-12-07
Adding Heartbleed Detection to Scan

anniephan:
Part of https://github.com/cloudflare/cfssl/wiki/Adding-Vulnerability-scaning-to-Scan.

Extends CFSSL Scan to scan a host for vulnerability to Heartbleed using https://github.com/FiloSottile/Heartbleed. Thoughts on using this third-party tool? Should a custom version of this scan be implemented locally instead?

Next steps would be adding more scans to detect TLS POODLE, BERserk, LogJam, and FREAK. I haven't been able to find helpful articles detailing how to detect these vulnerabilities, any links/resources suggestions would be great.

### DIFF
--- a/scan/vulnerability.go
+++ b/scan/vulnerability.go
@@ -1,0 +1,55 @@
+package scan
+
+import (
+	"fmt"
+	"log"
+	"net/url"
+
+	"github.com/FiloSottile/Heartbleed/heartbleed"
+)
+
+var Vulnerability = &Family{
+	Description: "Scans a host to detect TLS vulnerabilities.",
+	Scanners: map[string]*Scanner{
+		"Heartbleed": {
+			"Scan for Heartbleed vulnerability",
+			heartbleedScan,
+		},
+	},
+}
+
+func heartbleedScan(addr, hostname string) (grade Grade, output Output, err error) {
+	tgt := &heartbleed.Target{
+		Service: "https",
+		HostIp:  hostname,
+	}
+
+	u, err := url.Parse(tgt.HostIp)
+	if err == nil && u.Host != "" {
+		tgt.HostIp = u.Host
+		if u.Scheme != "" {
+			tgt.Service = u.Scheme
+		}
+	}
+
+	out, err := heartbleed.Heartbleed(tgt,
+		[]byte("github.com/FiloSottile/Heartbleed"), true)
+	if err == heartbleed.Safe {
+		str := fmt.Sprintf("%v - SAFE", tgt.HostIp)
+		grade, output = Good, str
+		err = nil
+	} else if err != nil {
+		if err.Error() == "Please try again" {
+			str := fmt.Sprintf("%v - TRYAGAIN: %v", tgt.HostIp, err)
+			grade, output = Warning, str
+		} else {
+			str := fmt.Sprintf("%v - ERROR: %v", tgt.HostIp, err)
+			grade, output = Warning, str
+		}
+	} else {
+		log.Printf("%v\n", out)
+		str := fmt.Sprintf("%v - VULNERABLE", tgt.HostIp)
+		grade, output = Bad, str
+	}
+	return
+}
--- a/scan/scan_common.go
+++ b/scan/scan_common.go
@@ -121,11 +121,12 @@ type FamilySet map[string]*Family
 
 // Default contains each scan Family that is defined
 var Default = FamilySet{
-	"Connectivity": Connectivity,
-	"TLSHandshake": TLSHandshake,
-	"TLSSession":   TLSSession,
-	"PKI":          PKI,
-	"Broad":        Broad,
+	"Vulnerability": Vulnerability,
+	"Connectivity":  Connectivity,
+	"TLSHandshake":  TLSHandshake,
+	"TLSSession":    TLSSession,
+	"PKI":           PKI,
+	"Broad":         Broad,
 }
 
 // ScannerResult contains the result for a single scan.


scan/vulnerability.go
```diff
@@ -0,0 +1,56 @@
...
+	},
+}
+
+func heartbleedScan(addr, hostname string) (grade Grade, output Output, err error) {
+	tgt := &heartbleed.Target{
+		Service: "https",
+		HostIp:  hostname,
+	}
+
+	u, err := url.Parse(tgt.HostIp)
```
jacobhaven: It's assumed that `hostname` will never have a scheme.

scan/vulnerability.go
```diff
@@ -0,0 +1,56 @@
...
+		tgt.HostIp = u.Host
+		if u.Scheme != "" {
+			tgt.Service = u.Scheme
+		}
+	}
+
+	out, err := heartbleed.Heartbleed(tgt,
+		[]byte("github.com/FiloSottile/Heartbleed"), true)
+	if err == heartbleed.Safe {
+		str := fmt.Sprintf("%v - SAFE", tgt.HostIp)
```
jacobhaven: There's no need to repeat the `hostname` in the outputs. I don't think I would set any outputs for any of these (the relevant messages are already set in the error).

scan/vulnerability.go
```diff
@@ -0,0 +1,56 @@
...
+	} else if err != nil {
+		if err.Error() == "Please try again" {
+			str := fmt.Sprintf("%v - TRYAGAIN: %v", tgt.HostIp, err)
+			grade, output = Warning, str
+		} else {
+			str := fmt.Sprintf("%v - ERROR: %v", tgt.HostIp, err)
+			grade, output = Warning, str
+		}
+	} else {
+		log.Printf("%v\n", out)
```
jacobhaven: Why not just: `grade, output = Bad, out`

scan/vulnerability.go
anniephan: Ah, ok. I'll be sure to take that out.

scan/vulnerability.go
anniephan: I see. I'll just leave out the outputs then. Thanks!

scan/vulnerability.go
anniephan: Noted, thanks!

scan/vulnerability.go
```diff
@@ -0,0 +1,119 @@
...
+		err = nil
+	} else if err != nil {
+		grade = Warning
+	} else {
+		grade, output = Bad, out
+	}
+	return
+}
+
+func BERserkScan(addr, hostname string) (grade Grade, output Output, err error) {
```
codyli520: looks like Travis failed here, since this function name's first letter is capitalized, add a comment above it or uncapitalized it. Reference: http://www.goinggo.net/2014/03/exportedunexported-identifiers-in-go.html

scan/vulnerability.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if pub.N.BitLen() == 1024 || pub.N.BitLen() == 2048 {
+		grade = Bad
+		err = errors.New("certificate has bad public key length")
+	}
+	return
+}
+
+func LogJamScan(addr, hostname string) (grade Grade, output Output, err error) {
```
codyli520: see above comment

scan/vulnerability.go
anniephan: Ah, yes. I'll get to that, thanks.

scan/vulnerability.go
j-delaney: Do you need some err handling after this?

scan/vulnerability.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	u, err := url.Parse(tgt.HostIp)
+	if err == nil && u.Host != "" {
+		tgt.HostIp = u.Host
+	}
+
+	out, err := heartbleed.Heartbleed(tgt,
+		[]byte("github.com/FiloSottile/Heartbleed"), true)
+	if err == heartbleed.Safe {
```
j-delaney: A switch statement might be a bit more clear here

scan/vulnerability.go
```diff
@@ -0,0 +1,119 @@
...
+		err = errors.New("certificate has bad public key length")
+	}
+	return
+}
+
+func LogJamScan(addr, hostname string) (grade Grade, output Output, err error) {
+	// do a DHE handshake and look at the public exponent used by the server.
+	// If it’s 1024 bits, then it’s vulnerable.
+
+	//DHE handshake?
```
jacobhaven: Go's `crypto/tls` doesn't support DHE TLS handshakes, you'll need to use https://github.com/cloudflare/cf-tls/blob/master/tls/cfsslscan_handshake.go to construct a custom client hello and then read the server hello `ServerDHParams`

scan/vulnerability.go
```diff
@@ -0,0 +1,119 @@
...
+	return
+}
+
+func BERserkScan(addr, hostname string) (grade Grade, output Output, err error) {
+	// Vulnerable if certificate has a 3 as the public exponent
+	// get certificate from host via certinfo
+
+	certData, err := certinfo.ParseCertificateDomain(hostname)
+	if err != nil {
+		grade = Bad
```
jacobhaven: `Bad` is (very intentionally) the zero value of the `Grade` type, so all of these `grade = Bad` statements are noops

anniephan: Ahh I see.

scan/vulnerability.go
```diff
@@ -0,0 +1,119 @@
...
+	} else if err != nil {
+		grade = Warning
+	} else {
+		grade, output = Bad, out
+	}
+	return
+}
+
+func BERserkScan(addr, hostname string) (grade Grade, output Output, err error) {
+	// Vulnerable if certificate has a 3 as the public exponent
```
FiloSottile: That is not a valid indicator of BERserk vulnerability. BERserk is a client side vuln, and the existence of ANY e=3 trusted roots make ALL websites vulnerable. (Modulo pinning, if you will, but I don't think that's ground for saying that e=3 is a server vulnerability.)

anniephan: I see, thanks for pointing that out.

scan/vulnerability.go
anniephan: Hmm, what does `ServerDHParams` refer to? I assume it's the cert returned by `SayHello()`, but I am having some trouble reading the public key length from that cert.

jacobhaven: No, the Diffie Hellman Parameters are in a separate [Server Key Exchange TLS message](https://tools.ietf.org/html/rfc5246#section-7.4.3). You'll need to read and parse that message and detect if `dh_p` is under 1024 bits.

scan/vulnerability.go
```diff
@@ -0,0 +1,223 @@
...
+	config.CurvePreferences = allCurvesIDs()
+	conn = tls.Client(tcpConn, config)
+	serverCipher, _, _, _, _, _ = conn.SayHello()
+	conn.Close()
+
+	// If Server uses a DHE cipher with a public exponent <= 1024 bits, then it's vulnerable
+	if serverCipher != 0 {
+		certData, err := certinfo.ParseCertificateDomain(hostname)
+		if err != nil {
+			grade = Bad
```
jacobhaven: All these `grade = Bad` can be removed and `return grade, output, err` is the same as just `return`

scan/vulnerability.go
```diff
@@ -0,0 +1,223 @@
...
+		return
+	}
+
+	config := defaultTLSConfig(hostname)
+
+	// Append DHE_EXPORT ciphers
+	ciphers := make([]uint16, 0, 2)
+	ciphers = append(ciphers, 0X0011)
+	ciphers = append(ciphers, 0X0014)
+	config.CipherSuites = ciphers
```
jacobhaven: `config.CipherSuites = []uint16{0x0011, 0x0014}`

scan/vulnerability.go
```diff
@@ -0,0 +1,223 @@
...
+
+	config.CurvePreferences = allCurvesIDs()
```
jacobhaven: `CurvePreferences` don't matter as we're not supporting any Elliptic Curve cipher suites.

scan/vulnerability.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+
+		ca, err := helpers.ParseCertificatePEM([]byte(certData.RawPEM))
+		if err != nil {
+			grade = Bad
+			return grade, output, err
+		}
+
+		// TODO: this public key isn't from the DHE handshake
+		// Not sure how to get that key...
```
jacobhaven: You'll need to add custom code to [cf-tls here](https://github.com/cloudflare/cf-tls/blob/master/tls/cfsslscan_handshake.go#L50) There is code already there for reading server key exchange messages for ECDH(E) cipher suites. You should be able to use the same `c.exchangeKeys()` to get the raw DH(E) [ServerKeyExchange](https://tools.ietf.org/html/rfc5246#section-7.4.3) message and parse out `dh_p` from the `ServerDHParams`.
It's not that trivial, as Go doesn't natively have support for parsing these DH(E) messages, but it should be doable.

scan/vulnerability.go
```diff
@@ -0,0 +1,223 @@
...
+
+// Helper to get all export grade ciphers
+func allExportCiphersIDs() []uint16 {
+	exportCount := 0
+	for _, name := range tls.CipherSuites {
+		if strings.Contains(name.Name, "_EXPORT_") {
+			exportCount = exportCount + 1
+		}
+	}
+	ciphers := make([]uint16, 0, exportCount)
```
jacobhaven: You can just define `var ciphers []uint16` and append onto it without counting out the capacity first.

scan/vulnerability.go
```diff
@@ -0,0 +1,223 @@
...
+		if strings.Contains(name.Name, "_EXPORT_") {
+			ciphers = append(ciphers, cipherID)
+		}
+	}
+	return ciphers
+}
+
+func freakScan(addr, hostname string) (grade Grade, output Output, err error) {
+	// See if the server will negotiate a EXPORT grade cipher
+	tcpConn, err := net.Dial(Network, hostname+":443")
```
jacobhaven: Why not just use the `sayHello` helper from [tls_handshake.go](https://github.com/cloudflare/cfssl/blob/master/scan/tls_handshake.go#L64) (you're still in the `scan` package so you can access it)

scan/vulnerability.go
anniephan: Ah, yes, added. Thanks.

jacobhaven:
I think it might be best to separate this out into a couple smaller PRs. Mainly I'd pull out the heartbleed check to simplify what's being added a bit and maybe Logjam as that's going to require quite a bit more work to finish.

lziest:
orphaned PR?

PR #455, opened 2015-12-09
Zi/new certdb

lziest:
2nd take of the certdb pacakge. This time we have one set of SQL queries that works for both postgres and sqlite.


lziest:
Ready to be reviewed.

certdb/certdb.go
```diff
@@ -0,0 +1,227 @@
...
+
+// InsertCertificate puts a CertificateRecord into db.
+func InsertCertificate(db *sql.DB, cr *CertificateRecord) error {
+	res, err := db.Exec(
+		insertSQL,
+		cr.Serial,
+		cr.CALabel,
+		cr.Status,
+		cr.Reason,
+		cr.Expiry.UTC(),
```
andrewbuss: I don't have a good idea of how timezones should be handled in software generally. Are callers of functions that return `CertificateRecord`s responsible for any conversion if we cast to UTC before inserting into the database?

certdb/certdb.go
```diff
@@ -0,0 +1,227 @@
...
+func RevokeCertificate(db *sql.DB, serial string, reasonCode int) error {
+	result, err := db.Exec(updateRevokeSQL, reasonCode, serial)
+
+	if err != nil {
+		return wrapCertStoreError(err)
+	}
+
+	numRowsAffected, _ := result.RowsAffected()
+
+	if numRowsAffected != 1 {
```
andrewbuss: Should we return a different error if a certificate with a given serial isn't in the database? That's a much less severe and more useful error - we'd want to return that all the way up to the client.

certdb/certdb.go
lziest: I think it is fine. https://golang.org/src/time/time.go?s=2427:2459#L39. Time is kept as the seconds and nseconds since the epoch. Since in sqlite we use 'timestamp' (without timezone), we convert them into UTC, and make them consistent to be compared with CURRENT_TIMESTAMP (a UTC timestamp). Time conversion can be easily done by changing Location, but the underlying representation has not changed.

certdb/certdb.go
lziest: sure, I will change accordingly.

certdb/testdb/testdb.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"database/sql"
+
+	_ "github.com/lib/pq"           // register postgresql driver
+	_ "github.com/mattn/go-sqlite3" // register sqlite3 driver
+)
+
+const (
+	pgTruncateTables = `
```
andrewbuss: What does this do?

doc/errorcode.txt
```diff
@@ -12,6 +12,8 @@
             1213: TooManyIntermediates
             1214: IncompatibleUsage
         1220: UnknownAuthority
+    1300: BadRequest
+    1400: MissingSerial
```
andrewbuss: It doesn't look like this error is used

andrewbuss: Never mind. Didn't see it in this diff but it's already in use.

certdb/testdb/testdb.go
lziest: It makes sure all tables in the database are empty. It is a snippet I shamelessly copy/pasted from my colleague. So it could be a overkill, and be simplified further.

certdb/certdb.go
```diff
@@ -0,0 +1,227 @@
...
+		return wrapCertStoreError(fmt.Errorf("%d rows are affected, should be 1 row", numRowsAffected))
+	}
+
+	return nil
+}
+
+// InsertOCSP puts a new OCSPRecord into the db.
+func InsertOCSP(db *sql.DB, rr *OCSPRecord) error {
```
andrewbuss: As I understood the intention of the ocsp_responses table, we want to regularly refresh the whole table with a new set of responses and a later expiration date. If a serial number is absent from the table (a new certificate since the last refresh), the UpdateOCSP function will fail, and if it's present; this one will. Would it be easier to expose an UpsertOCSP function which either inserts or updates a row?

certdb/REAMDE.md
```diff
@@ -0,0 +1,26 @@
+This directory stores [goose](https://bitbucket.org/liamstask/goose/) db migration scripts for various DB backends.
```
andrewbuss: Nit: this file should probably be README.md

certdb/certdb.go
```diff
@@ -0,0 +1,227 @@
...
+
+// CertificateRecord encodes a certificate and its metadata
+// that will be recorded in a database.
+type CertificateRecord struct {
+	Serial    string    `sql:"serial"`
+	CALabel   string    `sql:"ca_label"`
+	Status    string    `sql:"status"`
+	Reason    int       `sql:"reason"`
+	Expiry    time.Time `sql:"expiry"`
+	RevokedAt time.Time `sql:"revoked_at"`
```
andrewbuss: What value should RevokedAt take when a certificate has not been revoked?

certdb/certdb.go
```diff
@@ -0,0 +1,227 @@
...
+			return nil, wrapCertStoreError(err)
+		}
+		crs = append(crs, cr)
+	}
+
+	return crs, nil
+}
+
+// RevokeCertificate updates a certificate with a given serial number and marks it revoked.
+func RevokeCertificate(db *sql.DB, serial string, reasonCode int) error {
```
andrewbuss: Should this function be responsible for updating the OCSP response for this certificate, or is that left to the caller? If we do it here, we can combine these into a SQL transaction, although I'm not sure whether atomicity is of huge importance here.

certdb/certdb.go
lziest: default zero value, and it is idiomatic to test the zero value by https://golang.org/pkg/time/#Time.IsZero

certdb/certdb.go
lziest: I am not sure either. The original intent is to abstract sql query with certdb package. I still consider OCSP can be separated from certificate revocation since OCSP is only one of several revocation mechanisms. It left work for OCSP responder/generator which need to look up the certificate table and OCSP table to act correctly.

certdb/certdb.go
lziest: I like the idea of 'update if exists otherwise insert'. Now it is time to monkey-typing with the SQL alphabets to see if any viable query comes out.

lziest: for now, I would provide "upsert" without using the real sql upsert. 1. a quick peek into sqlite and postgresql docs gets me two different syntaxs again (not surprisingly).
I assume this is accurate and up to date. https://wiki.postgresql.org/wiki/UPSERT  
2. we don't need a strict simultaneous consistency between OCSP and certificate status. It's ok that a OCSP response still shows 'good' while the corresponding certificate is being revoked seconds ago, as long as the OCSP response catch up to be eventually consistent (within hours to days). So write race condition between CA and ocspgen on OCSP table is OK since we don't have write race condition on Certificate table and ocspgen will run periodically to let OCSP catch up .

lziest:
@andrewbuss I want to get this and yours PR merged by EOW

andrewbuss:
This PR looks good to me, though I'd feel better not being the only reviewer. 

My PR is based on this branch (f613512 at present) so it should merge cleanly into master once this is merged. It's ready for review.

lziest:
@grittygrease @kisom can you review and merge?

PR #458, opened 2015-12-10
Add revocation and OCSP generation support using zi's certdb

andrewbuss:
Rebased https://github.com/cloudflare/cfssl/pull/422 onto Zi's branch in #455 and squashed commits. 

At present the ocspgen command refreshes the entire OCSP response table
and then queries it again to generate an output file. We might in the
future spin this out into two commands: one to refresh the OCSP table
and another to generate an output file from it. What's the intended workflow?


andrewbuss:
I'm curious why Travis isn't running checks on this. Does that only run for merges into master?

lziest:
Yes, I think you are right, we want them separate. One refreshes OCSP table, the other generates a file from it.

cli/ocspgen/ocspgen.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		resp, err := s.Sign(req)
+		if err != nil {
+			log.Critical("Unable to sign OCSP response: ", err)
+			return err
+		}
+
+		// Update the OCSP record to expire in 1 day
+		certdb.UpdateOCSP(nil, cert.SerialNumber.String(), string(resp), time.Now().AddDate(0, 0, 1))
```
lziest: useful to have OCSP validity a cli parameter so we can tune from '1 day' to '2 day' to '7 days'. Could use existing parameter 'timeout' or you can invent a new one.

lziest: or use 'interval'

andrewbuss:
Rebased and split up ocspgen

lziest:
please rebase to master now. Thank you.

andrewbuss:
Superseded by #464 since this PR wasn't based on master

PR #463, opened 2015-12-15
Update scan cli to use a worker pool

### LINKED ISSUE #462: Use a worker pool for scan
In order to not exhaust the host, a fixed worker pool should be used for scanning a long list of hosts.

jacobhaven:
This resolves #462 and allows scan to be used for large sets of hosts (e.g. the entire Alexa top 1 million)


lziest:
lgtm

PR #466, opened 2015-12-16
Avoid 90 seconds of unnecessarily waiting on a TCP timeout in tests

### LINKED ISSUE #465: API client tests take 90 seconds on Mac and Go 1.4 builds
TestAuthSign, TestSign, and TestDefaultAuthSign try to connect to 1.0.0.1 with Go 1.4, and waits through 3 30-second timeouts, but in Go 1.5, Sign fails with "invalid domain name". I don't know why this occurs on Mac with Go 1.5 though. These timeouts account for ~30% of the Jenkins test duration for these platforms.

andrewbuss:
Closes #465 

`.X` is chosen to avoid a DNS lookup.

### DIFF
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -57,30 +57,30 @@ func TestInvalidPort(t *testing.T) {
 }
 
 func TestAuthSign(t *testing.T) {
-	s := NewServer("1.1")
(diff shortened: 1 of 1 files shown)


jacobhaven:
LGTM

PR #468, opened 2015-12-17
Just building seems to do nothing

davidk-zenefits:


andrewbuss:
This should probably be `go build ./cmd/...`

lziest:
Please provide some context of this change since we don't consider "go build" is broken.

andrewbuss:
`go build cmd/...` builds non-CFSSL packages like `cmd/cover` and `cmd/vet` rather than the intended `github.com/cloudflare/cfssl/cmd/...`

jacobhaven:
I've [updated the BUILDING.md instructions](https://github.com/cloudflare/cfssl/commit/0ec15debaac95e372ef331c76d80e8aa4887ae8e) to use `go install` for the path `./cmd/...`, fixing the problem of matching the standard library `cmd` package.

PR #471, opened 2015-12-21
Restrict curve scan to ECDHE ciphersuites.

grittygrease:
Add all non-PSK ECDHE ciphersuites as the cipher list for the EC curve scan.


lziest:
lgtm

kisom:
LGTM.

PR #472, opened 2015-12-21
refine postgres migration scritpts

lziest:
per @benburkert's suggestion, refining postgres sql migration.


kisom:
LGTM.

certdb/pg/migrations/001_CreateCertificates.sql
```diff
@@ -3,16 +3,25 @@
 
 CREATE TABLE certificates (
   id                serial,
```
benburkert: `serial primary key`

certdb/pg/migrations/001_CreateCertificates.sql
```diff
@@ -3,16 +3,25 @@
...
+  pem               bytea NOT NULL,
+  PRIMARY KEY(id, serial)
 );
 
+CREATE TABLE ocsp_responses (
+  id                serial PRIMARY KEY,
+  serial            bytea NOT NULL,
+  body              bytea NOT NULL,
+  expiry            timestamptz,
+  FOREIGN KEY(serial) REFERENCES certificates(serial)
```
benburkert: it's probably better to do the foreign key on `certificate.id`:
`certificate_id integer REFERENCES certificates (id)`

certdb/pg/migrations/001_CreateCertificates.sql
lziest: I made (id, serial) primary key, would that works as well?

certdb/pg/migrations/001_CreateCertificates.sql
lziest: hmm, what if we want to flush all ocsp responses manually and run a OCSP refresh? sounds like this foreign key on id would make it impossible?

lziest:
need to address #503, close this.

PR #474, opened 2015-12-22
Support dbconfigs with multirootca

drcapulet:


multiroot/config/testdata/roots_bad_db.config
```diff
@@ -0,0 +1,10 @@
+[ primary ]
```
lziest: file name should be roots_bad_db.conf

drcapulet: Nice catch, updated.

lziest:
LGTM

PR #475, opened 2015-12-22
Use the maxmimum size for serial numbers.

kisom:
This change uses a byte slice with the maximum RFC 5280 length as
specified in RFC 5280 4.1.2.2.

### DIFF
--- a/signer/local/local.go
+++ b/signer/local/local.go
@@ -14,8 +14,8 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
-	"math"
 	"math/big"
 	"net"
 	"net/mail"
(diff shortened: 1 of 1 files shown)


signer/local/local.go
```diff
@@ -257,11 +257,20 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
-		serialNumber, err := rand.Int(rand.Reader, new(big.Int).SetInt64(math.MaxInt64))
+		// RFC 5280 4.1.2.2:
+		// Certificate users MUST be able to handle serialNumber
+		// values up to 20 octets.  Conforming CAs MUST NOT use
+		// serialNumber values longer than 20 octets.
+		//
+		// If CFSSL is providing the serial numbers, it makes
+		// sense to use the max supported size.
+		serialNumber := make([]byte, 20)
+		_, err = io.ReadFull(rand.Reader, serialNumber)
```
lziest: one thought, shall we prevent serialNumber from being negative?

kisom: Agreed, added a change.

lziest:
LGTM

PR #477, opened 2015-12-22
Remove unnecessary initialization of CertificateRecord fields when signing a cert

andrewbuss:
Reason, Status, and RevokedAt should be initialized to "", 0, and Jan 1 1 (https://golang.org/pkg/time/#Time.IsZero) respectively.

### DIFF
--- a/signer/local/local.go
+++ b/signer/local/local.go
@@ -19,8 +19,6 @@ import (
 	"math/big"
 	"net"
 	"net/mail"
-	"time"
-
 	"github.com/cloudflare/cfssl/certdb"
 	"github.com/cloudflare/cfssl/config"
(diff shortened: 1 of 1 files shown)


signer/local/local.go
```diff
@@ -336,10 +336,7 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		var certRecord = &certdb.CertificateRecord{
 			Serial:    certTBS.SerialNumber.String(),
 			CALabel:   req.Label,
-			Status:    "",
```
lziest: 2nd thought, I think making it "good" by default is the desired behavior.

lziest:
LGTM

PR #480, opened 2015-12-23
Import goose so Godeps picks up the dependencies (correctly)

drcapulet:
This fixes the fact that goose is in Godeps but not actually "included" anywhere. The alternate PR I can submit is removing goose from `Godeps/Godeps.json`

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -6,6 +6,10 @@
 	],
 	"Deps": [
 		{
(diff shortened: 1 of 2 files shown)


lziest:
OK I now recalled that we used to have used goose in our test code and then we remove it. So yes I guess the correct move is to remove goose from Godeps.json

kisom:
I agree --- removing it from Godeps.json is probably the better choice here.

PR #486, opened 2015-12-31
add a new function to ocsp, enabling a certdb backed ocsp responder

lziest:
NewSourceFromDB instantiate a InMemorySource object from certdb.


lziest:
obsoleted

PR #487, opened 2016-01-19
Add support for password encoded PEM private key

mclem:
The CA private key I use to sign CSR and generate certificates must be encrypted because we have a requirement not to store it in clear text on the system.
This PR enables the use of such private keys by looking for a password in the "CFSSL_CA_PK_PASSWORD" environement variable.


lziest:
can you rebase?

lziest:
LGTM

PR #490, opened 2016-01-21
TLS Revocation Scan with OSCP Stapling check

jacobhaven:


scan/tls_revocation.go
```diff
@@ -0,0 +1,50 @@
...
+		RootCAs:    RootCAs,
+	})
+	if err != nil {
+		return
+	}
+	conn.Close()
+
+	connState := conn.ConnectionState()
+
+	if len(connState.OCSPResponse) == 0 {
```
jacobhaven: change this to `conn.OCSPResponse()` for pre Go 1.5 support. https://github.com/golang/go/blob/master/src/crypto/tls/conn.go#L1068

Issue #491, opened 2016-01-21
Break dependency between signer/local and certdb


jsha:
In https://github.com/letsencrypt/boulder/pull/1397, @hlandau is pulling in the latest CFSSL into Boulder so we can do OCSP extensions (to provide SCTs via OCSP). Unfortunately, that pulls in dependencies on sqlstruct, pq, and go-sqlite3, which we'd rather not pull in.

I believe this is because signer/local now depends on certdb, which depends on those other packages. Is it possible to break that dependency? I think it would probably be ideal if Signer was responsible only for signing, and the code that calls Signer was responsible for storing the result. That way the calling code can implement various storage strategies, retries, and so on.

Thanks,
Jacob

lziest:
we will make signer/local dependents on a interface only certdb package, which only depends on "database/sql", how would that sound?

jsha:
That works for me, thanks!

PR #492, opened 2016-01-22
Zi/certdb abstraction

lziest:
Needs some comments on this approach. It is definitely making signer more entangled with certdb. But the benefits is now signer depends on the sql and certdb's DBAccessor interface, which doesn't bring in db drivers and sql formatters.


lziest:
@grittygrease @jsha this is a attempt to fix #491.

hlandau:
What's the status of this? It seems ready.

lziest:
I still need to work on the db abstraction a little bit more because I want to get rid of setting both db and db accessor in a signer to have a certificate store in a db when issued.

lziest:
now ready for review.

jsha:
I haven't fully reviewed this, but I checked it out locally, and can confirm it fixes #491 (breaks dependencies).

certdb/certdb.go
```diff
@@ -29,252 +25,16 @@ type OCSPRecord struct {
...
-		return InsertOCSP(db, &OCSPRecord{Serial: serial, Body: body, Expiry: expiry})
-	}
-
-	if numRowsAffected != 1 {
-		return wrapCertStoreError(fmt.Errorf("%d rows are affected, should be 1 row", numRowsAffected))
-	}
-	return
+// DBAccessor abstracts the CRUD of certdb objects from a DB.
+type DBAccessor interface {
+	SetDB(db *sql.DB)
```
jsha: I think it would probably be cleaner to not have `SetDB` be a property of the `DBAccessor` interface, but instead have types that satisfy `DBAccessor` take care of setting the DB as part of their constructor.

lziest: Without `SetDB`, `DBAccessor` interface looks more general than a DB accessor, it could be just Accessor.

lziest:
OK I am done with renaming.

certdb/sql/database_accessor.go
```diff
@@ -0,0 +1,327 @@
...
+		return wrapSQLError(fmt.Errorf("%d rows are affected, should be 1 row", numRowsAffected))
+	}
+	return err
+}
+
+// UpsertOCSP update a ocsp response record with a given serial number,
+// or insert the record if it doesn't yet exist in the db
+// Implementation note:
+// We didn't implement 'upsert' with SQL statement and we lost race condition
+// prevention provided by underlying DMBS.
```
jsha: *DBMS

jsha:
This looks great to me, thanks!

Issue #416, opened 2015-11-06
Proposal: delete PKCS11 config code from universal signer and OCSP signer


jsha:
Right now, the universal signer and the OCSP signer allow configuring PKCS11 keys by passing a config variable. This means there's a dependency between those two packages and the pkcs11key package. The pkcs11key package in turn depends on the external pkcs11 package, which requires cgo and libtool-ltdl. The pkcs11key code is gated by the build tag `nopkcs11`, but is included by default, by request from my team working on Boulder (thanks for accommodating us!).

However, the cgo and libtool-ltdl dependencies are causing difficulty for other contributors. And currently, Boulder doesn't actually use the config options in the universal signer or the OCSP signer. Instead, we create an instance of pkcs11key.Key ourselves, and pass that into `NewSigner`. I'm not aware of any other project that uses those config options. I propose deleting them in order to narrow the scope of dependencies. This would make pkcs11key the only package in cfssl that depends on cgo. To make commands like `go test ./...` work for people without cgo, we could then move pkcs11key into a separate repo, either under Cloudflare's GitHub account or Let's Encrypt's (I would be happy with either; Let's Encrypt may make the most sense since we are doing the most active development on that package).

Note: If you'd like to preserve the config options that create PKCS11 keys, there's another way to do it: we can follow the example of `database/sql` and provide a `Register()` method that pkcs11key can call to register its availability. Users of cfssl could then choose whether or not to build in pkcs11key support, depending on their needs. However, I think the extra complexity is not warranted if no-one is currently using the config features.

If this proposal sounds good to people I'm happy to do the work.

kisom:
Sounds good here, and it sounds like it would solve a lot of our problems. Thanks!.

PR #493, opened 2016-01-22
Support using cfssl in appengine applications

adieu:
Appengine doesn't support pkcs11 and we can't use build tags when build and deploy gae apps. This patch simply add build tag to disable pkcs11 in appengine environment.

### DIFF
--- a/crypto/pkcs11key/key.go
+++ b/crypto/pkcs11key/key.go
@@ -1,4 +1,4 @@
-// +build !nopkcs11
+// +build !nopkcs11,!appengine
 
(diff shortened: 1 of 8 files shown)


grittygrease:
I'm fine with this. FYI: there is a plan by @jsha to update the source to not require pkcs#11 by default here: https://github.com/cloudflare/cfssl/issues/416, which may also help with your issue.

lziest:
we have turn off pkcs11 by default. I think this can be closed.

lziest:
Close this one and feel free to reopen it doesn't solve your needs.

PR #497, opened 2016-01-28
Rely on sqlx for driver independent bind vars

drcapulet:
I'd like to add MySQL support and this uses `sqlx` which is a wrapper over `database/sql` that helps with handling driver-independent variables.

### DIFF
--- a/certdb/certdb.go
+++ b/certdb/certdb.go
@@ -7,23 +7,23 @@ import (
 // CertificateRecord encodes a certificate and its metadata
 // that will be recorded in a database.
 type CertificateRecord struct {
-	Serial    string    `sql:"serial_number"`
-	AKI       string    `sql:"authority_key_identifier"`
-	CALabel   string    `sql:"ca_label"`
-	Status    string    `sql:"status"`
-	Reason    int       `sql:"reason"`
-	Expiry    time.Time `sql:"expiry"`
-	RevokedAt time.Time `sql:"revoked_at"`
-	PEM       string    `sql:"pem"`
+	Serial    string    `db:"serial_number"`
+	AKI       string    `db:"authority_key_identifier"`
+	CALabel   string    `db:"ca_label"`
+	Status    string    `db:"status"`
+	Reason    int       `db:"reason"`
+	Expiry    time.Time `db:"expiry"`
+	RevokedAt time.Time `db:"revoked_at"`
+	PEM       string    `db:"pem"`
 }
 
 // OCSPRecord encodes a OCSP response body and its metadata
 // that will be recorded in a database.
 type OCSPRecord struct {
-	Serial string    `sql:"serial_number"`
-	AKI    string    `sql:"authority_key_identifier"`
-	Body   string    `sql:"body"`
-	Expiry time.Time `sql:"expiry"`
+	Serial string    `db:"serial_number"`
+	AKI    string    `db:"authority_key_identifier"`
+	Body   string    `db:"body"`
+	Expiry time.Time `db:"expiry"`
 }
 
 // Accessor abstracts the CRUD of certdb objects from a DB.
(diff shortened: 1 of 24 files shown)


lziest:
Hi @drcapulet Thanks for you PR. We are also in the process of refactoring certdb so it is easier for you to change database/sql to other package so you don't have to change every package that uses certdb. Actually your change will be probably much smaller as we can switch in any db implementation and certdb users don't need to care.

drcapulet:
@lziest looks like you merged #492 so I'll rebase off of that.

drcapulet:
@lziest updated, PTAL.

certdb/certdb.go
```diff
@@ -7,32 +7,32 @@ import (
...
+	Body   string    `db:"body"`
+	Expiry time.Time `db:"expiry"`
 }
 
 // Accessor abstracts the CRUD of certdb objects from a DB.
 type Accessor interface {
 	InsertCertificate(cr *CertificateRecord) error
 	GetCertificate(serial string) (*CertificateRecord, error)
-	GetUnexpiredCertificates() ([]*CertificateRecord, error)
+	GetUnexpiredCertificates() ([]CertificateRecord, error)
```
lziest: I want to keep some consistencies between get one record vs get a slice of record, but it could be hard. Is it possible we don't change this function signature?

certdb/certdb.go
```diff
@@ -7,32 +7,32 @@ import (
...
+	Status    string    `db:"status"`
+	Reason    int       `db:"reason"`
+	Expiry    time.Time `db:"expiry"`
+	RevokedAt time.Time `db:"revoked_at"`
+	PEM       string    `db:"pem"`
 }
 
 // OCSPRecord encodes a OCSP response body and its metadata
 // that will be recorded in a database.
 type OCSPRecord struct {
```
lziest: just fyi, I am updating this struct as well, planning to add AKI (authority key identifier) so (aki, serial) uniquely identifier a OCSP request.

certdb/certdb.go
drcapulet: Unfortunately using `sqlx.Select` reads into a pointer - what if we had all the gets return values instead of pointers?

lziest: yeah, that's another solution. We can make them all non-pointers, even for insert. The only problem is `GetCertificate(serial string) (CertificateRecord, error)`  would return some zero-valued structure when it gets nothing. And the caller has to do some check to realize that get returns nothing. My idea is to make it `GetCertificate(serial string) ([]CertificateRecord, error)`

lziest: let me do the interface change in #504, so there would be less conflicts.

kisom:
Looks like this PR needs to be rebased.

drcapulet:
@kisom / @lziest Rebased, PTAL

lziest:
LGTM after glancing over, will take a deep look Monday. Thank you!

certdb/sql/database_accessor.go
```diff
@@ -220,56 +204,37 @@ func (d *Accessor) InsertOCSP(rr certdb.OCSPRecord) error {
...
-		rrs = append(rrs, rr)
-	}
-	return rrs, nil
+
+	return ors, nil
 }
 
 // GetUnexpiredOCSPs retrieves all unexpired certdb.OCSPRecord from db.
-func (d *Accessor) GetUnexpiredOCSPs() (rrs []certdb.OCSPRecord, err error) {
+func (d *Accessor) GetUnexpiredOCSPs() (ors []certdb.OCSPRecord, err error) {
```
lziest: thanks. my initial intention was name it R(esponse)R(ecord)s, but never correct it after I rename the table.

lziest:
LGTM.

lziest:
Need to verify docker build before merging.

lziest:
@drcapulet can you help me to rebase your PR again? I can merge then.

drcapulet:
@lziest should be good now!

PR #500, opened 2016-01-29
Finish support for e-mail address SANs

cabrel:
This pull request is to ensure that if e-mail addresses are provided in the CSR that they make it to the final certificate. Prior to this, e-mail addresses were being treated as dns names and weren't appearing in the SAN field as rfc822Name entities.


lziest:
LGTM, @kisom want to take a look too?

Issue #503, opened 2016-02-07
certdb's ocsp db migration scheme needs to take issuer certificate key identifier into account


lziest:
Currently the ocsp table is indexed by certificate's serial number only. It should also contain certificate's issuer key identifier (AKI) to avoid serial number collision.

PR #504, opened 2016-02-08
Zi/refine certdb

lziest:
address #503 and added a little postgres schema refinement.


lziest:
@jacobhaven please review. :)

certdb/pg/migrations/001_CreateCertificates.sql
```diff
@@ -3,16 +3,26 @@
 
 CREATE TABLE certificates (
   id                serial,
-  serial            bytea NOT NULL PRIMARY KEY,
+  serial            bytea NOT NULL UNIQUE,
```
jacobhaven: Should rename this to `serial_num` to not conflict with Postgres type

certdb/pg/migrations/001_CreateCertificates.sql
```diff
@@ -3,16 +3,26 @@
 
 CREATE TABLE certificates (
   id                serial,
```
jacobhaven: I'm not sure need these `id` fields.

certdb/pg/migrations/001_CreateCertificates.sql
```diff
@@ -3,16 +3,26 @@
...
   revoked_at        timestamptz,
-  pem               bytea NOT NULL
+  pem               bytea NOT NULL,
+  PRIMARY KEY(id, serial)
 );
 
+CREATE TABLE ocsp_responses (
+  id                       serial PRIMARY KEY,
+  serial                   bytea NOT NULL,
+  authority_key_identifier bytea NOT NULL,
```
jacobhaven: I would move `authority_key_identifier` to the `certificates` table and have

```SQL
PRIMARY KEY(serial, authority_key_identifier)
```

certdb/sqlite/migrations/002_CreateOCSPResponses.sql
```diff
@@ -2,9 +2,10 @@
 -- SQL in section 'Up' is executed when this migration is applied
```
jacobhaven: This is not the same as the Postgres table. Should probably have them match :P

lziest:
@jacobhaven @drcapulet is it OK I merge this by EOW?

drcapulet:
Overall LGTM, though I do find it odd we return slices in `GetCertificate` & `GetOCSP` while the method names sound singular and in the DB we're uniquely identifying both certificates + OCSP responses by (serial, aki).

lziest:
I understand. We can change that later. I want to settle the DB schema now, interface can definitely evolve. I want to do "insertCertificate(cert *x509.Certificate)" as well so we don't need to remember hex encoding the hash value. And if you are really annoyed, let's do it in your PR. :)

certdb/pg/migrations/001_CreateCertificates.sql
```diff
@@ -2,17 +2,27 @@
...
 CREATE TABLE certificates (
-  id                serial,
-  serial            bytea NOT NULL PRIMARY KEY,
-  ca_label          bytea NOT NULL,
-  status            bytea NOT NULL,
-  reason            int,
-  expiry            timestamptz,
-  revoked_at        timestamptz,
-  pem               bytea NOT NULL
+  serial_number            bytea NOT NULL UNIQUE,
```
drcapulet: Do we need the `UNIQUE` constraint here since the `PRIMARY KEY` covers it?

lziest: good point, we can remove it

lziest:
I am happy to merge now

kisom:
LGTM.

Issue #426, opened 2015-11-16
Add configurability to log facility


jsha:
Right now cfssl's log facility allows configuring levels, but as far as I can tell doesn't allow changing where the logs go. In Boulder we're relying on syslog and would like to be able to send logs there. I'd like to add a way for cfssl users to add their own logger.

kisom:
I like this approach, and also would like both the ability to log to a file and to separate INFO and below from WARNING up. It would also be nice to have a logging package that reported the component that was doing the logging (e.g. signer/local).

K4lium:
@jsha Does this meet your needs?

PR #505, opened 2016-02-08
Jwagnerk/use syslog

K4lium:
Address issue #426

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -24,6 +24,7 @@ Use "cfssl [command] -help" to find out more about a command.
 */
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -31,7 +32,6 @@ import (
 	"io/ioutil"
 	"os"
 
-	"encoding/base64"
 	"github.com/cloudflare/cfssl/config"
 )
 
--- a/log/log.go
+++ b/log/log.go
@@ -6,8 +6,10 @@
 package log
 
 import (
+	"flag"
 	"fmt"
 	golog "log"
+	"log/syslog"
 	"os"
 )
 
@@ -30,18 +32,72 @@ var levelPrefix = [...]string{
 	LevelFatal:    "[FATAL] ",
 }
 
+var levelPriority = [...]syslog.Priority{
+	LevelDebug:    syslog.LOG_DEBUG,
+	LevelInfo:     syslog.LOG_INFO,
+	LevelWarning:  syslog.LOG_WARNING,
+	LevelError:    syslog.LOG_ERR,
+	LevelCritical: syslog.LOG_CRIT,
+	LevelFatal:    syslog.LOG_EMERG,
+}
+
 // Level stores the current logging level.
 var Level = LevelDebug
 
+var (
+	//UseSyslog controls if syslog is used instead of stdout/err.
+	UseSyslog bool
+	//SyslogNetwork controls the network protocol syslog uses.
+	SyslogNetwork string
+	//SyslogRaddr is the address of a remote syslog instance.
+	SyslogRaddr string
+	//SyslogTag is the tag to pass to syslog.
+	SyslogTag string
+)
+
+func init() {
+	flag.IntVar(&Level, "loglevel", LevelDebug, "Log level")
+	flag.BoolVar(&UseSyslog, "syslog", false, "Whether or not to use syslog logging")
+	flag.StringVar(&SyslogNetwork, "syslog-network", "udp", "Syslog network to use.")
+	flag.StringVar(&SyslogRaddr, "syslog-remote", "", "Syslog server to use. Defaults to local on empty string.")
+	flag.StringVar(&SyslogTag, "syslog-tag", "udp", "Syslog tag to use.")
+}
+
 func outputf(l int, format string, v []interface{}) {
 	if l >= Level {
-		golog.Printf(fmt.Sprint(levelPrefix[l], format), v...)
+		msg := fmt.Sprintf(fmt.Sprint(levelPrefix[l], format), v...)
+		if UseSyslog {
+			logger, err := syslog.Dial(SyslogNetwork, SyslogRaddr, levelPriority[l], SyslogTag)
+			defer logger.Close()
+			if err != nil {
+				golog.Fatal(err)
+			}
+			golog.SetOutput(logger)
+		} else if l > 2 {
+			golog.SetOutput(os.Stderr)
+		} else {
+			golog.SetOutput(os.Stdout)
+		}
+		golog.Print(msg)
 	}
 }
 
 func output(l int, v []interface{}) {
 	if l >= Level {
-		golog.Print(levelPrefix[l], fmt.Sprint(v...))
+		msg := fmt.Sprint(levelPrefix[l], fmt.Sprint(v...))
+		if UseSyslog {
+			logger, err := syslog.Dial(SyslogNetwork, SyslogRaddr, levelPriority[l], SyslogTag)
+			defer logger.Close()
+			if err != nil {
+				golog.Fatal(err)
+			}
+			golog.SetOutput(logger)
+		} else if l > 2 {
+			golog.SetOutput(os.Stderr)
+		} else {
+			golog.SetOutput(os.Stdout)
+		}
+		golog.Print(msg)
 	}
 }
 
--- a/log/log.go
+++ b/log/log.go
@@ -8,97 +8,80 @@ package log
 import (
 	"flag"
 	"fmt"
-	golog "log"
+	"log"
 	"log/syslog"
 	"os"
 )
 
 // The following constants represent logging levels in increasing levels of seriousness.
 const (
+	// LevelDebug is the log level for Debug statements.
 	LevelDebug = iota
+	// LevelInfo is the log level for Info statements.
 	LevelInfo
+	// LevelWarning is the log level for Warning statements.
 	LevelWarning
+	// LevelError is the log level for Error statements.
 	LevelError
+	// LevelCritical is the log level for Critical statements.
 	LevelCritical
+	// LevelFatal is the log level for Fatal statements.
 	LevelFatal
 )
 
 var levelPrefix = [...]string{
-	LevelDebug:    "[DEBUG] ",
-	LevelInfo:     "[INFO] ",
-	LevelWarning:  "[WARNING] ",
-	LevelError:    "[ERROR] ",
-	LevelCritical: "[CRITICAL] ",
-	LevelFatal:    "[FATAL] ",
+	LevelDebug:    "DEBUG",
+	LevelInfo:     "INFO",
+	LevelWarning:  "WARNING",
+	LevelError:    "ERROR",
+	LevelCritical: "CRITICAL",
+	LevelFatal:    "FATAL",
 }
-
-var levelPriority = [...]syslog.Priority{
-	LevelDebug:    syslog.LOG_DEBUG,
-	LevelInfo:     syslog.LOG_INFO,
-	LevelWarning:  syslog.LOG_WARNING,
-	LevelError:    syslog.LOG_ERR,
-	LevelCritical: syslog.LOG_CRIT,
-	LevelFatal:    syslog.LOG_EMERG,
-}
-
-// Level stores the current logging level.
-var Level = LevelDebug
 
 var (
-	//UseSyslog controls if syslog is used instead of stdout/err.
-	UseSyslog bool
-	//SyslogNetwork controls the network protocol syslog uses.
-	SyslogNetwork string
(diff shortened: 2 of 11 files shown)


transport/transport_test.go
```diff
@@ -51,7 +51,6 @@ func removeIfPresent(path string) error {
 }
 
 func TestMain(m *testing.M) {
-	flag.IntVar(&log.Level, "loglevel", log.LevelInfo, "log level (0 = DEBUG, 4 = ERROR)")
 	flag.Parse()
```
jacobhaven: We should remove `flag` from this file entirely. (It's not longer parsing anything)

log/log.go
```diff
@@ -30,18 +32,83 @@ var levelPrefix = [...]string{
...
+	LevelDebug:    syslog.LOG_DEBUG,
+	LevelInfo:     syslog.LOG_INFO,
+	LevelWarning:  syslog.LOG_WARNING,
+	LevelError:    syslog.LOG_ERR,
+	LevelCritical: syslog.LOG_CRIT,
+	LevelFatal:    syslog.LOG_EMERG,
+}
+
 // Level stores the current logging level.
 var Level = LevelDebug
```
jacobhaven: This should default to `LevelInfo`

log/log.go
```diff
@@ -30,18 +32,83 @@ var levelPrefix = [...]string{
...
+	//SyslogNetwork controls the network protocol syslog uses.
+	SyslogNetwork string
+	//SyslogRaddr is the address of a remote syslog instance.
+	SyslogRaddr string
+	//SyslogTag is the tag to pass to syslog.
+	SyslogTag string
+)
+
+func init() {
+	flag.IntVar(&Level, "loglevel", LevelDebug, "Log level (0 = DEBUG, 5 = FATAL)")
```
jacobhaven: Similarly, default to `LevelInfo`

log/log.go
K4lium: Noted. Debug was prior behavior.

log/log.go
```diff
@@ -30,18 +32,83 @@ var levelPrefix = [...]string{
...
+	LevelError:    syslog.LOG_ERR,
+	LevelCritical: syslog.LOG_CRIT,
+	LevelFatal:    syslog.LOG_EMERG,
+}
+
 // Level stores the current logging level.
 var Level = LevelDebug
 
+var (
+	//SysLoggers maps each logging level to a *log.Logger that will be used for it.
```
jacobhaven: These comments are missing a starting space. Should be: `// SysLoggers[...]`

grittygrease:
LGTM

log/log.go
```diff
@@ -30,18 +32,83 @@ var levelPrefix = [...]string{
...
+			golog.Print(msg)
+		}
 	}
 }
 
 func output(l int, v []interface{}) {
 	if l >= Level {
-		golog.Print(levelPrefix[l], fmt.Sprint(v...))
+		msg := fmt.Sprint(levelPrefix[l], fmt.Sprint(v...))
+		if UseSyslog {
```
jacobhaven: This could just be
```Go
if UseSyslog {
	if logger := getSysLogger(l); logger != nil {
		logger.Print(msg)
		return
	}
}
golog.Print(msg)	
```

And it should probably be factored out into a helper function that both `output` and `outputf` call.

log/log.go
```diff
@@ -6,8 +6,10 @@
 package log
 
 import (
+	"flag"
 	"fmt"
 	golog "log"
```
jacobhaven: There's no real reason to rename this `golog`, but eh

log/log.go
```diff
@@ -30,18 +32,83 @@ var levelPrefix = [...]string{
...
+	//SyslogTag is the tag to pass to syslog.
+	SyslogTag string
+)
+
+func init() {
+	flag.IntVar(&Level, "loglevel", LevelDebug, "Log level (0 = DEBUG, 5 = FATAL)")
+	flag.BoolVar(&UseSyslog, "syslog", false, "Whether or not to use syslog logging")
+	flag.StringVar(&SyslogNetwork, "syslog-network", "udp", "Syslog network to use.")
+	flag.StringVar(&SyslogRaddr, "syslog-remote", "", "Syslog server to use. Defaults to local on empty string.")
+	flag.StringVar(&SyslogTag, "syslog-tag", "udp", "Syslog tag to use.")
```
jacobhaven: This should default to `""` (the syslog package will use the name of the executable)

log/log.go
```diff
@@ -30,18 +32,83 @@ var levelPrefix = [...]string{
...
+	//SyslogRaddr is the address of a remote syslog instance.
+	SyslogRaddr string
+	//SyslogTag is the tag to pass to syslog.
+	SyslogTag string
+)
+
+func init() {
+	flag.IntVar(&Level, "loglevel", LevelDebug, "Log level (0 = DEBUG, 5 = FATAL)")
+	flag.BoolVar(&UseSyslog, "syslog", false, "Whether or not to use syslog logging")
+	flag.StringVar(&SyslogNetwork, "syslog-network", "udp", "Syslog network to use.")
```
jacobhaven: This should default to "" (the syslog package will fail to dial if you leave `syslog-remote=""`)

log/log.go
```diff
@@ -30,18 +32,75 @@ var levelPrefix = [...]string{
...
+	flag.StringVar(&SyslogNetwork, "syslog-network", "", "Syslog network to use. Will fail to dial if left blank.")
+	flag.StringVar(&SyslogRaddr, "syslog-remote", "", "Syslog server to use. Defaults to local on empty string.")
+	flag.StringVar(&SyslogTag, "syslog-tag", "", "Syslog tag to use. Will use the name of the executable if left blank.")
+}
+
+func getSysLogger(l int) *golog.Logger {
+	if SysLoggers[l] != nil {
+		return SysLoggers[l]
+	}
+
```
jacobhaven: Maybe you can add some logic here:

```Go
net := SyslogNetwork
if SyslogRaddr!="" && net=="" {
	net = "udp"
}
w, err := syslog.Dial(net, SyslogRaddr, levelPriority[l], SyslogTag)
```
So that you still get the "default to UDP" behavior, but it still works with local syslog if both fields are left blank.

log/log.go
```diff
@@ -30,18 +32,80 @@ var levelPrefix = [...]string{
...
+	net := SyslogNetwork
+	// This block keeps makes this use local syslog if nothing else is configured.
+	if SyslogRaddr != "" && net == "" {
+		net = "udp"
+	}
+	w, err := syslog.Dial(net, SyslogRaddr, levelPriority[l], SyslogTag)
+	if err == nil {
```
jsha: This error check inverts the normal order of error checking in Go, which is confusing. Would be better to say `if err != nil {...`

log/log.go
```diff
@@ -30,18 +32,80 @@ var levelPrefix = [...]string{
...
+}
+
+func logHelper(l int, msg string) {
+	if UseSyslog {
+		if logger := getSysLogger(l); logger != nil {
+			logger.Print(msg)
+			return
+		}
+	}
+	golog.Print(msg)
```
jsha: Ideally we'd like to be able to use syslogging as an alternative to stdout logging. That is, if we have syslog enabled, we'd like stdout logging to be disabled. The reason: We want to additionally collect stdout/stderr using operating system facilities, but treat it as critical level, because panics get printed there. If stdout gets a duplicate of what goes to syslog, that means we can't collect it without getting duplicates.

log/log.go
```diff
@@ -30,18 +32,80 @@ var levelPrefix = [...]string{
...
+	if SyslogRaddr != "" && net == "" {
+		net = "udp"
+	}
+	w, err := syslog.Dial(net, SyslogRaddr, levelPriority[l], SyslogTag)
+	if err != nil {
+		golog.Println("Unable to dial syslog:", err)
+		return nil
+	}
+
+	SysLoggers[l] = golog.New(w, levelPrefix[l], 0)
```
jsha: This is a neat way to handle the syslog integration! Does it mean that syslog entries get doubled timestamps?

K4lium: The go log package and the syslog host will apply their timestamps independently, I think.

log/log.go
```diff
@@ -30,18 +32,80 @@ var levelPrefix = [...]string{
...
+	LevelFatal:    syslog.LOG_EMERG,
+}
+
 // Level stores the current logging level.
-var Level = LevelDebug
+var Level = LevelInfo
+
+var (
+	// SysLoggers maps each logging level to a *log.Logger that will be used for it.
+	SysLoggers = make([]*golog.Logger, LevelFatal+1)
```
jsha: These variables should probably not be exported, since you don't expect people to be able to set or read them externally, right?

For Boulder's needs, the nicest thing would be to call some exported method, e.g. `log.UseSyslog(writer)` to enable syslogging. I think that would imply, instead of the multiple syslogger instances created below, using a single syslog instance with a big switch:

```
switch level {
  case syslog.LOG_DEBUG:
    writer.Debug(message)
  ...
```

K4lium: Point. Correcting now.

A limitation of the log/syslog package is that each syslog writer can only have one priority. I'd prefer to have one syslogger for each priority, so that priority information is available to the syslog server.

jsha: Well, each of the `w.Debug(...)`, `w.Warning(...)`, `w.Err(...)` methods sets the severity corresponding to its name, disregarding what the syslogger was constructed with. We use this pattern in Boulder, and the information gets passed along to syslog nicely.

This would take the place of the `golog.New(...)` invocation, which would have the side effect of not double-logging timestamps.

log/log.go
jacobhaven: @jsha the All the Loggers created by this package have their flags set to `0`, so they don't output timestamps (but Syslog might)

jsha: Ah, cool. The Go syslog package doesn't add a timestamp to the message body, but timestamp is part of the metadata received by the log server, and most rsyslog configurations will write out the timestamp. So this looks fine.

log/log.go
```diff
@@ -6,45 +6,133 @@
...
+		if UseSyslog {
+			w, err := getSysLogger()
+			if err != nil {
+				log.Printf("Unable to dial syslog: %v for msg: %s\n", err, msg)
+				return
+			}
+
+			switch l {
+			case LevelDebug:
+				if err := w.Debug(msg); err != nil {
```
jsha: You can simplify this a bunch by moving the error check outside the switch statement. Like so:

```
switch l {
case LevelDebug:
  err = w.Debug(msg)
...
}
if err != nil {
  ...
}
```

Note the `=` instead of `:=`, which assigns to the err variable declared in the encompassing block, ensuring that it is visible outside of the select block.

log/log.go
```diff
@@ -6,45 +6,133 @@
...
+	var err error
+	sysLogger, err = syslog.Dial(net, SyslogRaddr, 0, SyslogTag)
+	return sysLogger, err
 }
 
-func output(l int, v []interface{}) {
+func print(l int, msg string) {
 	if l >= Level {
-		golog.Print(levelPrefix[l], fmt.Sprint(v...))
+		msg = fmt.Sprintf("[%s] %s", levelPrefix[l], msg)
```
jsha: You should only add the levelPrefix when using the Go log library, since it's part of the metadata passed to syslog.

log/log.go
```diff
@@ -6,45 +6,133 @@
...
 
-// Level stores the current logging level.
-var Level = LevelDebug
+var (
+	// Level stores the current logging level.
+	Level = LevelInfo
+	// sysLogger is a syslog Writer connection.
+	sysLogger *syslog.Writer
+	// UseSyslog controls if syslog is used instead of stdout/err.
+	UseSyslog bool
```
jsha: Do these variables (UseSyslog, SyslogNetwork, etc) need to be exported?

log/log.go
jacobhaven: Very good point.

log/log.go
jacobhaven: Yes, so that a user of the package can set them manually without relying on flag parsing. I would consider removing them entirely and instead only exporting `SetSyslog` (or a global `SysLogger *syslog.Writer` var), expecting the user to user `syslog.Dial` or `syslog.New` directly when enabling syslog.

log/log.go
jacobhaven: Fair enough. The Priority is not always clear on a syslog stream, but it's up to that configuration to determine how it is displayed.

cli/cli.go
```diff
@@ -32,6 +32,7 @@ import (
 	"os"
```
lziest: we shoud erase this blank line

lziest:
can we squash all the commits?

lziest:
lgtm

PR #507, opened 2016-02-10
add SerialNumber to signer.Subject

jmhodges:
This allows boulder and other clients to create certificates without CommonNames but with non-empty Subjects that might cause problems for clients.

Unlike in a certificate where its a number, RFC 5280 specify's a Subject's SerialNumber as a printable string. Hence, the Go `pkix.Name`'s type for its `SerialNumber`.

### DIFF
--- a/signer/local/local.go
+++ b/signer/local/local.go
@@ -80,10 +80,10 @@ func NewSignerFromFile(caFile, caKeyFile string, policy *config.Signing) (*Signe
 	if err != nil {
 		return nil, err
 	}
-	
+
 	strPassword := os.Getenv("CFSSL_CA_PK_PASSWORD")
 	password := []byte(strPassword)
-	if (strPassword == "") {
+	if strPassword == "" {
 		password = nil
 	}
 
@@ -163,7 +163,9 @@ func PopulateSubjectFromCSR(s *signer.Subject, req pkix.Name) pkix.Name {
 	replaceSliceIfEmpty(&name.Locality, &req.Locality)
 	replaceSliceIfEmpty(&name.Organization, &req.Organization)
 	replaceSliceIfEmpty(&name.OrganizationalUnit, &req.OrganizationalUnit)
-
+	if name.SerialNumber == "" {
+		name.SerialNumber = req.SerialNumber
+	}
 	return name
 }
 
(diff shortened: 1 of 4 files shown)


signer/local/local.go
```diff
@@ -163,7 +163,9 @@ func PopulateSubjectFromCSR(s *signer.Subject, req pkix.Name) pkix.Name {
 	replaceSliceIfEmpty(&name.Locality, &req.Locality)
 	replaceSliceIfEmpty(&name.Organization, &req.Organization)
 	replaceSliceIfEmpty(&name.OrganizationalUnit, &req.OrganizationalUnit)
-
+	if name.SerialNumber == "" {
```
jsha: I think this was supposed to be `!=`?

jmhodges: Nope, you've got it flipped. This is for if name doesn't have the SerialNumber set already. (e.g. boulder or another client didn't pass one in to the Subject)

jsha: Ah, got it.

jsha:
LGTM

jmhodges:
Added a test for `Subject.Name` just for completeness.

jmhodges:
Ping!

lziest:
Hi, I need to understand the purpose of `SerialNumber` in subject, is it to prevent the subject object being empty? feels like a little bit duct-taping if that is the case.

jmhodges:
Yep, that's the purpose. How else would we go about preventing the certificate's Subject being empty in the case where the CommonName and other properties are not given? We don't get to control the other parts of the certificate's Subject.

jmhodges:
To be clear what's going on:

Boulder is currently able to create certificates that violate the Baseline Requirements. Since boulder forces CNs to be non-empty by choosing a SAN to go in there if the CN is blank, it will sometimes be forced to pick a SAN that is larger than the 64 bytes allowed for a CN. That is, because of RFC 5280, CNs can only be 64 bytes, while SANs can be much larger.

We've gotten lots of feedback from people who would know that empty Subjects would be very not-good for a lot of clients, so we can't just stop putting a SAN into the CN. We're trying to get this patch in so we can both stop forcing SANs into the CN without also breaking clients that expect a certificate to have a Subject.

kisom:
Thanks for clarifying, @jmhodges. LGTM.

PR #508, opened 2016-02-10
Jwagnerk/pkcs 11 default off

K4lium:
As per issue #416, PKCS#11 support will now default to off. It can be included with the build tag of pkcs11.


jsha:
This doesn't actually do what I proposed in #416. You still need to break the dependency from the OCSP signer and the universal signer to the pkcs11key package. What I proposed there was to simply remove the config options that allow enable PKCS#11, since people who need PKCS#11 can manually construct a `pkcs11key.Key` and pass it to the signer and OCSP signer constructors.

Having broken that dependency, you no longer need the pkcs11key build tag at all.

If this change lands as-is, it will break Boulder.

PR #511, opened 2016-02-12
Add support for mutual TLS in server

dhiltgen:
This change introduces two new flags for the serve command to enable mutual
TLS.  The '-mutual-tls-ca' flag specifies a PEM file containing CA certificates
which the client must present a cert signed by.  The '-mutual-tls-cn' flag
allows further restrictions by requiring the client cert contain the exact
match for the specified CN string.

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -20,6 +20,8 @@ type Config struct {
 	CAKeyFile         string
 	TLSCertFile       string
 	TLSKeyFile        string
+	MutualTLSCAFile   string
+	MutualTLSCNRegex  string
 	KeyFile           string
 	IntermediatesFile string
 	CABundleFile      string
@@ -75,6 +77,8 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.StringVar(&c.CAKeyFile, "ca-key", "", "CA private key")
 	f.StringVar(&c.TLSCertFile, "tls-cert", "", "Other endpoint CA to set up TLS protocol")
 	f.StringVar(&c.TLSKeyFile, "tls-key", "", "Other endpoint CA private key")
+	f.StringVar(&c.MutualTLSCAFile, "mutual-tls-ca", "", "Mutual TLS - require clients be signed by this CA ")
+	f.StringVar(&c.MutualTLSCNRegex, "mutual-tls-cn", "", "Mutual TLS - regex for whitelist of allowed client CNs")
 	f.StringVar(&c.KeyFile, "key", "", "private key for the certificate")
 	f.StringVar(&c.IntermediatesFile, "intermediates", "", "intermediate certs")
 	f.StringVar(&c.CABundleFile, "ca-bundle", "", "path to root certificate store")
--- a/cli/serve/serve.go
+++ b/cli/serve/serve.go
@@ -2,12 +2,15 @@
 package serve
 
 import (
+	"crypto/tls"
 	"database/sql"
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -28,6 +31,7 @@ import (
 	"github.com/cloudflare/cfssl/cli"
 	ocspsign "github.com/cloudflare/cfssl/cli/ocspsign"
 	"github.com/cloudflare/cfssl/cli/sign"
+	"github.com/cloudflare/cfssl/helpers"
 	"github.com/cloudflare/cfssl/log"
 	"github.com/cloudflare/cfssl/ocsp"
 	"github.com/cloudflare/cfssl/signer"
@@ -42,14 +46,14 @@ Usage of serve:
                     [-ca-key key] [-int-bundle bundle] [-int-dir dir] [-port port] \
                     [-metadata file] [-remote remote_host] [-config config] \
                     [-responder cert] [-responder-key key] [-tls-cert cert] [-tls-key key] \
-                    [-db-config db-config]
+                    [-mutual-tls-ca ca] [-mutual-tls-cn regex] [-db-config db-config]
 
 Flags:
 `
 
 // Flags used by 'cfssl serve'
 var serverFlags = []string{"address", "port", "ca", "ca-key", "ca-bundle", "int-bundle", "int-dir", "metadata",
-	"remote", "config", "responder", "responder-key", "tls-key", "tls-cert", "db-config"}
+	"remote", "config", "responder", "responder-key", "tls-key", "tls-cert", "mutual-tls-ca", "mutual-tls-cn", "db-config"}
 
 var (
 	conf       cli.Config
@@ -244,9 +248,42 @@ func serverMain(args []string, c cli.Config) error {
 		log.Info("Now listening on ", addr)
 		return http.ListenAndServe(addr, nil)
 	}
+	if conf.MutualTLSCAFile != "" {
+		clientPool, err := helpers.LoadPEMCertPool(conf.MutualTLSCAFile)
+		if err != nil {
+			return fmt.Errorf("failed to load mutual TLS CA file: %s", err)
+		}
 
+		server := http.Server{
+			Addr: addr,
+			TLSConfig: &tls.Config{
+				ClientAuth: tls.RequireAndVerifyClientCert,
+				ClientCAs:  clientPool,
+			},
+		}
+
+		if conf.MutualTLSCNRegex != "" {
+			log.Debugf(`Requiring CN matches regex "%s" for client connections`, conf.MutualTLSCNRegex)
+			re, err := regexp.Compile(conf.MutualTLSCNRegex)
+			if err != nil {
+				return fmt.Errorf("malformed CN regex: %s", err)
+			}
+			server.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
+					if re.MatchString(r.TLS.PeerCertificates[0].Subject.CommonName) {
+						http.DefaultServeMux.ServeHTTP(w, r)
+						return
+					}
+					log.Warningf(`Rejected client cert CN "%s" does not match regex %s`,
+						r.TLS.PeerCertificates[0].Subject.CommonName, conf.MutualTLSCNRegex)
+				}
+				http.Error(w, "Invalid CN", http.StatusForbidden)
+			})
+		}
+		log.Info("Now listening with mutual TLS on https://", addr)
+		return server.ListenAndServeTLS(conf.TLSCertFile, conf.TLSKeyFile)
+	}
 	log.Info("Now listening on https://", addr)
-
 	return http.ListenAndServeTLS(addr, conf.TLSCertFile, conf.TLSKeyFile, nil)
 
 }


cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
 		log.Info("Now listening on ", addr)
 		return http.ListenAndServe(addr, nil)
 	}
+	if conf.MutualTLSCAFile != "" {
+		mutualTLSCAs, err := ioutil.ReadFile(conf.MutualTLSCAFile)
```
jacobhaven: You should use [helpers.LoadPEMCertPool](https://godoc.org/github.com/cloudflare/cfssl/helpers#LoadPEMCertPool), 

```Go
clientPool, err := helpers.LoadPEMCertPool(conf.MutualTLSCAFile)
if err != nil {
	return fmt.Errorf("failed to load mutual TLS CA file: %s", err)
}
```

cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
...
+		if err != nil {
+			return fmt.Errorf("Failed to load mutual TLS CA file: %s", err)
```
jacobhaven: Go error strings always begin lowercase (i.e. `"failed to..."`)

cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
...
+		server := http.Server{
+			Addr: addr,
+			TLSConfig: &tls.Config{
+				Certificates: []tls.Certificate{keypair},
+				ClientAuth:   tls.RequireAndVerifyClientCert,
+				ClientCAs:    clientPool,
+			},
+		}
 
+		if conf.MutualTLSCN != "" {
```
jacobhaven: This is a strange bit a logic. Common Names values are not usually constrained and thus couldn't be used like this for authentication. Do you have a specific use case in mind?

cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
...
+			log.Infof("Requiring CN=%s for client connections", conf.MutualTLSCN)
```
jacobhaven: I would use `log.Debugf` here.

cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
...
+				if r != nil && r.TLS != nil {
+					connState := r.TLS
+					for _, peerCert := range connState.PeerCertificates {
+						if peerCert.Subject.CommonName == conf.MutualTLSCN {
+							http.DefaultServeMux.ServeHTTP(w, r)
+							return
+						}
+					}
+				}
+				log.Debug("Rejected client cert for unexpected CN")
```
jacobhaven: And `log.Warning` here.

cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
...
+				ClientCAs:    clientPool,
+			},
+		}
 
+		if conf.MutualTLSCN != "" {
+			log.Infof("Requiring CN=%s for client connections", conf.MutualTLSCN)
+			server.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.TLS != nil {
+					connState := r.TLS
+					for _, peerCert := range connState.PeerCertificates {
```
jacobhaven: I would just check the leaf's common name, not the entire chain.

```Go
if len(r.TLS.PeerCertificates)>0 && r.TLS.PeerCertificates[0].Subject.CommonName == conf.MutualTLSCN {
	[...]
}
```

cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
...
+		}
+		cert, err := ioutil.ReadFile(conf.TLSCertFile)
+		if err != nil {
+			return fmt.Errorf("Failed to load cert file: %s", err)
+		}
+		key, err := ioutil.ReadFile(conf.TLSKeyFile)
+		if err != nil {
+			return fmt.Errorf("Failed to load cert file: %s", err)
+		}
+		keypair, err := tls.X509KeyPair(cert, key)
```
jacobhaven: You don't need to save the `cert`, `key`, or `keypair` here (or manually set `Certificates` in the `TLSConfig`. http.ListenAndServeTLS will [do that for you](https://golang.org/src/net/http/server.go?s=60146:60214#L2026)! Also, should the need arise there is a `tls.LoadX509KeyPair` helper that will read the files for you :)

cli/config.go
```diff
@@ -74,6 +76,8 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.StringVar(&c.CAKeyFile, "ca-key", "", "CA private key")
 	f.StringVar(&c.TLSCertFile, "tls-cert", "", "Other endpoint CA to set up TLS protocol")
 	f.StringVar(&c.TLSKeyFile, "tls-key", "", "Other endpoint CA private key")
+	f.StringVar(&c.MutualTLSCAFile, "mutual-tls-ca", "", "Mutual TLS - require clients be signed by this CA ")
+	f.StringVar(&c.MutualTLSCN, "mutual-tls-cn", "", "Mutual TLS - require clients cert have the given CN")
```
jacobhaven: You need a good justification of this CN-checking functionality. Generally, CNs on client certs are meaningless and not controlled (and thus worthless for authentication).

dhiltgen:
Thanks for the style suggestions.  I'll make those updates.

On the CN topic, what do you think about switching the implementation to support a regex pattern to match anything within the DN?  For my use, I could still match CN, but it would provide a more general model for other usage.

jacobhaven:
@dhiltgen I think generally, client certs are just verified based on the CA they're signed by. Then access to that specific client cert CA is controlled. It doesn't make sense to add more specific checks unless there's an actual use case for them.

dhiltgen:
Sure, let me try to capture our use-case.

CFSSL is **almost** what we need for a CA-as-a-service within the system I work on.  We utilize mutual TLS throughout the system, and specific trusted component(s) are responsible for allowing entities to join.  That trusted component, once it has verified permissions, generates the CSR and sends it off to CFSSL to sign.  I need to have CFSSL on the open network so that all the trusted components in a distributed environment can talk to it, but the problem I'm faced with is I need a way to constrain CFSSL so it only allows these trusted component(s) to make signing requests.  Since the CFSSL API isn't "open" and other services front-end all signing requests, we can easily control the DN, so that makes it a simple and effective way to secure the system.

You could make an argument that we could use a different root cert for the trusted entities, but I feel like that just adds more complexity to the system.  That said, my PR doesn't preclude people from doing that if that's how they want to operate (you can easily omit the DN filtering and it works fine on the CA filtering alone.)

jacobhaven:
Alright, so you have this use case: Multiple differently-privileged TLS clients with certs issued by an API that can control the CNs for you (i.e. multiple cfssl profiles with different `name_whitelist` set)

I still think it's probably simpler/cleaner to instantiate multiple CAs (still just different cfssl profiles) and use a different one for each of your differently-privileged clients.

dhiltgen:
At least in our system, using multiple CAs complicates the mutual trust relationship between entities.

There are quite a few other systems that follow this pattern, so I'm not sure why CFSSL can't support it too.  I'm not proposing this be mandatory, but an optional capability above and beyond simple CA matching.  Some prior-art examples:

http://www.postgresql.org/docs/9.2/static/auth-username-maps.html - User mapping based on regex across DN

http://www.dest-unreach.org/socat/doc/socat.html#OPTION_OPENSSL_COMMONNAME

http://www.cisco.com/c/en/us/support/docs/security/vpn-3000-series-concentrators/26324-vpn3k-cert-match.html -- Using DN matching for VPN access

https://publib.boulder.ibm.com/tividd/td/ITAME/SC32-1359-00/en_US/HTML/am51_webseal_guide88.htm - Tivoli product with optional DN matching

https://docs.oracle.com/cd/E19476-01/821-0506/using-subject-dn-user-attr-cert-mapper.html - Sun's OpenDS mapping DN to user accounts

I could dig up more references, but I think that gives a flavor for it.

cli/serve/serve.go
```diff
@@ -244,9 +248,55 @@ func serverMain(args []string, c cli.Config) error {
...
+			},
+		}
 
+		if conf.MutualTLSCN != "" {
+			log.Infof("Requiring CN=%s for client connections", conf.MutualTLSCN)
+			server.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.TLS != nil {
+					connState := r.TLS
+					for _, peerCert := range connState.PeerCertificates {
+						if peerCert.Subject.CommonName == conf.MutualTLSCN {
```
lziest: can we make it a regex filter? That would potentially make the whitelist more general.

jacobhaven:
Alright, I'll take your word that you'll use this responsibly :)

Just fix up the other comments and this should be fine.

dhiltgen:
Thanks!  I'll refactor this to use a regex on DN and make the other style updates and update the PR.

dhiltgen:
I've made the suggested style updates, and switched to a regex based model.  I wasn't able to find an existing routine to get a clean string from the DN, so I built a relatively simple one.  Open to suggestions if you know of a better way to convert the DN to a string for regex matching.

cli/serve/serve.go
```diff
@@ -203,6 +208,44 @@ func registerHandlers() {
 	log.Info("Handler set up complete.")
 }
 
+// Extract a string representation of the DN suitable for regex matching
+func subjectToString(peerCert *x509.Certificate) string {
```
jacobhaven: I would restrict matching to just the `CN`, not the full `Subject`

cli/serve/serve.go
```diff
@@ -244,9 +287,41 @@ func serverMain(args []string, c cli.Config) error {
...
+			Addr: addr,
+			TLSConfig: &tls.Config{
+				ClientAuth: tls.RequireAndVerifyClientCert,
+				ClientCAs:  clientPool,
+			},
+		}
 
+		if conf.MutualTLSDNRegex != "" {
+			log.Debugf("Requiring DN matches regex %s for client connections", conf.MutualTLSDNRegex)
+			re := regexp.MustCompile(conf.MutualTLSDNRegex)
```
jacobhaven: This should use `regexp.Compile` and return the `err` if it's not `nil`

cli/serve/serve.go
```diff
@@ -244,9 +287,41 @@ func serverMain(args []string, c cli.Config) error {
...
+			server.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.TLS != nil {
+					if len(r.TLS.PeerCertificates) > 0 {
+						dn := subjectToString(r.TLS.PeerCertificates[0])
+						if re.FindStringIndex(dn) != nil {
```
jacobhaven: This can simply be `if re.MatchString(r.TLS.PeerCertificates[0].Subject.CN)`

cli/serve/serve.go
```diff
@@ -244,9 +287,41 @@ func serverMain(args []string, c cli.Config) error {
...
+				ClientCAs:  clientPool,
+			},
+		}
 
+		if conf.MutualTLSDNRegex != "" {
+			log.Debugf("Requiring DN matches regex %s for client connections", conf.MutualTLSDNRegex)
+			re := regexp.MustCompile(conf.MutualTLSDNRegex)
+			server.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.TLS != nil {
+					if len(r.TLS.PeerCertificates) > 0 {
```
jacobhaven: This can all be one check: `if r != nil && r.TLS != nil && len(r.TLS.PeerCertificates) > 0`

dhiltgen:
Switched to CN only regex matching, and style updates.

jacobhaven:
LGTM

dhiltgen:
Thanks!

PR #512, opened 2016-02-12
Reduce PKCS#11 dependency

K4lium:
This is a second cut at issue #416


lziest:
LGTM

K4lium:
@jsha Thoughts?

ocsp/config/config.go
```diff
@@ -15,5 +21,5 @@ type Config struct {
 	ResponderCertFile string
 	KeyFile           string
 	Interval          time.Duration
-	PKCS11            pkcs11key.Config
+	PKCS11            PKCS11Config
```
jsha: We can delete this.

ocsp/config/config.go
```diff
@@ -2,10 +2,16 @@
...
 
-import (
-	"github.com/cloudflare/cfssl/crypto/pkcs11key"
-	"time"
-)
+import "time"
+
+// PKCS11Config contains configuration information required to use a PKCS
+// #11 key.
+type PKCS11Config struct {
```
jsha: Delete this.

crypto/pkcs11key/key_test.go
```diff
@@ -1,4 +1,4 @@
-// +build !nopkcs11
+// +build pkcs11
```
jsha: This shouldn't need a build tag.

crypto/pkcs11key/key.go
```diff
@@ -1,4 +1,4 @@
-// +build !nopkcs11
+// +build pkcs11
```
jsha: Shouldn't need a build tag. In general, the end state of this change should be that there are no more build tags, and `pkcs11key` is a standalone package that has no dependencies into or out of the other packages.

jsha:
Is this ready for review? I see you added a new commit, but there's still an unaddressed issue, that we should just do away with the pkcs11 / nopkcs11 build tags entirely.

crypto/pkcs11key/key_test.go
K4lium: Needs a build tag. key.go and key_test.go make use of miekg/pkcs11 data structures.

K4lium:
The build tags are required if we are to have crypto/pkcs11key in this repository. I could remove both build tags and crypto/pkcs11key, if you like.

crypto/pkcs11key/key_test.go
jsha: Yes, but folks who are building on systems that don't have libtool-ltdl can build just the components that they need, which don't depend on `pkcs11key` anymore. E.g. `go build ./cli`

jsha:
I would also be fine with moving pkcs11key into another repo.

crypto/pkcs11key/key_test.go
jkroll-cf: It's nice to have the nopkcs11 build, as cgo wants to link libtool at runtime, and it's sometimes impractical to package just the parts you need. I have a problem, for example, where I use cfssl libraries in another project and the transitive dependency graph brings in the miekg/pkcs11 package, which then requires the libltdl-dev headers to build and the binary libltdl to run.

Upshot: as long as we're importing the miekg/pkcs11 package, I'm in favor of the build tags.

lziest:
I think the PR itself is a win toward removing pkcs11 dependency by default. We can address the package separation in a new PR. An immature idea is to have a pkcs11signer package that implements cfssl.Signer interface

lziest:
@jsha @K4lium I am merging this. Feel free to open a issue to discuss the pkcs11key package separation.

jsha:
Filed https://github.com/cloudflare/cfssl/issues/416 to follow up with the repo split. I don't think we need a new pkcs11signer package, we just need to split the repo.

PR #516, opened 2016-02-18
Ignore unsupported CSR extensions.

jsha:
Previously, we would hard fail on any CSR extensions that weren't whitelisted.
However, as Hugo pointed out in
https://github.com/letsencrypt/boulder/pull/1224#issuecomment-174065959,
this will break a large number of clients, which include extensions in the CSR
that we don't explicitly process. The new behavior is to copy only the
whitelisted extensions into the certificate, and ignore the rest.


jsha:
Travis is showing these errors:

`can't load package: package github.com/cloudflare/cfssl/vendor/github.com/kardianos/osext: code in directory /home/travis/gopath/src/github.com/cloudflare/cfssl/vendor/github.com/kardianos/osext expects import "github.com/kardianos/osext"`

I think these are not caused by my change, but I could be wrong!

rolandshoemaker:
LGTM.

jcjones:
Thank you @jsha for addressing the unit test change in the previous diff. This change looks good to my eyes.

grittygrease:
@jacobhaven can you look into the failure? It looks to be related to your vendoring work.

signer/local/local.go
```diff
@@ -295,20 +295,18 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+					return nil, cferr.Wrap(cferr.CertificateError, cferr.InvalidRequest, err)
+				}
+
+				safeTemplate.ExtraExtensions = append(safeTemplate.ExtraExtensions, pkix.Extension{
+					Id:       oid,
+					Critical: ext.Critical,
+					Value:    rawValue,
+				})
 			}
-
```
grittygrease: It may be worth adding an informational log here stating that an non-whitelisted extension was skipped over.

jsha: I started to write this but decided that it would generate a lot of non-actionable log spam, because it's really quite common. In Boulder we increment a stat to indicate this case. I'd be happy to incorporate a stat here too if cfssl uses stats (e.g. through expvar)?

grittygrease: Fair enough. We do use metrics, but I'm fine leaving this out for now.

jacobhaven:
@jsha Please rebase off of master (which vendors `github.com/kardianos/osext`)

jacobhaven:
Odd, travis changed the behavior of `- tip` https://travis-ci.org/cloudflare/cfssl/jobs/110241785#L260 as opposed to https://travis-ci.org/cloudflare/cfssl/jobs/110056341#L262

jsha:
Done, thanks!

jsha:
On second thought, I think this is not the right approach, so closing.

Reason: The set of requested extensions is explicitly provided by the caller, rather than pulled from the CSR. Boulder's desired behavior is to ignore CSR extensions that are unrecognized. So the place that makes sense to do that is where we extract extensions from the CSR.

PR #518, opened 2016-02-19
Vendor crypto/tls in the scan package (replace cf-tls)

### LINKED ISSUE #515: Vendor crypto/tls in scan package
Instead of using https://github.com/cloudflare/cf-tls/tls we can simply create a `cfssl/scan/vendor/crypto/tls` directory with our additions.

jacobhaven:
Resolves #515


jacobhaven:
It took vendoring half of the crypto library (for Go 1.5 logic to work), but this should be good. Note: this will break Go 1.4 (or below) builds.

lziest:
LGTM

lziest:
so we vendor the cf-tls, i.e. copying crypto/tls and add our fork to it?

jacobhaven:
We were vendoring `"github.com/cloudflare/cf-tls/tls"`. Instead I modified it to vendor `"crypto/tls"` (and for Go 1.5 we also needed `"crypto"` `"crypto/rsa"`, `"crypto/sha*"`, `"crypto/md5"`) inside the scan package (by adding `scan/vendor/crypto/[...]`)

Issue #519, opened 2016-02-19
Split pkcs11key into a separate repo


jsha:
In https://github.com/cloudflare/cfssl/pull/512, we flipped the default of the `pkcs11` / `nopkcs11` build tags. Unfortunately this blocks Boulder from vendoring the latest version of cfssl, because we build with no build tags.

As discussed in https://github.com/cloudflare/cfssl/issues/416, we can split pkcs11key into its own repo. This is probably the best solution: That way users of cfssl can run `go test ./...` without needing to install libtool-ltdl, and additionally users of the pkcs11key package don't need to define any build tags, so long as they install the requirements to build that package (i.e., libtool-ltdl). This would be my preferred way forward.

jacobhaven:
This sounds great. I'm starting the creating of a `"github.com/cloudflare/pkcs11key"` repo. It'll need a new README (explaining the `libtool-ltdl` dependencies) and the removal of the `pkcs11` build tag.

jacobhaven:
@jsha Unless you want to move this code to the [letsencrypt](https://github.com/letsencrypt) organization. That'd probably also be fine.

jsha:
Yeah, that's probably for the best since AFAIK we're the ones that depend on it most heavily, and will be doing the most modifications. In terms of copyright, I'll combine cfssl's LICENSE file and say "parts of this code are (c) Cloudflare, and parts are (c) Let's Encrypt?

jacobhaven:
@jsha I think we're just going to end up making a CF organization repo.

jsha:
Sorry for the delay: I've got it set up this morning (history and all, thanks to `git filter-branch`): https://github.com/letsencrypt/pkcs11key.

PR #521, opened 2016-02-22
Move pkcs11key dependency to github.com/letsencrypt/pkcs11key and vendor

jacobhaven:
This addresses #519


jacobhaven:
@jsha How would you suggest handling use of the `"github.com/cloudflare/cfssl/signer/pkcs11"` library? I.e. in [cli/config.go](https://github.com/cloudflare/cfssl/blob/jacob/vendor-pkcs11key/cli/config.go#L122). This is the last place that the `pkcs11` build tag is necessary.

jsha:
I would suggest removing pkcs11 support from the cli; I'm not aware of anyone using it. Similarly in the RootFromConfig in that file.

jsha:
I think we can also remove all of signer/pkcs11.

The pkcs11uri -> pkcs11key dependency is a little harder to break, and pkcs11uri isn't dependend upon by anythin else. We might want to break that into its own repo too?

jacobhaven:
Alright, I'll go ahead and remove `signer/pkcs11`. Like you said, there's no dependency on `helpers/pkcs11uri`, so feel free to move that to `"github.com/letsencrypt/pkcs11key/uri"` (or leave it, it doesn't really matter)

jacobhaven:
I'll remove the `helpers/pkcs11uri` package (and the vendored pkcs11key dependency) if https://github.com/letsencrypt/pkcs11key/pull/1 is merged.

jsha:
Done.

lziest:
lgtm

kisom:
LGTM.

PR #527, opened 2016-02-25
unvendor cfssl_trust

lziest:


lziest:
cfssl_trust is not imported anywhere.

jacobhaven:
It's used in the containers. I'd be happier modifying them so that they don't require network acess to `git clone https://github.com/cloudflare/cfssl_trust.git  /etc/cfssl`
We could instead modify them to copy the vendored directory to that location (Actually, I thought I did this...)

jacobhaven:
I added it to be used in #528

PR #529, opened 2016-02-25
log: Change SyslogWriter to SetLogger()

riking:
Allow cfssl/log to use interfaces for its logging, such as Boulder's
AuditLogger.

This is a breaking change because setting a package-level field to a struct
value was a poor decision from the beginning.

Existing users can update by replacing:

  cfssllog.SyslogWriter = $a

to:

  cfssllog.SetLogger($a)

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -9,7 +9,6 @@ import (
 	"flag"
 	"fmt"
 	"log"
-	"log/syslog"
 	"os"
 )
 
@@ -38,12 +37,31 @@ var levelPrefix = [...]string{
 	LevelFatal:    "FATAL",
 }
 
-var (
-	// Level stores the current logging level.
-	Level = LevelInfo
-	// SysLogger is a syslog Writer to be used if not nil.
-	SysLogger *syslog.Writer
-)
+// Level stores the current logging level.
+var Level = LevelInfo
+
+// SyslogWriter specifies the necessary methods for an alternate output
+// destination passed in via SetLogger.
+//
+// SyslogWriter is satisfied by *syslog.Writer.
+type SyslogWriter interface {
+	Debug(string) error
+	Info(string) error
+	Warning(string) error
+	Err(string) error
+	Crit(string) error
+	Emerg(string) error
+}
+
+// syslogWriter stores the SetLogger() parameter.
+var syslogWriter SyslogWriter
+
+// SetLogger sets the output used for output by this package.
+// A *syslog.Writer is a good choice for the logger parameter.
+// Call with a nil parameter to revert to default behavior.
+func SetLogger(logger SyslogWriter) {
+	syslogWriter = logger
+}
 
 func init() {
 	flag.IntVar(&Level, "loglevel", LevelInfo, "Log level (0 = DEBUG, 5 = FATAL)")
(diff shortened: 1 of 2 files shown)


log/log.go
```diff
@@ -41,31 +40,51 @@ var levelPrefix = [...]string{
...
-	SysLogger *syslog.Writer
+	// outputInterface stores a SetLogger() parameter.
+	outputInterface Interface
 )
 
+// Interface specifies the necessary methods for an alternate output
+// destination passed in via SetLogger.
+//
+// Interface is satisfied by syslog.Writer.
+type Interface interface {
```
jacobhaven: `log.Interface` is not a good name for a `syslog.Writer` interface. Maybe `SyslogWriter`?

Nitpick: it's satisfied by the pointer type `*syslog.Writer`

log/log.go
```diff
@@ -41,31 +40,51 @@ var levelPrefix = [...]string{
 var (
 	// Level stores the current logging level.
 	Level = LevelInfo
-	// SysLogger is a syslog Writer to be used if not nil.
-	SysLogger *syslog.Writer
+	// outputInterface stores a SetLogger() parameter.
+	outputInterface Interface
```
jacobhaven: I would still call this `sysLogger` or if you *really* want to be generic, just `output`.

log/log.go
```diff
@@ -41,31 +40,51 @@ var levelPrefix = [...]string{
...
+type Interface interface {
+	Debug(string) error
+	Info(string) error
+	Warning(string) error
+	Err(string) error
+	Crit(string) error
+	Emerg(string) error
+}
+
+// Set the log output used for output by this package.
```
jacobhaven:
```Go
// SetLogger sets the output used for output by this package.
// A *syslog.Writer is a good choice for the logger parameter.
```

jacobhaven:
This is a fine change. As far as I know there are no consumers yet of the old method.

riking:
@jacobhaven made suggested changes

log/log.go
```diff
@@ -41,31 +40,50 @@ var levelPrefix = [...]string{
 var (
 	// Level stores the current logging level.
 	Level = LevelInfo
-	// SysLogger is a syslog Writer to be used if not nil.
-	SysLogger *syslog.Writer
+	// output stores a SetLogger() parameter.
+	output SyslogWriter
```
jacobhaven: So... `output` probably wasn't the best suggestions (There's already a function named that)
Let's try: `syslogWriter`?

Also nitpick: I would move the type declaration before the variable declaration.

log/log.go
```diff
@@ -41,31 +40,50 @@ var levelPrefix = [...]string{
...
+}
+
 func init() {
 	flag.IntVar(&Level, "loglevel", LevelInfo, "Log level (0 = DEBUG, 5 = FATAL)")
 }
 
 func print(l int, msg string) {
 	if l >= Level {
-		if SysLogger != nil {
+		if outputInterface != nil {
```
jacobhaven: These need to be updated to match the variable name.

jacobhaven:
It would also be nice for you to create a testing struct that satisfies `SysLogger` (probably backed by a `bytes.Buffer`) in `log_test.go`

log/log.go
```diff
@@ -38,12 +37,31 @@ var levelPrefix = [...]string{
...
+type SyslogWriter interface {
+	Debug(string) error
+	Info(string) error
+	Warning(string) error
+	Err(string) error
+	Crit(string) error
+	Emerg(string) error
+}
+
+// output stores a SetLogger() parameter.
```
jacobhaven: `// syslogWriter [...]`

riking:
Updated & squashed.

log/log_test.go
```diff
@@ -136,3 +136,57 @@ func TestDebug(t *testing.T) {
...
+	SetLogger(testSyslogger{buf})
+	Level = LevelDebug
+	outputf(LevelDebug, teststring, nil)
+
+	// outputf correctly prints string
+	if !strings.Contains(buf.String(), teststring) {
+		t.Fail()
+	}
+	SetLogger(nil)
+	return
```
jacobhaven: Nitpick: this `return` does nothing

jacobhaven:
Travis error is an unrelated data race in the bundler testsuite due to a change earlier today.
LGTM.

Issue #530, opened 2016-02-25
Responder should return unauthorized for a response with NextUpdate in the future


rolandshoemaker:
RFC 2560 (and 6960) specifies `[OCSP r]esponses whose nextUpdate value is earlier than the local system time value SHOULD be considered unreliable`. RFC 5019 also states `in order to ensure the database of revocation information does not grow unbounded over time, the responder MAY remove the status records of expired certificates`.

Since the window between a `nextUpdate` in the past and when a response can be removed is somewhat ambiguous it would be a good idea to add a config var to the responder specifying if it should happly serve stale responses, how long it should be allowed to serve stale responses for, or not to serve stale responses at all.

PR #531, opened 2016-02-26
Responder caching refactor

rolandshoemaker:
Always adds relevant `Cache-Control` headers instead of only when a valid response is returned to the client. Also adds the RFC 5019 specified `Cache-Control` modifiers to prevent intermediary proxies/caches from mangling responses and switches `ProducedAt` for `ThisUpdate` as specified in RFC 2560 for use in the `Last-Modified` header.

Finally this checks that the `NextUpdate` field is in the future before setting max-age as this previously allowed negative max-age parameters (In the case of a stale response the responder will return `max-age=0` for now, although stale responses should probably just not be returned, opened #530 to discuss).

Unblocks work on letsencrypt/boulder#1485.

### DIFF
--- a/ocsp/responder.go
+++ b/ocsp/responder.go
@@ -104,6 +104,17 @@ func NewResponder(source Source) *Responder {
 // strings of repeated '/' into a single '/', which will break the base64
 // encoding.
 func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Request) {
+	var maxAge *int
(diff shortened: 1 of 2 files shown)


lziest:
lgtm

riking:
@rolandshoemaker @lziest 

This is broken - the deferred function won't execute until after `response.WriteHeader()` has happened. You need to move that code out of the defer func.

PR #533, opened 2016-02-26
Revert "update/search intermediates in a goroutine when bundling cert…

lziest:
…ificate chains"

This reverts commit 45c9a02ec34286a2bb8a96b215f412b186954615.

### DIFF
--- a/bundler/bundler.go
+++ b/bundler/bundler.go
@@ -602,7 +602,7 @@ func (b *Bundler) Bundle(certs []*x509.Certificate, key crypto.Signer, flavor Bu
 
 		// verify and store input intermediates to the intermediate pool.
 		// Ignore the returned error here, will treat it in the second call.
-		go b.fetchIntermediates(certs)
+		b.fetchIntermediates(certs)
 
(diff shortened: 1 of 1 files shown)


bundler/bundler.go
```diff
@@ -602,7 +602,7 @@ func (b *Bundler) Bundle(certs []*x509.Certificate, key crypto.Signer, flavor Bu
 
 		// verify and store input intermediates to the intermediate pool.
 		// Ignore the returned error here, will treat it in the second call.
-		go b.fetchIntermediates(certs)
+		b.fetchIntermediates(certs)
```
jacobhaven: Why don't we store the error (but not fail on it) here, so we don't have to make a second call later?

lziest:
The intention was, if there is a newer intermediate issuing a leaf, we can use AIA record to find out, even if the leaf cert can be bundled correctly with existing intermediates, or with a longer chain.

The fetchIntermediates was on error handling logic, when bundling failed. I think I want to rewrite this part.

PR #534, opened 2016-02-26
Fix data races and add "optimitic" first round of bundling

jacobhaven:


lziest:
there are testserver code that won't pass race check.

jacobhaven:
@lziest Well, we should fix those.

lziest:
LGTM

PR #535, opened 2016-02-26
Move Cache-Control header setting before WriteHeader calls

rolandshoemaker:
#531 wrote the headers after `WriteHeader` calls, which in normal operation mean they won't be properly set and returned to the client. Since `httptest` seems to not properly follow this behavior tests still passed but the headers would [never actually be returned](https://godoc.org/net/http#ResponseWriter).

This moves the adding of the headers before any `WriteHeader` or `Write` calls so that the they are properly added to the response. Sorry about that!

### DIFF
--- a/ocsp/responder.go
+++ b/ocsp/responder.go
@@ -104,17 +104,11 @@ func NewResponder(source Source) *Responder {
 // strings of repeated '/' into a single '/', which will break the base64
 // encoding.
 func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Request) {
-	var maxAge *int
-	defer func() {
-		value := "max-age=0, no-cache"
-		if maxAge != nil {
-			value = fmt.Sprintf(
-				"max-age=%d, public, no-transform, must-revalidate",
-				*maxAge,
-			)
-		}
-		response.Header().Add("Cache-Control", value)
-	}()
+	// By default we set a 'max-age=0, no-cache' Cache-Control header, this
+	// is only returned to the client if a valid authorized OCSP response
+	// is not found or an error is returned. If a response if found the header
+	// will be altered to contain the proper max-age and modifiers.
+	response.Header().Add("Cache-Control", "max-age=0, no-cache")
 	// Read response from request
 	var requestBody []byte
 	var err error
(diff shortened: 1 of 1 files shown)


lziest:
Thanks, rule of thumb for me these days is never review code after 8pm. LGTM

lziest:
@riking please review too.

riking:
Why not use Header.Set() at the top and overwrite it once you know the max-age value?

riking:
so:

```
response.Header().Set("Cache-Control", "max-age=0, no-cache")
// ...
if now.Before(parsedResponse.NextUpdate) {
	expiresIn := int(parsedResponse.NextUpdate.Sub(now) / time.Second)
	response.Header().Set("Cache-Control", fmt.Sprintf(..., expiresIn))
} else {
	// XXX: we want max-age=0 but since this is technically an authorized OCSP response
	// (despite being stale) and 5019 forbids attaching no-cache we get a little tricky
	response.Header().Set("Cache-Control", "max-age=0, public, no-transform, must-revalidate")
}
```

ocsp/responder.go
```diff
@@ -195,10 +202,10 @@ func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Reques
 	now := rs.clk.Now()
 	if now.Before(parsedResponse.NextUpdate) {
 		expiresIn := int(parsedResponse.NextUpdate.Sub(now) / time.Second)
-		maxAge = &expiresIn
+		addCacheHeader(response, &expiresIn)
 	} else {
-		zero := 0      // XXX: we want max-age=0 but since this is technically an authorized OCSP response
-		maxAge = &zero //      (despite being stale) and 5019 forbids attaching no-cache we get a little tricky
+		zero := 0                       // XXX: we want max-age=0 but since this is technically an authorized OCSP response
```
jsha: At least in Boulder, we use XXX to indicate "Don't merge this," and TODO to indicates long-term TODO's in merged code.

ocsp/responder.go
```diff
@@ -93,6 +93,17 @@ func NewResponder(source Source) *Responder {
 	}
 }
 
+func addCacheHeader(response http.ResponseWriter, maxAge *int) {
```
jsha: This pattern of passing a pointer to an int is weird. How about instead take a regular int and:

```
if maxAge == 0 {
  ...max-age=0, no-cache
} else {
  ...max-age=%d
```

rolandshoemaker:
Pushed simplified logic.

jsha:
LGTM

lziest:
lgtm, can you squash it?

rolandshoemaker:
Done.

jsha:
I missed a comment on this: Since `httptest` isn't capable of testing the bug here, is there another way of writing the test such that the bug actually does get caught?

PR #537, opened 2016-02-28
Basic ETag support for OCSP responder (5019-ish behavior)

rolandshoemaker:
Adds (optional) basic ETag support (as recommended in RFC 5019). Note this is a best effort approach without `NextPublish` support, it might make sense to make the aggressiveness of the ETag validity period tunable using a config var but wanted to get feedback before trudging on with that.

### DIFF
--- a/ocsp/responder.go
+++ b/ocsp/responder.go
@@ -1,6 +1,7 @@
 package ocsp
 
 import (
+	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
@@ -202,6 +203,18 @@ func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Reques
 			maxAge,
 		),
 	)
+	responseHash := sha256.Sum256(ocspResponse)
+	response.Header().Add("ETag", fmt.Sprintf("\"%X\"", responseHash))
+
+	// RFC 7232 says that a 304 response must contain the above
+	// headers if they would also be sent for a 200 for the same
+	// request, so we have to wait until here to do this
+	if etag := request.Header.Get("If-None-Match"); etag != "" {
+		if etag == fmt.Sprintf("\"%X\"", responseHash) {
+			response.WriteHeader(http.StatusNotModified)
+			return
+		}
+	}
 	response.WriteHeader(http.StatusOK)
 	response.Write(ocspResponse)
 }
--- a/ocsp/responder_test.go
+++ b/ocsp/responder_test.go
@@ -95,19 +95,36 @@ func TestCacheHeaders(t *testing.T) {
 		{"Last-Modified", "Tue, 20 Oct 2015 00:00:00 UTC"},
 		{"Expires", "Sun, 20 Oct 2030 00:00:00 UTC"},
 		{"Cache-Control", "max-age=471398400, public, no-transform, must-revalidate"},
+		{"Etag", "\"8169FB0843B081A76E9F6F13FD70C8411597BEACF8B182136FFDD19FBD26140A\""},
 	}
 	for _, tc := range testCases {
 		headers, ok := rw.HeaderMap[tc.header]
 		if !ok {
 			t.Errorf("Header %s missing from HTTP response", tc.header)
+			continue
 		}
 		if len(headers) != 1 {
 			t.Errorf("Wrong number of headers in HTTP response. Wanted 1, got %d", len(headers))
+			continue
 		}
 		actual := headers[0]
 		if actual != tc.value {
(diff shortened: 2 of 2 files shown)


ocsp/responder.go
```diff
@@ -22,6 +24,43 @@ var (
...
 )
 
+type eTagPool struct {
+	pool map[string]time.Time
+	clk  clock.Clock
+	mu   *sync.RWMutex
+}
+
+func (etp *eTagPool) stillValid(etag string) bool {
+	etp.mu.RLock()
```
kisom: Perhaps better to `defer etp.mu.Unlock()` here.

ocsp/responder.go
```diff
@@ -22,6 +24,43 @@ var (
...
+	if expires, present := etp.pool[etag]; present {
+		if expires.Before(etp.clk.Now()) {
+			etp.mu.RUnlock()
+			etp.remove(etag)
+			return false
```
kisom: This is missing an unlock.

ocsp/responder.go
```diff
@@ -82,15 +121,20 @@ func NewSourceFromFile(responseFile string) (Source, error) {
...
 // NewResponder instantiates a Responder with the give Source.
-func NewResponder(source Source) *Responder {
-	return &Responder{
+func NewResponder(source Source, useETags bool) *Responder {
+	rs := &Responder{
 		Source: source,
 		clk:    clock.Default(),
 	}
+	if useETags {
+		rs.eTags = &eTagPool{make(map[string]time.Time), rs.clk, new(sync.RWMutex)}
```
kisom: It might be better to add a `newETagPool` function that handles setting up a new `eTagPool`; it would be cleaner use field names and easier to update in one place if the definition of an `eTagPool` changes.

ocsp/responder.go
```diff
@@ -147,6 +191,13 @@ func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Reques
...
 		return
 	}
+	if rs.eTags != nil && request.Header.Get("If-None-Match") != "" {
+		etag := request.Header.Get("If-None-Match")
+		if rs.eTags.stillValid(etag) {
+			response.WriteHeader(http.StatusNotModified)
+			return
+		}
+	}
 	// TODO log request
```
kisom: It looks like this TODO is done two lines below?

ocsp/responder_test.go
```diff
@@ -76,6 +77,7 @@ func TestCacheHeaders(t *testing.T) {
 	responder := Responder{
 		Source: source,
 		clk:    fc,
+		eTags:  &eTagPool{make(map[string]time.Time), fc, new(sync.RWMutex)},
```
kisom: Another candidate for use of a `newETagPool` function.

jsha:
I am not sure that the utility of this is worth the extra complexity. This effectively implements an in-memory cache of the Source, which is usually a database. If we need more efficient DB access, we can just use a read replica that has enough memory to keep all recently-used results in memory.

Also, I suspect the hit rate on this in-memory cache will be very low because we expect to sit behind an HTTP cache, which would get all the common requests.

I think a nicer, simpler implementation would be to omit all the inmemory caching semantics, always hit the Source, and return NotModified if the result from the Source matches the ETag.

rolandshoemaker:
*kicks self*, yup that's a much cleaner approach, I'll rework this tonight.

rolandshoemaker:
That should be better.

kisom:
LGTM.

ocsp/responder.go
```diff
@@ -202,6 +203,18 @@ func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Reques
 			maxAge,
 		),
 	)
+	responseHash := sha256.Sum256(ocspResponse)
+	response.Header().Add("ETag", fmt.Sprintf("\"%X\"", responseHash))
```
jsha: You should extract out this repeated Sprintf.

ocsp/responder_test.go
```diff
@@ -95,20 +95,37 @@ func TestCacheHeaders(t *testing.T) {
...
 			t.Errorf("Wrong number of headers in HTTP response. Wanted 1, got %d", len(headers))
+			continue
 		}
 		actual := headers[0]
 		if actual != tc.value {
 			t.Errorf("Got header %s: %s. Expected %s", tc.header, actual, tc.value)
 		}
 	}
+
+	rw = httptest.NewRecorder()
```
jsha: This should really be a new test function.

jsha:
Other than the above two nits, LGTM.

PR #541, opened 2016-03-02
Fix loglevel redefinition bug

jacobhaven:
The `"loglevel"` flag could be defined multiple times if multiple packages import the `log` pacakage. (e.g. `A` imports `"log", "B"` and `B` imports `"log"`)

 We can prevent this using a `sync.Once`

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


lziest:
LGTM

lziest:
looks like the once.do usage is incorrect.

PR #542, opened 2016-03-02
Better fix of loglevel redefinition bug

jacobhaven:
When vendoring, the `sync.Once` variable could be defined multiple times. Since `init` functions run in a single thread, we can simply check if the flag has already been defined before defining it.


lziest:
lgtm

Issue #545, opened 2016-03-08
CGO dependency in github.com/mattn/go-sqlite3 preventing cross-compiliation


jacobhaven:
```
$ script/build -os="linux" -arch="386"
[...]
Number of parallel builds: 1

-->       linux/386: github.com/cloudflare/cfssl/cmd/multirootca
-->       linux/386: github.com/cloudflare/cfssl/cmd/cfssl-newkey
-->       linux/386: github.com/cloudflare/cfssl/cmd/cfssl
-->       linux/386: github.com/cloudflare/cfssl/cmd/cfssl-bundle
-->       linux/386: github.com/cloudflare/cfssl/cmd/cfssl-certinfo
-->       linux/386: github.com/cloudflare/cfssl/cmd/cfssljson
-->       linux/386: github.com/cloudflare/cfssl/cmd/cfssl-scan
-->       linux/386: github.com/cloudflare/cfssl/cmd/mkbundle

2 errors occurred:
--> linux/386 error: exit status 1
Stderr: certdb/dbconf/db_config.go:12:2: C source files not allowed when not using cgo or SWIG: sqlite3-binding.c

--> linux/386 error: exit status 1
Stderr: certdb/dbconf/db_config.go:12:2: C source files not allowed when not using cgo or SWIG: sqlite3-binding.c
```

jacobhaven:
Removing the `"github.com/mattn/go-sqlite3"` dependency solves this, but removes the ability to use a SQLite DB. (Is there a SQLite driver that doesn't depend on CGO?)

lziest:
should be fixed.

PR #546, opened 2016-03-09
certdb/dbconf package should not touch specific db drivers

lziest:
Trying to address #545 @jacobhaven can you verify that work? I have tested locally and it seemed to work.


lziest:
OK, need to fix tests

lziest:
OK I think the PR is ready.

kisom:
LGTM.

lziest:
verified that it solves #545 since we get rid of sqlite support in cfssl binaries, sqlite is only used for testing.

PR #547, opened 2016-03-09
Unbreak ocspserve by providing a properly initialised Responder object

mathieuk:
ocspserve was crashing on a call to `rs.clk.Now()` because `rs.clk` was nil due to being initialised manually instead of using `ocsp.NewResponder()`.


lziest:
Thanks. LGTM, will merge once CI passes.

PR #548, opened 2016-03-11
check revocation signature in package revoke

### LINKED ISSUE #338: revoke: does not verify authenticity of CRLs or OSCP responses
Package revoke does not perform any kind of cryptographic verification on fetched CRLs or OCSP responses. An attacker in a privileged network position can spoof unrelated CRLs or OCSP responses and cause package revoke to treat arbitrary certificates as having been revoked.

lziest:
fix #338

### DIFF
--- a/revoke/revoke.go
+++ b/revoke/revoke.go
@@ -1,7 +1,6 @@
 // Package revoke provides functionality for checking the validity of
 // a cert. Specifically, the temporal validity of the certificate is
-// checked first, then any CRL in the cert is checked. OCSP is not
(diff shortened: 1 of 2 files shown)


kisom:
LGTM. Is there a test that validates the signature verification?

lziest:
Replace the good cert test case with a Comodo Intermediate CA cert. It has issuer url, CRL url and OCSP url. Verified through coverage tool that issuer certificate is used to verify the revocation result.

kisom:
Great, thanks!

PR #549, opened 2016-03-12
Include Issuer and SerialNumber in certinfo

davitb:
Certificate Issuer and SerialNumber are missing from certinfo. Would be great to report these as well.


lziest:
LGTM

PR #553, opened 2016-03-17
Add revocation checking to transport.Dial.

kisom:
This adds revocation checking in the client-side only; right now, the crypto/tls package doesn't facilitate performing revocation checks on client certificates in the same way. Users of this package will have to manually call the revocation check on new connections for now.


grittygrease:
Looks good

PR #557, opened 2016-03-20
Fix CI for forked repos

drcapulet:
Sadly if any package wants to pass around a type from a vendored package, everything breaks - figured moving the source was the easiest option.


grittygrease:
LGTM

PR #562, opened 2016-03-24
Add MySQL support to certdb

drcapulet:


drcapulet:
Ping @kisom / @lziest

vendor/github.com/go-sql-driver/mysql/LICENSE
```diff
@@ -0,0 +1,24 @@
+Copyright (c) 2014 CloudFlare Inc.
```
kisom: Is this the right LICENSE file?

kisom:
@drcapulet just an FYI, currently reviewing this. It's taking a while as there's a lot of vendored code to read through.

vendor/github.com/go-sql-driver/mysql/LICENSE
drcapulet: Nice catch, I updated `gvt` and it solved the problem.

drcapulet:
@kisom Awesome, thanks!

kisom:
LGTM,.

deanefrati:
any idea when this will be merged? we would really like to see cfssl support mysql since it will be easier to stand up a clustered implementation this way using mariadb.

lziest:
LGTM, so the SQL statements we put down works with mysql as well? I am happy and feeling unreal.

lziest:
still have conflicts though

kisom:
@deanefrati @drcapulet Once this is rebased, it can be merged.

drcapulet:
@kisom / @lziest rebased, should be good now! The SQL statements work due to the changes a little while back to use sqlx since it knows how to handle different drivers - in theory you could also add Oracle support fairly quickly as well.

kisom:
Thanks!

PR #570, opened 2016-04-01
gencert: take into account csr hosts for CAB host warning

tam7t:
When creating certificate through the CLI with a CSR that has `hosts` (like [testdata/csr.json](https://github.com/cloudflare/cfssl/blob/master/testdata/csr.json)) `cfssl` would incorrectly log the following warning:

```
This certificate lacks a "hosts" field. This makes it unsuitable for
websites. For more information see the Baseline Requirements for the Issuance and Management
of Publicly-Trusted Certificates, v.1.1.6, from the CA/Browser Forum (https://cabforum.org);
specifically, section 10.2.3 ("Information Requirements").
``` 

This PR corrects that by inspecting both the `-hostname` argument _and_ the `hosts` filed on the supplied CSR.

### DIFF
--- a/cli/gencert/gencert.go
+++ b/cli/gencert/gencert.go
@@ -119,14 +119,14 @@ func gencertMain(args []string, c cli.Config) error {
 		}
 
 		var cert []byte
-		req := signer.SignRequest{
+		signReq := signer.SignRequest{
 			Request: string(csrBytes),
 			Hosts:   signer.SplitHosts(c.Hostname),
 			Profile: c.Profile,
 			Label:   c.Label,
 		}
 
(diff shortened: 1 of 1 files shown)


kisom:
I think it would be useful to populate the `signReq.Hosts` field with the contents of `req.Hosts` if `c.Hostnames` is empty.

tam7t:
@kisom the `OverrideHosts` function in https://github.com/cloudflare/cfssl/blob/master/signer/local/local.go#L246 effectively does that. I didn't want to duplicate/leak that logic into the `cli` presentation code.

kisom:
@tam7t You're right, I mis-remembered. LGTM.

PR #572, opened 2016-04-01
Remove data race in initca (fixed #571).

### LINKED ISSUE #571: Data race encountered in FillTemplate
During development, we encountered the following data race during testing:

```
WARNING: DATA RACE
Read by goroutine 34:
  github.com/cloudflare/cfssl/signer.FillTemplate()
      /home/diogo/go/src/github.com/cloudflare/cfssl/signer/signer.go:231 +0x15b
  github.com/cloudflare/cfssl/signer/local.(*Signer).sign()
      /home/diogo/go/src/github.com/cloudflare/cfssl/signer/local/local.go:100 +0xc5
  github.com/cloudflare/cfssl/signer/local.(*Signer).Sign()
      /home/diogo/go/src/github.com/cloudflare/cfssl/signer/local/local.go:360 +0x244e
  github.com/cloudflare/cfssl/initca.New()
      /home/diogo/go/src/github.com/cloudflare/cfssl/initca/initca.go:83 +0x95a
  ...  
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:473 +0xdc

Previous write by goroutine 50:
  github.com/cloudflare/cfssl/signer.FillTemplate()
      /home/diogo/go/src/github.com/cloudflare/cfssl/signer/signer.go:232 +0x17b
  github.com/cloudflare/cfssl/signer/local.(*Signer).sign()
      /home/diogo/go/src/github.com/cloudflare/cfssl/signer/local/local.go:100 +0xc5
  github.com/cloudflare/cfssl/signer/local.(*Signer).Sign()
      /home/diogo/go/src/github.com/cloudflare/cfssl/signer/local/local.go:360 +0x244e
  github.com/cloudflare/cfssl/initca.New()
      /home/diogo/go/src/github.com/cloudflare/cfssl/initca/initca.go:83 +0x95a
  ...
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:473 +0xdc
```

The source of the problem seems to be when the [global CAPolicy is set](https://github.com/cloudflare/cfssl/blob/bd27f2753e0aedf08d75b212c3fdbeebc9adb8c9/initca/initca.go#L80). Later, this is set by `signer.FillTemplate` concurrently. Because this is operating on a global, both goroutines share the same instance when the tests are run in parallel.

I do not have a minimal reproduction.

The likely fix here is to make a copy of the default signing profile in `initca.New`. 

cc @diogomonica @grittygrease

kisom:
The default policy was previously a global variable, which caused
issues when multiple CAs were being initialised in parallel (such
as in testing, e.g. [1]).

This change causes the default policy to be a function returning a
default policy. The function can still be swapped out in testing as
needed, but the functions now operate on local copies of the default
policy. A change to the default policy (which is only used in testing)
can now be accomplished by changing the function itself.

[1] https://github.com/cloudflare/cfssl/issues/571

### DIFF
--- a/initca/initca.go
+++ b/initca/initca.go
@@ -47,10 +47,11 @@ func validator(req *csr.CertificateRequest) error {
 
 // New creates a new root certificate from the certificate request.
 func New(req *csr.CertificateRequest) (cert, csrPEM, key []byte, err error) {
+	policy := CAPolicy()
 	if req.CA != nil {
 		if req.CA.Expiry != "" {
-			CAPolicy.Default.ExpiryString = req.CA.Expiry
-			CAPolicy.Default.Expiry, err = time.ParseDuration(req.CA.Expiry)
+			policy.Default.ExpiryString = req.CA.Expiry
+			policy.Default.Expiry, err = time.ParseDuration(req.CA.Expiry)
 		}
 
 		if req.CA.PathLength != 0 {
@@ -77,7 +78,7 @@ func New(req *csr.CertificateRequest) (cert, csrPEM, key []byte, err error) {
 		log.Errorf("failed to create signer: %v", err)
 		return
 	}
-	s.SetPolicy(CAPolicy)
+	s.SetPolicy(policy)
 
 	signReq := signer.SignRequest{Hosts: req.Hosts, Request: string(csrPEM)}
 	cert, err = s.Sign(signReq)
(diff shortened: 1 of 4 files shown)


stevvooe:
LGTM

Were you able to reproduce?

kisom:
@stevvooe I was not able to repro, but it's clear from the data race traceback how it could happen. Better safe than sorry, especially with such a short fix.

stevvooe:
Thanks for the quick fix!

If you want us to confirm before merging, let me know.

kisom:
I'd prefer if you could confirm this, just so that if the problem still exists we can fix it and document the cause in this PR. Thanks!

jsha:
This is kind of an awkware fix for the problem. As I read it, the culprit is this write in signer.go:

```

	// The third value returned from Usages is a list of unknown key usages.
	// This should be used when validating the profile at load, and isn't used
	// here.
	ku, eku, _ = profile.Usages()
	if profile.IssuerURL == nil {
		profile.IssuerURL = defaultProfile.IssuerURL
	}
```

The problem is that profile may refer to CAPolicy.Default, which initca implicitly considers immutable. In general, this is a surprising side-effect of FillTemplate, that it could modify one of its profile inputs.

Instead, I think a better fix would be to declare a local variable `issuerURL`, which is set to `profile.IssuerURL` if available, or `defaultProfile.IssuerURL` otherwise. Then use that local variable elsewhere in FillTemplate. Then you fix both the unexpected side effect and the race condition.

jsha:
BTW, I'd recommend introducing the `-race` flag and `GORACE="halt_on_error=1"` environment variable into your Travis tests, to help find such race conditions in the future.

kisom:
The immutable assumption in initca should, in my opinion, either be backed up by a `const` statement (which Go doesn't allow here), or the way it's being used now. Then the immutability is guaranteed, and it requires modifications to be done locally.

I do agree that a local `issuerURL` is a good solution here, as is adding the race detector. I've added both in addition to the initca fix.

jsha:
LGTM

Issue #575, opened 2016-04-07
support both pathlen=0 and unspecified pathlen when doing initca


lziest:
Use the new flag "MaxPathLenZero bool" in https://golang.org/pkg/crypto/x509/#Certificate to distinguish these two cases.

PR #576, opened 2016-04-07
#575 initca now supports issuing CA certificate with pathlen=0

lziest:


lziest:
The command "go get golang.org/x/tools/cmd/vet" failed and exited with 1, O_O

kisom:
@lziest can you rebase with master? I've fixed this on master.

lziest:
done

fhossain:
@lziest thanks for the change. I tested your branch and I can see that the self signed certificate after initca is created with pathlen = 0. However, the next step is to to sign with the Root or another intermediate CA -
cfssl sign -ca=ca_root.pem -ca-key=ca_root-key.pem -config=config/ca-config.json -profile=ca ca_int.csr | cfssljson -bare ca_int
The new signed intermediate certificate after running the above command is still created with pathlen = 1. The above command generates the final signed certificate from a CSR but I am not sure if pathlen=0 can be retrieved from a CSR. If that is not possible it will have to come form a profile i.e. ca0. A profile has option to specify "is_ca": true but for ca0 profile we would need to be able to specify 'pathlen' and 'pathlenzero' as well.

lziest:
@fhossain Yes, intermediate CA is still issued with pathlen=1 (hardcoded), next step is to enable that, this PR is only for init root CA.

kisom:
LGTM.

fhossain:
@lziest awesome. Sorry, for jumping the gun.

PR #578, opened 2016-04-07
Adjust travis config for Go tool moves.

kisom:
c.f. https://golang.org/doc/go1.5#moving


jsha:
LGTM. We made the same change in Boulder to remove `vet`.

PR #583, opened 2016-04-08
Remove error return from log functions.

jsha:
These don't need to be an intrinsic part of the logging interface. In Boulder
we're removing the error returns from our logging interface, in favor of
handling internally any failures to write to syslog (by writing to stderr
instead).


kisom:
LGTM.

Issue #495, opened 2016-01-23
Option to specify pathlen during sign or gencert for Intermediate CA


zaninime:
I didn't find an option to specify (aka change from the default value of 1) the pathlen constraint while issuing/signing a CA. Am I missing smth or it's just not implemented?

Thanks!

dalbani:
See #417.

benwhite-deltas:
'Intermediate CA' is the key here - cfssl prevents you from specifying a `ca` section, and therefore a `pathlen`, in the csr JSON file for anything other than a initca - see [this gist](https://gist.github.com/benwhite-deltas/a2ee7c5107bcc5a350bf541e363ac0ac) for reproduceable result.

This is odd, as after [some investigation](http://security.stackexchange.com/questions/119577/should-openssl-verify-recognise-an-exceeded-path-length-constraint-on-a-root) it seems Root CA certs' `pathlen` value is ignored whereas intermediate CAs `pathlen` value is respected.

I would hazard a guess that the cause of this issue is that the constraint of not allowing a `ca` section in a non-root csr should only be applied when the `is_ca` value of the policy being applied is `false`

tam7t:
This is how I generate root/intermediate certs with `pathlen`'s:

### Generate a root (self signed)

`./root/root.json`
```
{
  "CN": "tam7t Root CA",
  "ca": {
    "pathlen": 2
  },
  "names": [
    {
      "C": "US",
      "L": "San Francisco",
      "O": "CloudFlare",
      "OU": "Systems Engineering",
      "ST": "California"
    }
  ]
}
```
```
$ cfssl genkey -initca root/root.json | cfssljson -bare root/root
```

### Generate an issuing certificate signing request

`./issuer/issuer.json`
```
{
  "CN": "tam7t Issuer CA",
  "ca": {
    "pathlen": 1
  },
  "names": [
    {
      "C": "US",
      "L": "San Francisco",
      "O": "CloudFlare",
      "OU": "Systems Engineering",
      "ST": "California"
    }
  ]
}
```

`./issuer/conf.json`
```
{
 "signing": {
   "default": {
     "usages": [
       "signing",
       "key encipherment",
       "cert sign",
       "crl sign"
      ],
      "expiry": "4320h",
      "is_ca": true
    }
  }
}
```

```
$ cfssl genkey -initca issuer/issuer.json | cfssljson -bare issuer/issuer
[...]

PR #586, opened 2016-04-09
BasicConstraint CSR support

tam7t:
Addresses https://github.com/cloudflare/cfssl/issues/495

This adds more complete support for the `BasicConstraint` extension when generating and singing CSRs. Continues from the work in https://github.com/cloudflare/cfssl/pull/576 adding...
- `pathlen` to the CSRs generated by cfssl
- use the `pathlen` info from CSRs when signing requests


tam7t:
@lziest would appreciate a look

lziest:
Yes, will definitely review it tomorrow.

signer/local/testdata/inter_pathlen_0-key.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN EC PRIVATE KEY-----
```
lziest: I realize you would never use the -key and cert.pem in those local signer tests, better not check them in. Just keep the generated around.

lziest:
I see the approach here is to let CSR contains CA pathlen info and pass it down to the final cert. The other side of control is the CA signing config, where we define how signer controls what the final certificate should look like. We still want the signing config take priority over CSR value, i.e. signing config should check/override CSR.

Can you simplify the CSR part of the code to have appendCAInfoToCSR only append unspecified pathlen, and let signer to decide what value should be filled in?

benwhite-deltas:
Thanks for movement on this. If I may comment from a user perspective:

> We still want the signing config take priority over CSR value

I think this is good, provided the signing config pathlen value is lower than the CSR pathlen value - ultimately I believe the lower of the two should be used if they are both specified, as the signing authority shouldn't generate a certificate with a pathlen longer than that being requested.

Additionally it would be useful to reflect back to the user a warning if the pathlen value requested in the CSR has been lowered by the signing value.

tam7t:
I think that the CSR should contain the requested `pathlen` so that cfssl generated CSRs would be compatible with other certificate authorities (in case I wanted to generate an intermediate and leaf certs with cfssl, but have it signed by my my organization's active directory CA).

I can add a check/warning that the requested cert has a lower pathlen than the signing cert (if the singing cert has a pathlen restriction).

signer/local/testdata/inter_pathlen_0-key.pem
tam7t: fixed!

csr/csr.go
```diff
@@ -176,6 +177,12 @@ func (cr *CertificateRequest) Name() pkix.Name {
 	return name
 }
 
+// BasicConstraints CSR information RFC 5280, 4.2.1.9
+type BasicConstraints struct {
```
lziest: do you think this can be an unexported struct?

lziest:
rebase on top of current master may help me see the diff better

lziest:
I agree we should make CSR more complete/compatible so it works better with other existing CAs.

tam7t:
This PR is pretty large, I'm thinking I will split it into ~3 smaller PR's that build off eachother and should hopefully be easier to review.

kisom:
@tam7t That would help out a lot --- I've had a hard time finding time to sit down and go through this.

PR #587, opened 2016-04-09
Register sqlite backend

vespian:
This fixes "unknown driver `sqlite`" error while trying to use sqlite backend.


kisom:
LGTM. Thanks!

lziest:
I had a report from @jacobhaven that including sqlite will result in docker issue. So I am a little concerned about this change.

PR #593, opened 2016-04-12
Fix documented command for installing goose

srstevenson:
The command specified in the documentation to install goose fails due to the protocol prefix (go 1.6):

    % go get https://bitbucket.org/liamstask/goose/
    package https:/bitbucket.org/liamstask/goose: "https://" not allowed in import path

The [upstream goose repository](https://bitbucket.org/liamstask/goose/) specifies a different command, which works:

    % go get bitbucket.org/liamstask/goose/cmd/goose
    % goose

    goose is a database migration management system for Go projects.
    ...

### DIFF
--- a/certdb/README.md
+++ b/certdb/README.md
@@ -21,7 +21,7 @@ Currently supported:
 
 ### Get goose
 
-    go get https://bitbucket.org/liamstask/goose/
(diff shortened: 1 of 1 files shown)


kisom:
Thanks!

PR #597, opened 2016-04-15
[WIP] Add HTTP/2 and SPDY Reporting

### LINKED ISSUE #444: Add HTTP/2 and SPDY reporting
CFSSL scan should report the server's capability for HTTP/2 and SPDY using the ALPN extension.

https://github.com/jgrahamc/h2scan

APTy:
This PR improves `cfssl scan` by using ALPN to determine which application-layer protocols that a server supports (among HTTP/1, HTTP/2, and SPDY/3). The `ALPN` scanner is introduced as part of the `TLSHandshake` family of scanners (might need to be in a new family of `TLSExtensions`).

**Example of `cfssl scan`**

```
$ cfssl scan -scanner=ALPN cloudflare.com
Scanning cloudflare.com...
=== cloudflare.com ===
{
  "TLSHandshake": {
    "ALPN": {
      "grade": "Good",
      "output": {
        "HTTP/1": true,
        "HTTP/2": true,
        "SPDY/3": true
      }
    }
  }
}
```

**Example of `cfssl scan -list`**

```
$ cfssl scan -list

...

"TLSHandshake": {
    "description": "Scans for host's SSL/TLS version and cipher suite negotiation",
    "scanners": {
      "ALPN": {
        "description": "Determines host's supported application-layer protocols"
      },

...
```

Resolves #444


APTy:
I'm getting a funny issue where the tests added in 7629cea work in isolation (say, in their own package), but fail with `handshake error` when run inside the `scan/` package. Any thoughts on this or any global config that could be having an effect? (mentioning @grittygrease since you filed this issue, but let me know if anyone else has bandwidth to take a look) Thanks!

kisom:
@APTy I run into this issue even when testing from the top level:

```
~/src/github.com/cloudflare/cfssl/
(0) <straka:kyle> $ go test ./...
# some output elided
--- FAIL: TestTLSConnection (1.00s)
        tls_handshake_test.go:82: client error at Dial(): remote error: handshake failure
FAIL
FAIL    github.com/cloudflare/cfssl/scan        1.004s
# more output elided
```

APTy:
@kisom Yeah I definitely am getting that same issue when I run this test. But it runs fine when its not inside the `cfssl` package generally. I've tried digging into any project-specific TLS settings that could be the cause, but have had no luck so far. Any thoughts?

APTy:
Btw I noticed that adding the following crypto libs to `scan/vendor/` fixes that specific error.
```
crypto/aes/
crypto/des/
crypto/ecdsa/
crypto/elliptic/
crypto/x509/
```
This does break other parts of the package, so my guess is there is some weird interplay between the various vendor files in the project. Looking into it

lziest:
orphaned PR?

PR #607, opened 2016-04-30
Updates to transport package.

kisom:
1. The [`KeyProvider`](https://godoc.org/github.com/cloudflare/cfssl/transport/kp) interface now includes a method `SignalFailure`. This was needed to support a failover system in a certain `KeyProvider` implementation. The `StandardProvider` has been updated to accomodate this. The `RefreshKeys` method now uses this to determine whether to retry refreshing keys.

2. A local implementation of the transport `CertificateAuthority` interface is now supported. It is designed to facilitate improved testing for clients using the transport package.


lziest:
LGTM

PR #609, opened 2016-05-06
Added database connection capabilities to OCSP responder

longsilver:
## OCSP responder is now backed by a CertDB instance
Now the `ocspserve` command accepts `-db-config` and `-interval` flags in order to establish a connection to a `CertDB` instance and query new responses every given interval of time.

Along with this, some issues were solved regarding to the OCSP responder, and some logs were added  to the `ocsprefresh` command.

## Further development
I noticed that the OCSP has no way to know when a CA revokes a certificate. By knowing if a certificate was revoked, the OCSP could update its responses to match the current state of the certificate.

For instance, if a new certificate is issued and its response validity is set to `2h0m0s`. In case said certificate was revoked before its validity time expired, the entity owner of the certificate will be entitled to access service or resources that it shouldn't have access to _(since its OCSP response would still be valid)_.

**Shouldn’t be added some kind of inline updates for certificate revocation to CFSSL? So this way the OCSP responder can update revoked certificates so their responses wouldn't be valid any more**

### DIFF
--- a/cli/ocsprefresh/ocsprefresh.go
+++ b/cli/ocsprefresh/ocsprefresh.go
@@ -64,6 +64,10 @@ func ocsprefreshMain(args []string, c cli.Config) error {
 
 	// Set an expiry timestamp for all certificates refreshed in this batch
 	ocspExpiry := time.Now().Add(c.Interval)
+	start := time.Now()
+
+	log.Info("OCSP responses are being updated")
+
 	for _, certRecord := range certs {
 		cert, err := helpers.ParseCertificatePEM([]byte(certRecord.PEM))
 		if err != nil {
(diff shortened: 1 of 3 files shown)


lziest:
right, our idea on OCSP responder is it should be dumb. It's CA's job to revoke a cert and put states in the DB. OCSP responder simply trusts one source of truth, and it's not its job to alter the source. And it's OK to have delay. e.g. OCSP responses have nextUpdate field, a cert could be revoked before the nextUpdate and yet related OCSP responses would still say it is valid. As long as the delay is limited to hours or even days, we are good.

longsilver:
**Understood**,

Yeah if you think about the use case where multiple certificates (issued by different `CA`s) are being verified by the same `OCSP`, it's the best to let the `CA`s handle the state of their certificates.

lziest:
golint error "/home/travis/gopath/src/github.com/cloudflare/cfssl/ocsp/responder.go:91:1: comment on exported type DBsource should be of the form "DBsource ..." (with optional leading article)"

jvermillard:
It's something I would like to use, anything I can do to help to get this merged?

PR #610, opened 2016-05-07
Fix L/ST swap in print-defaults csr.

csnook:
"cfssl print-defaults csr" swaps the "ST" and "L" fields inside "names". While the proper field naming is obvious to an American, it might not be obvious to someone from a country that does not have sub-entities called "states", particularly if they are not native English speakers. This trivial change avoids any confusion.


lziest:
LGTM

PR #613, opened 2016-05-20
remove flag definition in init()

lziest:
It causes unexpected consequences at runtime.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -113,6 +113,8 @@ func Start(cmds map[string]*Command) error {
 		flag.Usage()
 		return errors.New("undefined command")
 	}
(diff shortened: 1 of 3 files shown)


lziest:
hmmm, not quite right, now need to define flag for log.Level at every cmd package. Will reopen with them.

lziest:
@jkroll create a PR as you suggested

lziest:
I would appreciate some feedbacks on this.

kisom:
I very much like this PR; I really don't like flags getting defined in packages. LGTM.

lziest:
Now we have to find all the 3rd party cmd/cli consumer of cfssl/log and define flag for them.

PR #620, opened 2016-06-24
fix known property access bugs in UI-ng + SRI

terinjokes:
This pull request comes with two changes.

The first guards against known property access bugs, where the API's result doesn't contain a key, but we're attempting to get the value of a deep property.

The second updates the versions of bootswatch and mithril in use, forces them to connect to CDNJS over HTTPS, and adds Subresource Integrity values.

(For those playing along at work, this is SECENG-1659)


kisom:
LGTM.

PR #622, opened 2016-06-26
signer: use a stubbed CT server for tests

tam7t:
I tried running tests for `signer/local/.` but they were failing with

```
*** Test killed with quit: ran too long (10m0s).
```

Seems that the `TestCTFailure` test was [infinite looping](https://github.com/google/certificate-transparency/blob/master/go/client/logclient.go#L204) with the repeated error...

```
2016/06/26 14:03:05 [INFO] submitting poisoned precertificate to https://ct.googleapis.com/pilot
2016/06/26 14:03:05 Got Post https://ct.googleapis.com/pilot/ct/v1/add-pre-chain: stream error: stream ID 1; PROTOCOL_ERROR, backing off.
2016/06/26 14:03:05 Got Unknown, backing-off 10 seconds.
```

I'm not sure why `https://ct.googleapis.com/pilot` isn't working anymore but the test passed with another servers such as `https://ct.startssl.com`. I though it was best for the test to not interact with a 3rd party live server so switched the test to use a stubbed `httptest` server.

### DIFF
--- a/signer/local/local_test.go
+++ b/signer/local/local_test.go
@@ -8,6 +8,8 @@ import (
 	"encoding/hex"
 	"encoding/pem"
 	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
 	"reflect"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


lziest:
Thanks, LGTM

kisom:
LGTM.

PR #637, opened 2016-07-18
revoke: implement thread-safe compatibility and local file CRL

kayrus:
subj


revoke/revoke.go
```diff
@@ -142,6 +158,31 @@ func certIsRevokedCRL(cert *x509.Certificate, url string) (revoked, ok bool) {
 		}
 	}
 
+	u, err := neturl.Parse(url)
+	if err != nil {
+		log.Warningf("failed to parse CRL url: %v", err)
+		return false, false
+	}
+	if u.Scheme == "" && shouldFetchCRL {
```
kisom: Maybe this should use the `file:` scheme?

kayrus: file scheme doesn't work with `ioutil.ReadFile`. And do you really want to convert regular path to `file://` and then convert it back for `ReadFile`?

kayrus:
@kisom I also would like to remove the `github.com/cloudflare/cfssl/log` from include to allow external app to use its own logger (not quite sure how to implement that, any clue?). And make `HardFail` to be an option for functions to make it independent for multiple subapps/threads.

revoke/revoke.go
kisom: If you have a file:// URL, wouldn't the `Path` field work for the file name? RFC 5280 has a [number of options](https://tools.ietf.org/html/rfc5280#section-4.2.1.13) for CRL distribution points, and the one that seems most relevant is a URI (other options include directory names, e.g. for LDAP, or `nameRelativeToIssuer`). It seems then that the file scheme is the most appropriate. If it's a local file, wouldn't it make sense to put it in CRLSet as a URI, not a plain path?

kisom:
@kayrus I don't really have a problem with removing log except that it's useful for debugging --- if you have a good way around that, I'm all ears. I'm also fine with making `HardFail` an option. It might be useful to make a revocation configuration structure instead and add the checking functions onto that as methods, or make the functions accept that structure (e.g. crypto/x509's `VerifyOptions`).

kayrus:
@kisom added few improvements, could you please check?

revoke/revoke.go
```diff
@@ -22,14 +24,21 @@ import (
...
+	HardFail bool
+	// CRLSet associates a PKIX certificate list with the URL the CRL is
+	// fetched from.
+	CRLSet map[string]*pkix.CertificateList
+}
 
-// CRLSet associates a PKIX certificate list with the URL the CRL is
-// fetched from.
-var CRLSet = map[string]*pkix.CertificateList{}
+// NewRevokeChecker creates Revoke config structure
```
kisom: Per Go package conventions, this can just be `New`.

kayrus: Fixed

revoke/revoke.go
```diff
@@ -11,32 +11,43 @@ import (
...
+	// Logger
+	Logger *log.Logger
+}
 
-// CRLSet associates a PKIX certificate list with the URL the CRL is
-// fetched from.
-var CRLSet = map[string]*pkix.CertificateList{}
+// New creates Revoke config structure
+func New() *Revoke {
+	return &Revoke{false, map[string]*pkix.CertificateList{}, nil}
```
lziest: prefer a named initialization: {
   HardFail : false, 
...
}

revoke/revoke.go
```diff
@@ -45,6 +56,54 @@ func ldapURL(url string) bool {
 	return false
 }
 
+func (r *Revoke) logErrorf(format string, args ...interface{}) {
+	if r.Logger != nil {
+		r.Logger.Errorf(format, args...)
+	} else {
+		log.Errorf(format, args...)
```
lziest: Errorf is not defined for built-in log

revoke/revoke.go
```diff
@@ -45,6 +56,54 @@ func ldapURL(url string) bool {
...
+	}
+}
+
+func (r *Revoke) logWarningf(format string, args ...interface{}) {
+	if r.Logger != nil {
+		r.Logger.Warningf(format, args...)
+	} else {
+		log.Warningf(format, args...)
```
lziest: ditto, please refer to https://golang.org/pkg/log/

revoke/revoke.go
```diff
@@ -45,6 +56,54 @@ func ldapURL(url string) bool {
...
+	}
+}
+
+func (r *Revoke) logInfof(format string, args ...interface{}) {
+	if r.Logger != nil {
+		r.Logger.Infof(format, args...)
+	} else {
+		log.Infof(format, args...)
```
lziest: ditto.

cmd/mkbundle/mkbundle.go
```diff
@@ -24,6 +24,8 @@ import (
 // certificates are then checked for revocation.
 func worker(paths chan string, bundler chan *x509.Certificate, pool *sync.WaitGroup) {
 	defer (*pool).Done()
+	revokeChecker := revoke.New()
+	revokeChecker.Logger = log
```
lziest: bad syntax

revoke/revoke.go
```diff
@@ -123,69 +195,139 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+	shouldFetchCRL := !r.isInMemoryCRLValid(path)
+
+	u, err := neturl.Parse(path)
+	if err != nil {
+		r.logWarningf("failed to parse CRL url: %v", err)
+		return err
+	}
+
+	if u.Scheme == "" && (shouldFetchCRL || force) {
+		if _, err := os.Stat(path); err == nil {
```
lziest: not idiom usage, treat error with error != nil

kayrus: Fixed. Didn't manage the log stuff. Removed...

kayrus:
@lziest @kisom tests added

revoke/revoke.go
```diff
@@ -22,14 +25,29 @@ import (
...
-// verification to fail (a hard failure).
-var HardFail = false
+// Revoke type contains configuration for each new revoke instance
+type Revoke struct {
+	// HardFail determines whether the failure to check the revocation
+	// status of a certificate (i.e. due to network failure) causes
+	// verification to fail (a hard failure).
+	HardFail bool
+	// HardFailLck is a Mutex locker for the HardFail
+	HardFailLck sync.Mutex
```
lziest: let's merge the two locks, it is unclear why we want to lock on HardFail other than a testing artifact. Even if we do, we can use one lock.

kayrus: Done

revoke/revoke.go
```diff
@@ -123,69 +154,145 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
-// as revCheck.
-func certIsRevokedCRL(cert *x509.Certificate, url string) (revoked, ok bool) {
-	crl, ok := CRLSet[url]
+// checks whether CRL in memory is valid
+func (r *Revoke) isInMemoryCRLValid(key string) bool {
+	crl, ok := r.CRLSet[key]
 	if ok && crl == nil {
 		ok = false
-		delete(CRLSet, url)
+		r.CRLSetLck.Lock()
```
lziest: this actually unsafe operation.  One thread will do "r.CRLSet[key]" read and the other will do delete. Only locking on delete doesn't prevent the read. I suggest we don't do locking for now and focusing on functionality.

revoke/revoke_test.go
```diff
@@ -112,7 +146,11 @@ lBlGGSW4gNfL1IYoakRwJiNiqZ+Gb7+6kHDSVneFeO/qJakXzlByjAA6quPbYzSf
 +AZxAeKCINT+b72x
 -----END CERTIFICATE-----`)
 
-var goodCert = mustParse(goodComodoCA)
+var (
+	goodCert      = mustParse(goodComodoCA)
+	revokeChecker = New(false)
```
lziest: I suggest we do testing on a fresh revokeChecker on each individual test.

revoke/revoke_test.go
```diff
@@ -129,35 +167,65 @@ func mustParse(pemData string) *x509.Certificate {
 	return cert
 }
 
+func insertLocalCRL(revoke *Revoke) {
+	cert, err := x509.ParseCRL([]byte(localCRLList))
+	if err != nil {
+		panic(err.Error())
+	}
+	revoke.CRLSetLck.Lock()
```
lziest: bad syntax

revoke/revoke.go
kayrus: thread safe is quite mandatory for apps which will use this library. what do you suggest? use [RWMutex](https://golang.org/pkg/sync/#RWMutex)?

transport/client.go
```diff
@@ -311,7 +315,7 @@ func Dial(address string, tr *Transport) (*tls.Conn, error) {
 
 	for _, chain := range state.VerifiedChains {
 		for _, cert := range chain {
-			revoked, ok := revoke.VerifyCertificate(cert)
```
lziest: My idea is we still keep the original methods around: declare an unexported package variable `defaultChecker` with hardfail=true and a empty map, run those package level methods through this defaultChecker. So you don't need to update everywhere revoke is used.

revoke/revoke.go
lziest: RWMutex may improve efficiency, but placing the lock around map operation needs extra efforts. So far what I see is complicating this PR, which I thought the main focus is to run CRL check against a CRL file.

kayrus: This is already not a main focus. Main focus is to make this lib to be more suitable for external apps.

revoke/revoke_test.go
kayrus: what do you suggest here?

revoke/revoke_test.go
kayrus: done

lziest:
@kayrus thanks for your contribution. I am all for thread safety. I think thread safety through mutex can be a separate PR, which makes it easier to reason about.

transport/client.go
kayrus: Sorry, didn't get where exactly should we declare `defaultChecker`. Can you post an example?

kayrus:
@lziest will split when all notes will be resolved.

transport/client.go
lziest: one example, flag.Int is actually "flag.CommandLine.Int"
https://golang.org/src/flag/flag.go?s=17993:18044#L584
All methods are declared on FlagSets, and CommandLine is the default FlagSet.

kayrus: As I understood, this example can handle only one var per app. What if app serves several TLS enabled listeners with different certs and CRL sets?

lziest: I don't understand your question. 1. if app wants multiple checker, app can New(). 2. listeners can share checker.

kayrus: How can they share checker if there could be different TLS confs? I.e. 444 uses /path/crl1.crl and hardfail=true (for exmaple with self signed certs ), 443 uses InCert URLs (ocsp, crl) and hardfail=false with a regular CA.

kayrus:
@lziest I hid public options and added wrappers instead.

transport/client.go
lziest: I feel there are several issues we are trying to solve here. I want to clarify with you together to get our PR defined by a feature request. 
Let me digress a little bit first.
1. self-signed certs probably never needs revocation or CRL, since it is easier to just throw those away from trust stores.
2. With file based CRL, we may simplify the CRL distribution (we don't need a server anyway), this will help a customized internal PKI.
3. Public CA probably don't need this file-based CRL feature.

We want to do 2 as an option right? and we don't want to complicate 3. 
Now back to this change. What I saw here at this line is that tr.RevokeChecker is not different from a defaultChecker which can be defined universally. Maybe we will want a customized checker later, but not now, at least from this change I can't justify it.

kayrus: 1) no. please take into consideration kubernetes, etcd which use cert auth. It is much easier to update a local CRL list rather than reconfigure whole cluster.
2) this resolves the 1st problem
3) well... at least we can automatically fetch public CRL list into fs and use it when there are some connectivity problems with CRL server.

Now I understand what you want me to implement. Will do soon.

kayrus: Done

revoke/revoke.go
```diff
@@ -22,14 +24,86 @@ import (
...
+func New(hardfail bool) *Revoke {
+	return &Revoke{
+		localCRL: "",
+		hardFail: hardfail,
+		crlSet:   map[string]*pkix.CertificateList{},
+	}
+}
+
+// SetLocalCRL sets localCRL path into the Revoke struct
+func (r *Revoke) SetLocalCRL(localCRLpath string) error {
```
kisom: I think this should go into the `New` function; it's probably not something that should be changed often (ever?) and therefore probably doesn't need its own function.

revoke/revoke.go
```diff
@@ -22,14 +24,86 @@ import (
...
+	localCRL string
+	// HardFail determines whether the failure to check the revocation
+	// status of a certificate (i.e. due to network failure) causes
+	// verification to fail (a hard failure).
+	hardFail bool
+	// crlSet associates a PKIX certificate list with the URL the CRL is
+	// fetched from.
+	crlSet map[string]*pkix.CertificateList
+	// lck is a Mutex locker for the struct
+	lck sync.Mutex
```
kisom: It's somewhat picky, but can this be `lock`? It's one extra character and it's more readable.

revoke/revoke.go
kayrus: Main point was to allow structure update when app has been triggered by sighup

revoke/revoke.go
lziest: I agree with @kisom, we need to be RFC compatible. We need to use URI here.

revoke/revoke.go
kayrus: done

revoke/revoke.go
kayrus: @lziest RFC explains the certificate extension, and there is nothing about the local file. In our situation we set CRL path from the app and it doesn't relate to this RFC.

transport/client.go
```diff
@@ -71,6 +71,9 @@ type Transport struct {
 	// as part of AutoUpdate.
 	Backoff *core.Backoff
 
+	// RevokeChecker is used to configure revoke package.
+	RevokeChecker *revoke.Revoke
```
lziest: remove this for now.

lziest:
@kayrus please make sure you PR pass tests first. It's very time consuming to nitpick typos.

revoke/revoke.go
```diff
@@ -22,14 +24,86 @@ import (
...
+// SetHardFail allows to dynamically set hardfail bool into the
+// default var struct
+func SetHardFail(hardfail bool) {
+	defaultChecker.lock.Lock()
+	defaultChecker.hardFail = hardfail
+	defaultChecker.lock.Unlock()
+}
+
+// HardFail returns hardfail bool from the default var struct
+func HardFail() bool {
```
lziest: Use IsHardFail()

revoke/revoke.go
```diff
@@ -22,14 +24,86 @@ import (
...
+// SetHardFail allows to dynamically set hardfail bool into the
+// Revoke struct
+func (r *Revoke) SetHardFail(hardfail bool) {
+	r.lock.Lock()
+	r.hardFail = hardfail
+	r.lock.Unlock()
+}
+
+// HardFail returns hardfail bool from the Revoke struct
+func (r *Revoke) HardFail() bool {
```
lziest: ditto

lziest:
I found those lock() unlock() problematic, please keep this PR lock free first.

kayrus:
@lziest It would be much easier to test if you'd merge these PRs before: #646 #647 
Anyway I've fixed types. Please let me know whether I have to squash commits.
Thanks!

revoke/revoke.go
kayrus: @lziest @kisom added `file:///` scheme support

kayrus:
> I found those lock() unlock() problematic, please keep this PR lock free first.
Which problems did you face? As for me it doesn't make sense to merge this PR without mutexes. @heyitsanthony @xiang90, could you please take a look on this PR?

xiang90:
@kayrus @lziest Is the revoke struct expected to be accessed by multiple threads (go routines) concurrently? If not, we should remove the lock. If yes, we have to keep it. Also we should move the lock above the fields that it protects.

kayrus:
@xiang90 etcd and k8s-apiserver will access this struct on every user's https query. so I guess - yes.

xiang90:
@kayrus Then we should keep the lock.

lziest:
My preference is that a separate commit adding mutex operations. I feel the critical zone is not well defined in this PR. I think the whole structure is critical, so for each r/w method, the idiomatic way is to lock the method receiver (usually a pointer) at the very beginning, and do "defer p.Unlock()" immediately. The "only-lock-around-map-write-access" is proven to be problematic.

revoke/revoke.go
```diff
@@ -22,14 +24,85 @@ import (
...
+	}
+
+	return fmt.Errorf("Path is not valid: %s", localCRLpath)
+}
+
+// SetHardFail allows to dynamically set hardfail bool into the
+// default var struct
+func SetHardFail(hardfail bool) {
+	defaultChecker.lock.Lock()
+	defaultChecker.hardFail = hardfail
```
lziest: simply it by return defaultChecker.IsHardFail()

revoke/revoke.go
```diff
@@ -287,15 +449,15 @@ func sendOCSPRequest(server string, req []byte, issuer *x509.Certificate) (*ocsp
 
 	switch {
 	case bytes.Equal(body, ocsp.UnauthorizedErrorResponse):
-		return nil, errors.New("OSCP unauthorized")
+		return nil, fmt.Errorf("OSCP unauthorized")
```
lziest: can we undo those "errors.New" -> "fmt.Errorf" changes?

revoke/revoke.go
```diff
@@ -123,69 +209,145 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+
+	return true
+}
+
 // VerifyCertificate ensures that the certificate passed in hasn't
 // expired and checks the CRL for the server.
+// Comparing to the next public method, this function uses
+// defaultChecker variable.
 func VerifyCertificate(cert *x509.Certificate) (revoked, ok bool) {
-	if !time.Now().Before(cert.NotAfter) {
```
lziest: simplify this by defaultChecker.VerifyCertificate(cert)

lziest:
@kayrus the mutex operation you put in this PR is not making it thread safe. for example, it only locks map write/delete operation, while map read can race with map delete. It doesn't make sense for me to merge a PR with problematic thread safety code.

revoke/revoke.go
kayrus: `errors.New` doesn't support formatting. `fmt.Errorf` - does. Why do we need to import both `fmt` and `errors` packages?

revoke/revoke.go
kayrus: done

revoke/revoke.go
kayrus: done

kayrus:
@lziest i'd like to work on thread-safe stuff within this PR. Just fixed problem strings.

lziest:
Thanks, I will give a detailed review over the weekend.

kayrus:
@lziest Thank you for your patience and cooperation :)

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+		log.Infof("Certificate isn't valid until %s", cert.NotBefore)
+		return false
+	}
+
+	return true
+}
+
 // VerifyCertificate ensures that the certificate passed in hasn't
 // expired and checks the CRL for the server.
+// Comparing to the next public method, this function uses
```
lziest: I thing godoc has its own way of ordering and formatting documents, so this comment is unnecessary and confusing, should be removed.

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+		log.Warningf("%v", err)
+		return false, false
 	}
 
-	for _, revoked := range crl.TBSCertList.RevokedCertificates {
+	defer r.lock.Unlock()
+	r.lock.Lock()
+	for _, revoked := range r.crlSet[url].TBSCertList.RevokedCertificates {
 		if cert.SerialNumber.Cmp(revoked.SerialNumber) == 0 {
-			log.Info("Serial number match: intermediate is revoked.")
```
lziest: why remove this line?

revoke/revoke.go
```diff
@@ -60,32 +133,49 @@ func ldapURL(url string) bool {
...
 		if ldapURL(url) {
 			log.Infof("skipping LDAP CRL: %s", url)
 			continue
 		}
 
-		if revoked, ok := certIsRevokedCRL(cert, url); !ok {
+		if revoked, ok := r.certIsRevokedCRL(cert, url, true); !ok {
 			log.Warning("error checking revocation via CRL")
-			if HardFail {
+			if r.hardFail {
```
lziest: mix use of r.hardFail and hardFail

revoke/revoke.go
```diff
@@ -60,32 +133,49 @@ func ldapURL(url string) bool {
 //
 //  true, false:  failure to check revocation status causes
 //                  verification to fail
-func revCheck(cert *x509.Certificate) (revoked, ok bool) {
+func (r *Revoke) revCheck(cert *x509.Certificate) (revoked, ok bool) {
+	r.lock.Lock()
+	localCRL := r.localCRL
+	hardFail := r.hardFail
+	r.lock.Unlock()
```
lziest: I don't understand this unlock.

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
 		ok = false
-		delete(CRLSet, url)
+		delete(r.crlSet, key)
 	}
 
-	var shouldFetchCRL = true
-	if ok {
-		if !crl.HasExpired(time.Now()) {
-			shouldFetchCRL = false
+	if ok && !crl.HasExpired(time.Now()) {
```
lziest: bad condition, crl may be nil here, better rewrite this,  if ok { if crl == nil {} else if crl.HasExpire() {}}

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+		r.crlSet[url] = crl
+		r.lock.Unlock()
+	}
+
+	return nil
+}
+
+// check a cert against a specific CRL. Returns the same bool pair
+// as revCheck. If remote is false - will assume that url is a file.
+func (r *Revoke) certIsRevokedCRL(cert *x509.Certificate, url string, remote bool) (revoked, ok bool) {
```
lziest: this url and remote combination is weird. It's where we use "file://" and "http(s)://" to distinguish.

revoke/revoke.go
```diff
@@ -60,32 +133,49 @@ func ldapURL(url string) bool {
...
+		if revoked, ok := r.certIsRevokedCRL(cert, url, true); !ok {
 			log.Warning("error checking revocation via CRL")
-			if HardFail {
+			if r.hardFail {
 				return true, false
 			}
 			return false, false
 		} else if revoked {
-			log.Info("certificate is revoked via CRL")
+			log.Info("certificate is revoked by '%s' CRL (CN=%s, Serial: %s)", url, cert.Subject.CommonName, cert.SerialNumber)
```
lziest: Infof

revoke/revoke.go
```diff
@@ -60,32 +133,49 @@ func ldapURL(url string) bool {
...
+		if revoked, ok := certIsRevokedOCSP(cert, hardFail); !ok {
 			log.Warning("error checking revocation via OCSP")
-			if HardFail {
+			if hardFail {
 				return true, false
 			}
 			return false, false
 		} else if revoked {
-			log.Info("certificate is revoked via OCSP")
+			log.Info("certificate is revoked by '%s' OCSP (CN=%s, Serial: %s)", url, cert.Subject.CommonName, cert.SerialNumber)
```
lziest: infof

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+		if err != nil {
+			return fmt.Errorf("failed to read local CRL path: %v", err)
+		}
+		crl, err := x509.ParseCRL(tmp)
+		if err != nil {
+			return fmt.Errorf("failed to parse local CRL file: %v", err)
 		}
+
+		r.lock.Lock()
+		r.crlSet[newLocalCRL] = crl
```
lziest: given we don't want to cache nil crl, better check crl != nil

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+		if r.localCRL != newLocalCRL {
+			if _, ok := r.crlSet[r.localCRL]; ok {
```
lziest: this if check is redundant, delete is no-op for absent key.

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+// FetchRemoteCRL fetches remote CRL into internal map,
+// force overwrites previously read CRL
+func (r *Revoke) FetchRemoteCRL(url string, cert *x509.Certificate, force bool) error {
+	shouldFetchCRL := !r.isInMemoryCRLValid(url)
+
 	issuer := getIssuer(cert)
 
-	if shouldFetchCRL {
+	if force || shouldFetchCRL {
 		var err error
```
lziest: this line can be removed.

revoke/revoke.go
lziest: right now err != nil  ==> crl != nil, but I don't want to assume that is always true.

revoke/revoke_test.go
```diff
@@ -129,35 +166,68 @@ func mustParse(pemData string) *x509.Certificate {
 	return cert
 }
 
+func insertLocalCRL(revoke *Revoke) {
+	cert, err := x509.ParseCRL([]byte(localCRLList))
```
lziest: use `crl` instead of `cert` please.

revoke/revoke_test.go
```diff
@@ -129,35 +166,68 @@ func mustParse(pemData string) *x509.Certificate {
...
 func TestRevoked(t *testing.T) {
-	if revoked, ok := VerifyCertificate(revokedCert); !ok {
+	rc := New(false)
+	if revoked, ok := rc.VerifyCertificate(revokedCert); !ok {
+		fmt.Fprintf(os.Stderr, "Warning: soft fail checking revocation")
+	} else if !revoked {
+		t.Fatalf("revoked certificate should have been marked as revoked")
+	}
+
+	insertLocalCRL(rc)
```
lziest: I see insertLocalCRL makes no difference. Also, this insertLocalCRL is a short cut. It hardly tests any file fetch logic you put in.

And we can now test directly with functions defined directly (and test defaultChecker's thread safety). The reason I suggested using new checker at every test is that I saw you spent too much time writing the locking logic
at that time is not yet proper. With a fresh checker with every test, we can focus on code logic first.

revoke/revoke_test.go
```diff
@@ -129,35 +166,68 @@ func mustParse(pemData string) *x509.Certificate {
...
 func TestExpired(t *testing.T) {
-	if revoked, ok := VerifyCertificate(expiredCert); !ok {
+	rc := New(false)
+	if revoked, ok := rc.VerifyCertificate(expiredCert); !ok {
+		fmt.Fprintf(os.Stderr, "Warning: soft fail checking revocation")
+	} else if !revoked {
+		t.Fatalf("expired certificate should have been marked as revoked")
+	}
+
+	insertLocalCRL(rc)
```
lziest: this insertLocalCRL and test again isn't giving any more line coverage, restore it to the original state.

revoke/revoke_test.go
```diff
@@ -129,35 +166,68 @@ func mustParse(pemData string) *x509.Certificate {
...
 func TestGood(t *testing.T) {
-	if revoked, ok := VerifyCertificate(goodCert); !ok {
+	rc := New(false)
+	if revoked, ok := rc.VerifyCertificate(goodCert); !ok {
 		fmt.Fprintf(os.Stderr, "Warning: soft fail checking revocation")
 	} else if revoked {
 		t.Fatalf("good certificate should not have been marked as revoked")
 	}
 
+	insertLocalCRL(rc)
```
lziest: same, please do mindful tests.

revoke/revoke_test.go
lziest: I don't see a real test hitting the newly added crl and caused revoked == true.

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+
+	return nil
+}
+
+// check a cert against a specific CRL. Returns the same bool pair
+// as revCheck. If remote is false - will assume that url is a file.
+func (r *Revoke) certIsRevokedCRL(cert *x509.Certificate, url string, remote bool) (revoked, ok bool) {
+	var err error
+	if remote {
+		err = r.FetchRemoteCRL(url, cert, false)
```
lziest: why don't we return the fetched CRL here? So we don't need to lock() unlock below.

revoke/revoke.go
```diff
@@ -22,14 +24,85 @@ import (
...
+		hardFail: hardfail,
+		crlSet:   map[string]*pkix.CertificateList{},
+	}
+}
+
+// SetLocalCRL sets localCRL path into the Revoke struct
+func (r *Revoke) SetLocalCRL(localCRLpath string) error {
+	if localCRLpath == "" {
+		r.lock.Lock()
+		if _, ok := r.crlSet[r.localCRL]; ok {
```
lziest: unnecessary check, just delete

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+			return fmt.Errorf("failed to read local CRL path: %v", err)
+		}
+		crl, err := x509.ParseCRL(tmp)
+		if err != nil {
+			return fmt.Errorf("failed to parse local CRL file: %v", err)
 		}
+
+		r.lock.Lock()
+		r.crlSet[newLocalCRL] = crl
+		if r.localCRL != newLocalCRL {
```
lziest: I found this whole block redundant. I prefer this function doesn't take the path as parameter, and just read from r.localCRL. Fetch should do the job of fetching, and only updating the in-memory cache.

revoke/revoke.go
```diff
@@ -123,69 +213,141 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+		}
+		r.lock.Unlock()
 	}
 
+	return nil
+}
+
+// FetchRemoteCRL fetches remote CRL into internal map,
+// force overwrites previously read CRL
+func (r *Revoke) FetchRemoteCRL(url string, cert *x509.Certificate, force bool) error {
```
lziest: cert is not being used at all, consider passing in issuer directly.

revoke/revoke.go
```diff
@@ -22,14 +24,85 @@ import (
...
+		localCRL: "",
+		hardFail: hardfail,
+		crlSet:   map[string]*pkix.CertificateList{},
+	}
+}
+
+// SetLocalCRL sets localCRL path into the Revoke struct
+func (r *Revoke) SetLocalCRL(localCRLpath string) error {
+	if localCRLpath == "" {
+		r.lock.Lock()
```
lziest: The practice of lock/unlock tightly around a read/write operation isn't going to buy something. In fact this introduces the exact race condition that deletion is reaching lock faster than write, so a cache deletion after a cache write may have chances finishing before write finishes, making deletion an no-op. My suggestion is as follows:
1. lock at the very beginning of a read/write/deletion function. defer unlock. Let's not worry about efficiency right now, correctness must be maintained as the first priority.
2. the CRL fetch works as follows: 1. check hashmap, return hit. 2. in case of miss, fetch from remote and update hashmap
3. revocation check actually happens after crl fetch, so revocation check is lock free.

revoke/revoke.go
kayrus: done

revoke/revoke.go
kayrus: There is another lock inside the `certIsRevokedCRL` and it causes double lock, i.e. block the app forever.

revoke/revoke.go
kayrus: fixed

revoke/revoke.go
kayrus: fixed

revoke/revoke.go
kayrus: fixed

revoke/revoke.go
kayrus: fixed

revoke/revoke.go
kayrus: fixed

revoke/revoke.go
kayrus: Tried to get rid of log* and forgot to restore. Fixed.

revoke/revoke.go
kayrus: fixed

revoke/revoke_test.go
kayrus: fixed

kayrus:
Will push commits fixing the issues soon.

revoke/revoke.go
kayrus: I think this would be better:

```go
        if ok && crl == nil {
                ok = false
                delete(r.crlSet, key)
        } else if crl == nil {
                delete(r.crlSet, key)
                return false
        }

        if ok && !crl.HasExpired(time.Now()) {
                return true
        }
```

revoke/revoke.go
kayrus: Ok, fixed.

revoke/revoke.go
kayrus: Let's imagine we set invalid localCRL path. According to your comment we have to set invalid `localCRL` into protected struct, then check it, and if it is not ok - we have to revert changes back. Current implementation works well in this scenario.

revoke/revoke.go
kayrus: done

revoke/revoke.go
kayrus: Don't want to add additional checks for the url/path. So we just use bool. Renamed `url` to the `crlPath`.

revoke/revoke.go
kayrus: UPD: Not fixed. Because we have to check path before update the value. And if I'd return CRL here, I have to implement additional checks in other functions, which will make the code more complicated (take a look at `SetLocalCRL`).

revoke/revoke.go
kayrus: Removed

revoke/revoke_test.go
kayrus: fixed

revoke/revoke_test.go
kayrus: fixed

revoke/revoke_test.go
kayrus: fixed

revoke/revoke.go
kayrus: Yep, it's kinda non-trivial to make it thread safe and also add possibility to reload configuration. Will try to find some time to reimplement this part again.

revoke/revoke.go
lziest: so we don't lock in certIsRevokedCRL. lock/unlock makes sure the update happens in atomic transaction, simply placing lock/unlock around read/write is not implementing thread safety. I do suggest you study about ACID principle and multithread programming.

revoke/revoke.go
lziest: User sets invalid path, that's user's fault, user makes this checker into a bad state. I don't understand why you want to revert the change. Trying to be clever/robust against human error is dumb, it makes your code logic hard to be reasoned about. Please, just don't do it, because I won't approve it.

kayrus:
@lziest please have a look once again. i've reworked lock logic. it is not optimized for remote URL fetching, but it is completely safe now.

kayrus:
@lziest do you have any news on this PR? I'd like CRL stuff to be merged into k8s v1.4. This PR is a blocker.

lziest:
sorry, I don't have time to review it until next week.

kayrus:
@lziest np. thanks for letting know.

revoke/revoke.go
```diff
@@ -123,48 +263,90 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+		return nil, fmt.Errorf("failed to read local CRL path: %v", err)
+	}
+
+	crl, err := x509.ParseCRL(tmp)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse local CRL file: %v", err)
+	}
+
+	if crl == nil {
+		return nil, fmt.Errorf("Local CRL is nil")
```
lziest: why return error with nil CRL? nil CRL is fine, right? It means empty list.

revoke/revoke.go
```diff
@@ -123,48 +263,90 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
+	}
+
+	if crl.HasExpired(time.Now()) {
```
lziest: if you really worry about crl == nil, you can check it here.

revoke/revoke.go
```diff
@@ -22,14 +24,139 @@ import (
...
+func (r *Revoke) setLocalCRL(localCRLpath string) error {
+	if localCRLpath == "" {
+		r.unsetLocalCRL()
+		return nil
+	}
+
+	if u, err := neturl.Parse(localCRLpath); err != nil {
+		r.unsetLocalCRL()
+		return err
+	} else if u.Scheme == "" {
```
lziest: style issue:
don't do if err != nil {} else if {}, "if err != nil" should be standalone, especially when error will be returned immediately.  It makes code logic hard to follow.

revoke/revoke.go
```diff
@@ -22,14 +24,139 @@ import (
 	"github.com/cloudflare/cfssl/log"
 )
 
-// HardFail determines whether the failure to check the revocation
-// status of a certificate (i.e. due to network failure) causes
-// verification to fail (a hard failure).
-var HardFail = false
+type localCRL struct {
+	// embedded Mutex locker for the localCRL
+	sync.RWMutex
```
lziest: Putting a mutex over each individual component is not the right way to write thread-safe. It's not well justified. it increase deadlock risk.

revoke/revoke.go
```diff
@@ -22,14 +24,139 @@ import (
...
+	path string
+	crl  *pkix.CertificateList
+}
+
+type crlSet struct {
+	// embedded Mutex locker for the crlSet
+	sync.RWMutex
```
lziest: don't

revoke/revoke.go
```diff
@@ -22,14 +24,139 @@ import (
...
+	// crlSet associates a PKIX certificate list with the URL the CRL is
+	// fetched from.
+	set map[string]*pkix.CertificateList
+}
+
+type hardFail struct {
+	// embedded Mutex locker for the hardFail
+	sync.RWMutex
```
lziest: don't

revoke/revoke.go
```diff
@@ -123,48 +263,90 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
...
-func certIsRevokedCRL(cert *x509.Certificate, url string) (revoked, ok bool) {
-	crl, ok := CRLSet[url]
+// checks whether CRL in memory is valid
+func (r *Revoke) isInMemoryCRLValid(key string) bool {
+	crl, ok := r.crl.set[key]
 	if ok && crl == nil {
 		ok = false
-		delete(CRLSet, url)
+		delete(r.crl.set, key)
+	} else if crl == nil {
```
lziest: this 'if else' block is very confusing.

if ok == false and crl == nil, it means hashmap miss, and we do delete?

revoke/revoke_test.go
```diff
@@ -183,23 +216,35 @@ func TestCRLFetchError(t *testing.T) {
...
+	}
+
+	SetHardFail(true)
+	defaultChecker.localCRL.path = ""
 	if revoked, ok := VerifyCertificate(ldapCert); ok || !revoked {
 		t.Fatalf("Fetching error not encountered, hardfail not registered")
 	}
-	HardFail = false
+
+	defaultChecker.localCRL.path = "InvalidPath"
```
lziest: as I understand, this doesn't trigger local CRL load, so it's a bogus test.

revoke/revoke_test.go
```diff
@@ -219,6 +263,39 @@ func TestRemoteFetchError(t *testing.T) {
 
 }
 
+func TestLocalCRL(t *testing.T) {
+	crl, err := x509.ParseCRL([]byte(testCRL))
+	if err != nil {
+		panic(err.Error())
+	}
+	defaultChecker.localCRL.path = "/RootSignPartners.crl"
```
lziest: again this path is a invalid path, but because crl is directly assigned, the test below works, this is another bad example of testing.

lziest:
Let's start fresh. I will close this, and please re-submit with a separate PR.

revoke/revoke.go
kayrus: Do you mean locking the whole structure just to read the `hardFail` value is ok? If yes, could you provide a proper example?

revoke/revoke.go
kayrus: ParseCRL returns ParseDERCRL (https://golang.org/src/crypto/x509/x509.go?s=51887:51948#L1647). ParseDERCRL returns crl=nil only when there is an error. I don't know whether new golang releases would return something else. That is why I added extra check. If you think I have to remove this check, so ok.

PR #640, opened 2016-07-18
update transport.core.Backoff, sync.Mutex to be instantiated immediat…

lziest:
…ely so Backoff is race free

@kisom I am not 100% sure about this change. Please review.


lziest:
But I want to use it in gokeyless and it won't create a r/w race with this new Backoff

kisom:
The current version of back off doesn't use a mutex. I'll update the transport package version.

PR #641, opened 2016-07-18
Update backoff code.

kisom:
This imports the changes from another codebase that cause `Backoffs` to
no longer be locked.

This should address the same issues in #640.

### DIFF
--- a/transport/core/backoff.go
+++ b/transport/core/backoff.go
@@ -1,13 +1,33 @@
 package core
 
-// http://www.awsarchitectureblog.com/2015/03/backoff.html
+// Package backoff contains an implementation of an intelligent backoff
+// strategy. It is based on the approach in the AWS architecture blog
+// article titled "Exponential Backoff And Jitter", which is found at
+// http://www.awsarchitectureblog.com/2015/03/backoff.html.
+//
+// Essentially, the backoff has an interval `time.Duration`; the nth
+// call to backoff will return a `time.Duration` that is 2^n *
+// interval. If jitter is enabled (which is the default behaviour),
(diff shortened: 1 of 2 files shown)


transport/core/backoff_test.go
```diff
@@ -0,0 +1,177 @@
...
+	duration := b.Duration()
+	fmt.Println(duration)
+
+	// try 3, below decay
+	time.Sleep(duration)
+	duration = b.Duration()
+	fmt.Println(duration)
+
+	// try 4, resets
+	time.Sleep(duration + decay)
```
lziest: so ideally try 5 will trigger reset right? (duration + decay) is somewhat artificial for triggering, right?

transport/core/backoff.go
```diff
@@ -18,73 +38,161 @@ var DefaultInterval = 5 * time.Minute
...
+		return
+	}
+
+	if b.lastTry.IsZero() {
+		b.lastTry = time.Now()
+		return
+	}
+
+	lastDuration := b.duration(b.n - 1)
+	decayed := time.Since(b.lastTry) > lastDuration+b.decay
```
lziest: do you think it will be simpler with time.Since(b.lastTry) > b.decay

transport/core/backoff_test.go
kisom: Yes, and yes.

transport/core/backoff.go
kisom: This is discussed [in the paper](http://www.awsarchitectureblog.com/2015/03/backoff.html), if I recall correctly.

lziest:
LGTM. it's just small gotchas.

PR #644, opened 2016-07-19
Fix error where whitelist does not match

clly:
I was working on a config with a name_whitelist and got a bunch of invalid policy errors. It was only when I ended up removing the whitelist that I realized what the error was. I found where the error was originally thrown and added a new error code and message specifically for whitelist errors..


kisom:
LGTM, can you update the docs with this error code (e.g. `doc/errorcode.txt`)?

clly:
Yes. That commit should be added now

lziest:
LGTM

PR #646, opened 2016-07-20
test: allow to run tests on the machine without proper go env

kayrus:
Now all you need are only go and gofmt binaries.

Before `test.sh` failed with the following msg:

```
$ ./test.sh 
ls: cannot access /src/github.com/cloudflare/cfssl: No such file or directory
```

And even with the GOPATH:

```
$ ./test.sh 
api          cli     Dockerfile          initca     scan       ubiquity
auth         cmd     Dockerfile.build    LICENSE    script     vendor
BUILDING.md  config  Dockerfile.minimal  log        selfsign   whitelist
bundler      crl     errors              multiroot  signer
certdb       crypto  gopath              ocsp       testdata
certinfo     csr     helpers             README.md  test.sh
CHANGELOG    doc     info                revoke     transport
can't load package: package _/home/user/git/cfssl/api: cannot find package "_/home/user/git/cfssl/api" in any of:
        /home/user/go/src/_/home/user/git/cfssl/api (from $GOROOT)
        /home/user/git/cfssl/gopath/src/_/home/user/git/cfssl/api (from $GOPATH)
can't load package: package _/home/user/git/cfssl/api/bundle: cannot find package "_/home/user/git/cfssl/api/bundle" in any of:
        /home/user/go/src/_/home/user/git/cfssl/api/bundle (from $GOROOT)
        /home/user/git/cfssl/gopath/src/_/home/user/git/cfssl/api/bundle (from $GOPATH)
```

PR #647, opened 2016-07-21
build: added build script which works without docker

kayrus:
didn't add `gopath` into `.gitignore` becasue it was already added in #646


kisom:
LGTM, thanks!

Issue #636, opened 2016-07-18
cfssl ocsprefresh spit database error


jvermillard:
I'm using the mysql backend (didn't tested with the PG one)

➜  cfssl ocsprefresh -loglevel=0 -ca-key=ca-key.pem -ca=ca.pem -db-config=../dbconf/db.json -config=./server-config.json -responder=ocsp/ocsp.pem -responder-key=ocsp/ocsp.key
2016/07/18 14:50:58 [DEBUG] loading configuration file from ./server-config.json
2016/07/18 14:50:58 [DEBUG] parse expiry in profile
2016/07/18 14:50:58 [DEBUG] expiry is valid
2016/07/18 14:50:58 [DEBUG] parse expiry in profile
2016/07/18 14:50:58 [DEBUG] expiry is valid
2016/07/18 14:50:58 [DEBUG] validating configuration
2016/07/18 14:50:58 [DEBUG] validate local profile
2016/07/18 14:50:58 [DEBUG] profile is valid
2016/07/18 14:50:58 [DEBUG] validate local profile
2016/07/18 14:50:58 [DEBUG] profile is valid
2016/07/18 14:50:58 [DEBUG] configuration ok
2016/07/18 14:50:58 [DEBUG] Loading issuer cert: ca.pem
2016/07/18 14:50:58 [DEBUG] Loading responder cert: ocsp/ocsp.pem
2016/07/18 14:50:58 [DEBUG] Loading responder key: ocsp/ocsp.key
2016/07/18 14:50:58 [DEBUG] loading db configuration file from ../dbconf/db.json
{"code":11000,"message":"sql: Scan error on column index 2: unsupported Scan, storing driver.Value type []uint8 into type *time.Time"}

I'm not missing migrations:
➜  new goose -path $GOPATH/src/github.com/cloudflare/cfssl/certdb/mysql up
goose: no migrations to run. current version: 1

jvermillard:
works with the pg based backend, so probably a mysql only issue

lziest:
maybe the mysql migration script is not up to date with the schema we are using. Also, it may be helpful to note down the mysql version here.

forthguy:
Do you have "?parseTime=true" on your data source URI?  That cleared up the error for me.  (It's in the test configs and the goose yaml, but not in the certdb README.)

lziest:
if that's the case, PR is welcomed.

jvermillard:
[...]

PR #659, opened 2016-08-25
Added "parseTime=true" to query string in example MySQL db-config fil…

forthguy:
Adds "parseTime=true" query string parameter to example MySQL db-config file, for reasons described here: https://github.com/go-sql-driver/mysql#timetime-support (and as already done in the mysql/dbconf.yml and testdb.go) #636

### DIFF
--- a/certdb/README.md
+++ b/certdb/README.md
@@ -72,4 +72,4 @@ or
  
 or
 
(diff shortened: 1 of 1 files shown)


lziest:
lgtm

PR #665, opened 2016-09-19
Note that as of Go 1.7, self-signing omits AKI.

### LINKED ISSUE #664: AuthorityKeyID is no longer being set in Go 1.7 for self-signed certificates
I upgraded to 1.7 and one of my tests started failing. I tracked it down to this change in Go 1.7.

I am not sure if it needs to be handled in any way in cfssl, or maybe add documentation around it.

https://go.googlesource.com/go/+/b623b71509b2d24df915d5bc68602e1c6edf38ca

kisom:
Closes #664.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,9 @@ a JSON file with the type:
 }
 ```
 
+**N.B.** As of Go 1.7, self-signed certificates will not include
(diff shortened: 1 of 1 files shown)


lziest:
LGTM

Issue #608, opened 2016-05-05
Bundle does not work for client certificates


fhossain:
I could not get bundle to work for client certificates. Running -loglevel 0 I can see the following error -
[DEBUG] verification failed: x509: certificate specifies an incompatible key usage
The problem is that golangs strict extended key usage validation which requires the full chain to allow key usage. Similar problem recommends that Root and Intermediate CA certificates should be created with key usage of "any". However, I could not get initca or sign for intermediate certificates to accept key usage of any.
http://grokbase.com/t/gg/golang-nuts/1412vfsyjg/go-nuts-tls-clientauth-requires-extkeyusageclientauth-through-whole-certificate-chain
It would be great if this was working for client certificates. Bundle works fine for intermediate certificates and servers certificates. My current workaround is to call "info" on the issuer. Then use the issuer certificate to get a bundle. Then manually prepend client certificate to the intermediate CA bundle.

lziest:
should be fixed.

PR #666, opened 2016-09-20
Add support for bundling certs with Client Auth Key Usage

### LINKED ISSUE #588: API bundle endpoint can't be used for bundling client certificates
The key usages used for verification within the bundler package seem to be hardcoded to a limited subset of those possible, which doesn't include `clientAuth`: https://github.com/cloudflare/cfssl/blob/92f037e39eb103fb30f9151be40d9ed267fc4ae2/bundler/bundler.go#L158

jacobhaven:
Should fix #588 and #608

r? @kisom @lziest @grittygrease


lziest:
add a specific test on this?

jacobhaven:
@lziest My main concern is that it requires generating Root, Intermediate and Client Auth Leaf Certificates and the tests will fail when they expire.

lziest:
I would write a script using cfssl to automate that, then we don't need to worry too much.

jacobhaven:
@lziest *Test certs along with script to generate them added*

lziest:
lgtm

Issue #667, opened 2016-09-20
Add optional "bundle" argument to /newcert API endpoint


jacobhaven:
It would be useful for `/newcert` to optionally return a full bundle for the certs it generates. Right now, there's a necessary second API call if you wish to get a certificate bundle.

This should either be a boolean or a "flavor" string.

PR #668, opened 2016-09-20
Add bundle option to newcert API

jacobhaven:
Addresses #667

### DIFF
--- a/api/generator/generator.go
+++ b/api/generator/generator.go
@@ -191,23 +191,20 @@ func NewCertGeneratorHandler(validator Validator, caFile, caKeyFile string, poli
 
 // NewCertGeneratorHandlerFromSigner returns a handler directly from
 // the signer and validation function.
-func NewCertGeneratorHandlerFromSigner(validator Validator, caBundleFile, intBundleFile string, signer signer.Signer) (http.Handler, error) {
-	handler := &CertGeneratorHandler{
-		generator: &csr.Generator{Validator: validator},
-		signer:    signer,
-	}
-
-	if caBundleFile != "" || intBundleFile != "" {
-		var err error
-		if handler.bundler, err = bundler.NewBundler(caBundleFile, intBundleFile); err != nil {
-			return nil, err
-		}
-	}
-
+func NewCertGeneratorHandlerFromSigner(validator Validator, signer signer.Signer) http.Handler {
 	return api.HTTPHandler{
-		Handler: handler,
+		Handler: &CertGeneratorHandler{
+			generator: &csr.Generator{Validator: validator},
+			signer:    signer,
+		},
 		Methods: []string{"POST"},
-	}, nil
+	}
+}
+
+// SetBundler allows injecting an optional Bundler into the CertGeneratorHandler.
+func (cg *CertGeneratorHandler) SetBundler(caBundleFile, intBundleFile string) (err error) {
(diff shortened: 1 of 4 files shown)


api/generator/generator.go
```diff
@@ -150,6 +151,7 @@ func (g *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
 // sending the CSR to the server.
 type CertGeneratorHandler struct {
 	generator *csr.Generator
+	bundler   *bundler.Bundler
```
kisom: Would it be easier to just use the API client to bundle the cert instead of adding a lot more parameters to the `NewCertGeneratorHandler`?

jacobhaven: The point is to avoid always having two API calls, each one possibly routed to different backend CAs.

jacobhaven:
@kisom Our use case is we have a single offline root CA and many "identical" intermediates, with each having a different cert/key stored in an HSM. These intermediates are all accessed via a load-balancing virtual IP.

This causes an issue where a `newcert` API call and the following `bundle` API call can be routed to different CAs (leading to a `bundle` failure). It would simplify things greatly to have a single api call combining the two functionalities.

So, I've implemented an optional extension to `newcert` so that the bundling will be done on the same CA as the cert generation. Let me know and I'll try to fix any problems you have with this.

fhossain:
In this case the intermediates are also short lived and it's not easy to make sure that they are "cloned"/synched to all the hosts. Our trust is based on the offline root, so we don't really care which intermediate we get a certificate from as long as the cert chain is created with proper bundle.

jacobhaven:
@lziest \*Updated to use injector function to set Bundler\*. `NewCertGenHandlerFromSigner` is now unaltered.

api/generator/generator.go
```diff
@@ -199,10 +201,17 @@ func NewCertGeneratorHandlerFromSigner(validator Validator, signer signer.Signer
 	}
 }
 
+// SetBundler allows injecting an optional Bundler into the CertGeneratorHandler.
+func (cg *CertGeneratorHandler) SetBundler(caBundleFile, intBundleFile string) (err error) {
```
kisom: I much prefer this approach.

api/generator/generator.go
```diff
@@ -275,6 +284,15 @@ func (cg *CertGeneratorHandler) Handle(w http.ResponseWriter, r *http.Request) e
 		},
 	}
 
+	if req.Bundle && cg.bundler != nil {
```
kisom: Should a message be sent to the user if `req.Bundle` is `true` but no bundler is defined?

doc/api/endpoint_newcert.txt
```diff
@@ -16,6 +16,8 @@ Optional parameters:
     the CSR, useful when interacting with cfssl server that stands
     in front of a remote multi-root CA signer
     * profile: a string specifying the signing profile for the signer
+    * bundle: a boolean specifying whether to include an "optimal"
```
kisom: Thanks for updating the docs!

api/generator/generator.go
jacobhaven: @kisom Updated to send error message. I wasn't sure of the exact best type, so I just went with (`PolicyError`, `InvalidRequest`)

PR #671, opened 2016-09-26
s/usage/usages/ in doc/cmd/cfssl.txt.

rektide:
The documentation for the cfssl command had mixed references to usage and usages, however the codebase seems to expect only usages.


lziest:
lgtm, eventually we need to clean up the naming in the struct too https://github.com/cloudflare/cfssl/blob/master/config/config.go#L75

PR #675, opened 2016-09-29
bugfix: fix a nil deference in transport key provider package

lziest:
- also we can just self bootstrap by using the incoming csr.CertificateRequest


lziest:
I may add a test later.

PR #681, opened 2016-10-17
Add yaml tags to CSR structs.

kisom:
This allows the transport package to use both YAML and JSON for
storing requests.


lziest:
lgtm

PR #682, opened 2016-10-19
Use a proper net/http.Client in api/client.

kisom:
Transport connections are reused, and constantly using a new Transport
can cause connections to stick around. It's better to scope them to
the lifetime of the client object.


lziest:
lgtm

PR #691, opened 2016-11-30
fix capitalized error string

lziest:


lziest:
The build on MacOS hangs forever. Cancel the build and confirm all tests pass otherwise.

kisom:
Curious why the third test never gets queued.

PR #699, opened 2016-12-14
Pass through OCSP response issuer hash function in ocsp.SignRequest

rolandshoemaker:
As of commit b07d8c9 golang.org/x/crypto/ocsp now supports letting the
consumer set which hashing function they wish to use when creating OCSP
responses. This change updates golang.org/x/crypto/ocsp to the latest
master commit and adds the field 'IssuerHash' to SignRequest which is
passed through to ocsp.CreateResponse.


api/ocsp/ocspsign.go
```diff
@@ -34,10 +36,11 @@ func NewHandler(s ocsp.Signer) http.Handler {
...
 type jsonSignRequest struct {
-	Certificate string `json:"certificate"`
-	Status      string `json:"status"`
-	Reason      int    `json:"reason,omitempty"`
-	RevokedAt   string `json:"revoked_at,omitempty"`
+	Certificate string      `json:"certificate"`
+	Status      string      `json:"status"`
+	Reason      int         `json:"reason,omitempty"`
+	RevokedAt   string      `json:"revoked_at,omitempty"`
+	IssuerHash  crypto.Hash `json:"issuer_hash,omitempty"`
```
kisom: I think it would make more sense to have this as a string; otherwise interacting with the API would require some knowledge of the Go standard library.

kisom:
Thanks, @rolandshoemaker.

PR #702, opened 2016-12-15
Add bundle option to /sign similar to /newcert

jacobhaven:
r? @kisom @lziest @fhossain

### DIFF
--- a/api/signhandler/signhandler.go
+++ b/api/signhandler/signhandler.go
@@ -170,7 +170,8 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
 
 // An AuthHandler verifies and signs incoming signature requests.
 type AuthHandler struct {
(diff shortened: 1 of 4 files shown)


jacobhaven:
I just did /sign and not /authsign as that's all we needed, but it would be trivial to copy the code if we want to have both.

fhossain:
@jacobhaven I think we should do authsign for the sake of completeness :-)

fhossain:
@jacobhaven I tested out the pull request branch and it fixes all the use cases I had with the second bundle call. All good from my end, Thanks.

jacobhaven:
@kisom Yeah, it's odd that it was; it's in the `.gitignore`... \*fixed\*

fhossain:
@kisom I think your changes were addressed. Is there some additional review changes outstanding with this pull request?

PR #704, opened 2016-12-27
MySQL: Certificate Serial Number should be 128 in length.

### LINKED ISSUE #703: MySQL: Wrong size for serial number column
Currently, the MySQL backend is non-functional due to a very simple mistake in https://github.com/cloudflare/cfssl/blob/master/certdb/mysql/migrations/001_CreateCertificates.sql#L5 : the size of the serial_number field should be 128 instead of 20, as obviously apparent in https://github.com/cloudflare/cfssl/blob/437c6213e569dde639fd116272c27557fd38eaf9/scan/vendor/crypto/tls/generate_cert.go#L107

I'd do a PR myself, but I'm not sure whether I should update the existing migration or create one. Consider that as it stands today, the current DB is non-functional in terms of revocation.

Waiting for input.

drf:
Modified previous migration as per @lziest suggestion.

Fixes #703

### DIFF
--- a/certdb/mysql/migrations/001_CreateCertificates.sql
+++ b/certdb/mysql/migrations/001_CreateCertificates.sql
@@ -2,7 +2,7 @@
 -- SQL in section 'Up' is executed when this migration is applied
 
 CREATE TABLE certificates (
-  serial_number            varbinary(20)  NOT NULL,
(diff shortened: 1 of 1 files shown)


lziest:
lgtm

Issue #711, opened 2017-01-08
Make CFSSL extensible with respect to local signers


smithbk:
CFSSL should be extensible with respect to local signers to enable various HSM solutions to be plugged in.  This applies to both CSR signing as well as OCSP processing in which a private key is assumed to be accessible in memory, rather than in hardware as is the case for an HSM.

smithbk:
Hi, any comments on this would be appreciated.  Thanks

jsha:
Looks like this was resolved by #712 and can be closed.

PR #712, opened 2017-01-08
Make CFSSL extensible with respect to local signers

smithbk:
This change has two parts:
1) Allow a local signer to be prepended to the list in universal.go
2) Split a function into two parts in OCSP processing
   to allow a signer object to be passed in.

See https://github.com/cloudflare/cfssl/issues/711

### DIFF
--- a/signer/universal/universal.go
+++ b/signer/universal/universal.go
@@ -50,6 +50,11 @@ var localSignerList = []localSignerCheck{
 	fileBackedSigner,
 }
 
+// PrependLocalSignerToList prepends signer to the local signer's list
+func PrependLocalSignerToList(signer localSignerCheck) {
+	localSignerList = append([]localSignerCheck{signer}, localSignerList...)
+}
+
 func newLocalSigner(root Root, policy *config.Signing) (s signer.Signer, err error) {
 	// shouldProvide indicates whether the
 	// function *should* have produced a key. If


smithbk:
Hi, any update to this would be appreciated.  Thanks

ocsp/ocsp.go
```diff
@@ -101,8 +101,27 @@ func ReasonStringToCode(reason string) (reasonCode int, err error) {
...
+		return nil, err
+	}
+
+	return NewSignerFromFileAndKey(issuerFile, responderFile, key, interval)
+}
+
+// NewSignerFromFileAndKey reads the issuer cert and the responder cert
+// from PEM files, and takes a crypto signer an interval in seconds,
+// and returns a signer
+func NewSignerFromFileAndKey(issuerFile, responderFile string, key crypto.Signer, interval time.Duration) (Signer, error) {
```
vpaprots: Was considering that name too, can make it so, not all that particular.. just.. it seemed too much like a "sea-shells-from-the-see" tongue twister "NewSigner..AndSigner"..

smithbk:
I removed the changes to ocsp.go, so now this pull request simply adds a function to prepend to the local signer list in universal.go

lziest:
LGTM, @kisom please review too. :)

PR #718, opened 2017-01-25
fix unused value assignments caught by static checker

lziest:
fix test errors in https://travis-ci.org/cloudflare/cfssl/jobs/195007036


lziest:
@kisom we finally get rid of static check errors.

PR #727, opened 2017-02-12
Optionally generate OCSP response in cert revocation API endpoint

lidavidm:
The certificate revocation handler in `api/revoke` now, when given a signer, will generate an OCSP response and insert it into the certdb; otherwise it is unchanged from before.

### DIFF
--- a/api/revoke/revoke.go
+++ b/api/revoke/revoke.go
@@ -2,20 +2,26 @@
 package revoke
 
 import (
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"io/ioutil"
 	"net/http"
+	"time"
 
 	"github.com/cloudflare/cfssl/api"
 	"github.com/cloudflare/cfssl/certdb"
 	"github.com/cloudflare/cfssl/errors"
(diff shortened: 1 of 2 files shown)


api/revoke/revoke.go
```diff
@@ -66,6 +83,53 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
...
+		if err != nil {
+			return err
+		}
+
+		if err = h.dbAccessor.InsertOCSP(certdb.OCSPRecord{
+			Serial: req.Serial,
+			AKI:    req.AKI,
+			Body:   string(ocspResponse),
+			Expiry: ocspParsed.NextUpdate,
+		}); err != nil {
```
lziest: style issue: split the InsertOCSP out of the if err != nil

lziest:
lgtm, it will be nice to fix the existing api_revoke test so it is more realistic.

lidavidm:
@lziest, I've added a new test of `api_revoke` that additionally checks that an OCSP response is generated.

lziest:
LGTM

PR #728, opened 2017-02-12
add Sqlite backend support

bjt79:
TODO: add tests


ocsp/responder.go
```diff
@@ -49,6 +51,49 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
...
+	aki := hex.EncodeToString(req.IssuerKeyHash)
+
+	sn := req.SerialNumber
+	if sn == nil {
+		return nil, false
+	}
+	strSN := sn.String()
+
+	records, err := src.Accessor.GetOCSP(strSN, aki)
+	if err == nil || len(records) == 0 {
```
lziest: This condition looks incorrect. Probably mean err != nil

lziest: Also, consider to add some logs. I think it's better to differentiate the errors here. if err != nil, log the error. If no record is returned, just return (nil, false)

ocsp/responder.go
```diff
@@ -19,6 +20,7 @@ import (
 
 	"github.com/cloudflare/cfssl/log"
 	"github.com/jmhodges/clock"
+	"github.com/cloudflare/cfssl/certdb"
```
lziest: According to `gofmt`, you need to move this line two lines above, so the alphabetic order of import block is maintained.

kisom: Alternatively, you can use [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) to handle import management automatically; it's a drop-in replacement for `gofmt`.

ocsp/responder.go
```diff
@@ -49,6 +51,55 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
...
+	// Extract the AKI string from req.IssuerKeyHash
+	aki := hex.EncodeToString(req.IssuerKeyHash)
+
+	sn := req.SerialNumber
+	if sn == nil {
+		return nil, false
+	}
+	strSN := sn.String()
+
+	records, err := src.Accessor.GetOCSP(strSN, aki)
```
lziest: Check Accessor is not nil, then we should be good.

lziest:
LGTM, now need some tests.

ocsp/responder_test.go
```diff
@@ -158,3 +165,61 @@ func TestNewSourceFromFile(t *testing.T) {
...
+
+	// call response method on request and check output
+	response, present := src.Response(req)
+	if !present {
+		t.Error("No response present for given request")
+	}
+
+	// print Serial Number for returned response and verify it is the same
+	resp, err := goocsp.ParseResponse(response, nil)
+	// ??? why can't I access any of the attributes of resp
```
lziest: becase the response would a "Test OCSP" which is not a valid OCSP response, so ParseResponse will return err, check err.

lziest: easier check is to check the response is exactly "Test OCSP"

ocsp/responder_test.go
```diff
@@ -158,3 +165,59 @@ func TestNewSourceFromFile(t *testing.T) {
...
+	//use Accessor to create new SqliteSource
+	src := NewSqliteSource(accessor)
+
+	// call response method on request and check output
+	response, present := src.Response(req)
+	if !present {
+		t.Error("No response present for given request")
+	}
+	responseString := string(response)
+	if responseString != "Test OCSP" {
```
lziest: simply do string(response) != "Test OCSP"

ocsp/responder_test.go
```diff
@@ -158,3 +165,129 @@ func TestNewSourceFromFile(t *testing.T) {
...
+		NextUpdate:   time.Now().AddDate(0, 1, 0),
+	}
+	keyPEM, _ := ioutil.ReadFile("testdata/ca-key.pem")
+	priv, err := helpers.ParsePrivateKeyPEM(keyPEM)
+	if err != nil {
+		t.Errorf("Error parsing private key: %s", err)
+	}
+
+	// create response
+	response, err := goocsp.CreateResponse(issuer, cert, template, priv)
```
lziest: nevertheless, check err,

ocsp/responder.go
```diff
@@ -49,6 +51,59 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
 	return
 }
 
+// SqliteSource represnts a source of OCSP responses backed by certdb
+type SqliteSource struct {
```
lziest: As I think it through, you can name this DBSource.

ocsp/responder.go
```diff
@@ -49,6 +51,59 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
 	return
 }
 
+// DBSource represnts a source of OCSP responses backed by certdb
```
kisom: This should be a complete sentence.

ocsp/responder.go
```diff
@@ -49,6 +51,59 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
...
+type DBSource struct {
+	Accessor certdb.Accessor
+}
+
+// NewDBSource creates a new DBSource type with associated dbAccessor
```
kisom: This should be a complete sentence.

ocsp/responder.go
```diff
@@ -49,6 +51,59 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
...
+func NewDBSource(dbAccessor certdb.Accessor) Source {
+	return DBSource{
+		Accessor: dbAccessor,
+	}
+}
+
+// Response implements cfssl.ocsp.responder.Source, returning the OCSP response
```
kisom: This should be a complete sentence.

kisom: The comment should also explain the meaning of the return arguments: what does the bool indicate? What's in the byte slice?

ocsp/responder.go
```diff
@@ -49,6 +51,59 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
...
+// with the expiration date furthest in the future
+func (src DBSource) Response(req *ocsp.Request) ([]byte, bool) {
+	if req == nil {
+		return nil, false
+	}
+
+	// Extract the AKI string from req.IssuerKeyHash
```
kisom: I don't think this comment is needed.

ocsp/responder.go
```diff
@@ -49,6 +51,59 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
...
+	}
+	strSN := sn.String()
+
+	if src.Accessor == nil {
+		log.Errorf("No DB Accessor")
+		return nil, false
+	}
+	records, err := src.Accessor.GetOCSP(strSN, aki)
+
+	// Log on errors obtaining OCSP response
```
kisom: This should be a complete sentence.

kisom: Should anything else happen when there's an error? Should the function continue? It would be useful to document that in a comment here.

ocsp/responder.go
```diff
@@ -49,6 +51,59 @@ func (src InMemorySource) Response(request *ocsp.Request) (response []byte, pres
...
+	if err != nil {
+		log.Errorf("Error obtaining OCSP response: %s", err)
+	}
+
+	if len(records) == 0 {
+		return nil, false
+	}
+
+	// Find the OCSPRecord with the expiration date furthest in the future
```
kisom: This should be a complete sentence.

ocsp/responder_test.go
```diff
@@ -158,3 +165,132 @@ func TestNewSourceFromFile(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestSqliteTrivial(t *testing.T) {
+
```
kisom: The comments in this function should be complete sentences.

kisom: This newline at the beginning isn't needed.

ocsp/responder_test.go
```diff
@@ -158,3 +165,132 @@ func TestNewSourceFromFile(t *testing.T) {
...
+func TestSqliteRealResponse(t *testing.T) {
+	// create new Accessor
+	sqliteDBfile := "testdata/sqlite_test.db"
+	db := testdb.SQLiteDB(sqliteDBfile)
+	accessor := sql.NewAccessor(db)
+
+	certFile := "testdata/cert.pem"
+	issuerFile := "testdata/ca.pem"
+	certContent, _ := ioutil.ReadFile(certFile)
```
kisom: The errors here should be checked.

ocsp/responder_test.go
```diff
@@ -158,3 +165,132 @@ func TestNewSourceFromFile(t *testing.T) {
...
+	accessor := sql.NewAccessor(db)
+
+	// populate the db with record and check response handles request appropiately
+	ocsp := certdb.OCSPRecord{
+		AKI:    hex.EncodeToString(req.IssuerKeyHash),
+		Body:   "Test OCSP",
+		Expiry: time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC),
+		Serial: req.SerialNumber.String(),
+	}
+	accessor.InsertOCSP(ocsp)
```
kisom: `InsertOCSP` returns an error, which should be checked here.

PR #731, opened 2017-02-18
Implement API endpoint for adding certs to a certdb

lidavidm:
This adds a new `http.Handler` that can insert certificates to a given certdb. Optionally, if given an OCSP signer, it will also generate and insert OCSP responses as certificates come in.

### DIFF
--- a/api/certadd/insert.go
+++ b/api/certadd/insert.go
@@ -1,4 +1,4 @@
-package insert
+package certadd
 
 import (
 	"bytes"
--- a/api/certadd/insert.go
+++ b/api/certadd/insert.go
@@ -15,8 +15,9 @@ import (
 	"github.com/cloudflare/cfssl/helpers"
 	"github.com/cloudflare/cfssl/ocsp"
 
-	stdocsp "golang.org/x/crypto/ocsp"
 	"encoding/base64"
+
+	stdocsp "golang.org/x/crypto/ocsp"
 )
 
 // This is patterned on
@@ -44,6 +45,8 @@ func NewHandler(dbAccessor certdb.Accessor, signer ocsp.Signer) http.Handler {
 	}
 }
 
+// AddRequest describes a request from a client to insert a
+// certificate into the database.
 type AddRequest struct {
 	Serial    string    `json:"serial_number"`
 	AKI       string    `json:"authority_key_identifier"`
--- a/api/insert/insert.go
+++ b/api/insert/insert.go
@@ -1,0 +1,180 @@
+package insert
+
+import (
+	"bytes"
+	"encoding/hex"
+	"encoding/json"
+	"io/ioutil"
+	"math/big"
+	"net/http"
+	"time"
+
+	"github.com/cloudflare/cfssl/api"
+	"github.com/cloudflare/cfssl/certdb"
+	"github.com/cloudflare/cfssl/errors"
+	"github.com/cloudflare/cfssl/helpers"
+	"github.com/cloudflare/cfssl/ocsp"
+
+	stdocsp "golang.org/x/crypto/ocsp"
+)
+
+// This is patterned on
+// https://github.com/cloudflare/cfssl/blob/master/api/revoke/revoke.go
+
+// A Handler accepts new SSL certificates and inserts them into the
+// certdb, creating an appropriate OCSP response for them.
+type Handler struct {
+	dbAccessor certdb.Accessor
+	signer     ocsp.Signer
+}
+
+// NewHandler creates a new Handler from a certdb.Accessor and ocsp.Signer
+func NewHandler(dbAccessor certdb.Accessor, signer ocsp.Signer) http.Handler {
+	return &api.HTTPHandler{
+		Handler: &Handler{
+			dbAccessor: dbAccessor,
+			signer:     signer,
+		},
+		Methods: []string{"POST"},
+	}
+}
+
+type jsonAddRequest struct {
+	Serial    string    `json:"serial_number"`
+	AKI       string    `json:"authority_key_identifier"`
+	CALabel   string    `json:"ca_label"`
+	Status    string    `json:"status"`
+	Reason    int       `json:"reason"`
+	Expiry    time.Time `json:"expiry"`
+	RevokedAt time.Time `json:"revoked_at"`
+	PEM       string    `json:"pem"`
+}
+
+// Map of valid reason codes
+var validReasons = map[int]bool{
+	stdocsp.Unspecified:          true,
+	stdocsp.KeyCompromise:        true,
+	stdocsp.CACompromise:         true,
+	stdocsp.AffiliationChanged:   true,
+	stdocsp.Superseded:           true,
+	stdocsp.CessationOfOperation: true,
+	stdocsp.CertificateHold:      true,
+	stdocsp.RemoveFromCRL:        true,
+	stdocsp.PrivilegeWithdrawn:   true,
+	stdocsp.AACompromise:         true,
+}
+
+// Handle handles HTTP requests to add certificates
+func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+	r.Body.Close()
+
+	var req jsonAddRequest
+
+	err = json.Unmarshal(body, &req)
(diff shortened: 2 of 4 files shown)


lidavidm:
Not too sure why the test is failing, since it's in an entirely different module. Perhaps it's a timing issue?

lziest:
some tests are using live internet, so the stability is questionable at times. :)

api/insert/insert.go
```diff
@@ -0,0 +1,180 @@
...
+	return &api.HTTPHandler{
+		Handler: &Handler{
+			dbAccessor: dbAccessor,
+			signer:     signer,
+		},
+		Methods: []string{"POST"},
+	}
+}
+
+type jsonAddRequest struct {
```
lziest: we can make this exported.

api/insert/insert.go
```diff
@@ -0,0 +1,180 @@
+package insert
```
lziest: certadd? I am not very good at naming though.

api/insert/insert.go
```diff
@@ -0,0 +1,180 @@
...
+			Body:   string(ocspResponse),
+			Expiry: ocspParsed.NextUpdate,
+		}
+
+		if err = h.dbAccessor.InsertOCSP(ocspRecord); err != nil {
+			return err
+		}
+	}
+
+	result := map[string]string{}
```
lziest: we may want to return {"certificate": the pem string, "ocsp_response": base64_encoded(ocspResponse)}, it gives API user a good UX, also it can enrich our testing by inspecting the return json object.

api/insert/insert.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if _, present := ocsp.StatusCode[req.Status]; !present {
+		return errors.NewBadRequestString("Invalid certificate status")
+	}
+
+	if ocsp.StatusCode[req.Status] == stdocsp.Revoked && req.RevokedAt == (time.Time{}) {
+		return errors.NewBadRequestString("Revoked certificate should specify when it was revoked")
+	}
+
+	if _, present := validReasons[req.Reason]; !present {
```
lziest: only check if that if the status is revoked?

api/insert/insert.go
lidavidm: I named this to be in parallel with `api/revoke`, but it makes sense to be more explicit about what it does.

api/insert/insert.go
lidavidm: I'm not sure I understand why we should also return the certificate data. I understand why having the OCSP response would be useful, however, the API caller would have provided the certificate PEM in the first place, so wouldn't it be a waste of (admittedly probably very little) bandwidth to simply echo it back to them?

lziest: It's somewhat wasteful to echo, but since OCSP response will be optional, I was thinking about returning something that works as confirmation, maybe a `"success": true`?

lziest: I also wanted to pair up OCSP resonse with the cert, just my personal weak preference.

lidavidm: Would it make more sense to additionally embed the OCSP response in the certificate in that case? That way we're at least freeing the client from some work.

lidavidm:
Changes:
- Export the insertion request struct
- Only validate the revocation reason if the cert is actually revoked
- Return the OCSP response to the caller
- Rename package to `certadd` to be in line with `certinfo`

api/certadd/insert.go
```diff
@@ -0,0 +1,190 @@
...
+	return &api.HTTPHandler{
+		Handler: &Handler{
+			dbAccessor: dbAccessor,
+			signer:     signer,
+		},
+		Methods: []string{"POST"},
+	}
+}
+
+type AddRequest struct {
```
lziest: thanks! But now you need to put comments/documents for this type.

api/insert/insert.go
lidavidm: Oh whoops, I mixed up OCSP and CT stapling.

For `"success": true`, doesn't the fact that we use `api.SendResponse` already include that? This is how a response currently looks:
```
{"success":true,"result":{"ocsp_response":"MIIE2goBAKCCBNMwggTPBgkrBgEFBQcwAQEEggTAMIIEvDCBr6EmMCQxIjAgBgNVBAoTGUNvcm5lbGwgQ1MgNTE1MiBSZXNwb25kZXIYDzIwMTcwMzAyMTUxMTAwWjB0MHIwSjAJBgUrDgMCGgUABBQjMeG40BC+y9MwFMu1wlMuRbbG6QQUyTAQC0ffA+MPxMYWqutvnjIcEP4CEQD/HB5oBaqHa5iDYeQzzNNxgAAYDzIwMTcwMzAyMTUwMDAwWqARGA8yMDE3MDMwMjE2MDAwMFowDQYJKoZIhvcNAQELBQADggEBAFKTiCwk2hv4XmicNgCN5OzHuzL6ETdxH/L6alnPia+tEM13HwHnNNQ9l0iOtcS1ynvxQJizGnWip7Yy29/1Yk3OT2NTqp8svp7pk7n38Y1PXoRrXGF/GWhPcR/AHvneUmc0+HMzvHCSR5njT/hoXFZLBKHvWPiEKMjd86Jc+dQ9MaWYoChkShfl193CVH9VU2jrDDhnXk2G/CbjfRjKdSXrMDUheoCYd7M6d8aO1Yq1TB3HPnOjf1wrDRwXJNYdLHJx9y75vc8P4vK2UTVGd3pVKJ6at+oQvaecpy82p9Nufh98dZNnjLr0M+NmFlN7jJthvl3w4NpRBeZGBs9lfa+gggLyMIIC7jCCAuowggHSoAMCAQICEQDEfqDVUUMWfUMHx4oKsDMoMA0GCSqGSIb3DQEBCwUAMCQxIjAgBgNVBAoTGUNvcm5lbGwgQ1MgNTE1MiBSZXNwb25kZXIwIhgPMDAwMTAxMDEwMDAwMDBaGA8wMDAxMDEwMTAwMDAwMFowJDEiMCAGA1UEChMZQ29ybmVsbCBDUyA1MTUyIFJlc3BvbmRlcjCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAJVDcQNCkhdNCSPmcOwfbpalc7VHma1NViAVcG1OrjPuLVpeYnll7pM3VAZ/0vpT0bPSuPmMSzJH9NG5EOHqzBoMve0/Babk7rk7X+qgSW5ROSAHKoyCFR0IQL1ELELzGmyzUUzbAObZqBPCMQW8ceLqD42rFuLJlCCIvxr/4+LmcC3io7If9V34l6gI3FUfQ3uWl8At3LO0gzBJ/PQ7OktZkZPbDz69moZXSFs4nbAkI39EtivI7LLmFHt3qqr+Su7bfoVboD96iQfuc5bR8MJj3Ses8SR+CQSWxHd/tY/LsGDA0O/gAo7L7BR+0V/pDBOfuWMn8ZpNpGaOoPz4TG8CAwEAAaMTMBEwDwYDVR0jBAgwBoAEKioqKzANBgkqhkiG9w0BAQsFAAOCAQEAfdzDT6yQA9I+6bIM/wKk3BOfJ6PTefrek4iamW56+fy3/bD66Qkl/18xoGrd/MspQy1b1TXOrNQiDYMY62bDDRjBBR5m0IP1N6OT2p30E3APE0KU0obqUXhFcfr1XjQUR+ImDNoIFUMXxVQ4sI3M5emSDNzq0aZhsAyiDwzdaf2Rd50Qzio9h3p3rqgtQBGQ/j5upXRT+HTXNsVq5hFXbycUcZJCDjSkaq0Z7G2OJclrpCdmd0mxFWI7yBIoDncL9nnohsiyGXV2F2RyrtXVaW4JzZAeklP/gAfn0ROdihDd6OIBtYIb4KsTy/uXrvoC6R4b9XF539lKh0Y1+pW9dA=="},"errors":[],"messages":[]}
```

Quick question - in the case that the certificate being added is not signed by the issuer, we can't create an OCSP response. Should this be considered an error in our response to the client? Or should we use something like `api.SendResponseWithMessage` to indicate that?

lziest: apologies, I missed that `"success": true` is there by default. For not being able to give OCSP response, let's keep it fail-open for now.

lziest:
LGTM

PR #734, opened 2017-03-07
Fix nil pointer panic in ParseCSRPEM 

cyli:
Currently `ParseCSRPEM` will panic if the CSR bytes to be parsed cannot be decoded as a valid PEM block.  This adds a check to ensure that the block is valid before passing the DER bytes to `x509.ParseCertificateRequest`.

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -450,8 +450,10 @@ func ParseCSR(in []byte) (csr *x509.CertificateRequest, rest []byte, err error) 
 // locally.
 func ParseCSRPEM(csrPEM []byte) (*x509.CertificateRequest, error) {
 	block, _ := pem.Decode([]byte(csrPEM))
-	der := block.Bytes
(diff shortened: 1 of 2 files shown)


helpers/helpers.go
```diff
@@ -449,8 +449,11 @@ func ParseCSR(in []byte) (csr *x509.CertificateRequest, rest []byte, err error)
 // It does not check the signature. This is useful for dumping data from a CSR
 // locally.
 func ParseCSRPEM(csrPEM []byte) (*x509.CertificateRequest, error) {
+	var der []byte
 	block, _ := pem.Decode([]byte(csrPEM))
-	der := block.Bytes
```
lziest: more elegant way is that check block and throw error immediately, instead of relying on CSR parsing to catch the empty input.

lziest:
won't wait for the snail-speed OSX CI env

cyli:
Thanks!

lziest:
Thank you!

PR #736, opened 2017-03-13
Verify delegated & CA signed OCSP correctly

vanbroup:
Only check the OCSP response signature from the issuer when no delegated signing certificate is included in the response. When a delegated OCSP signing certificate is present verify this is issued by the same CA.

### DIFF
--- a/revoke/revoke.go
+++ b/revoke/revoke.go
@@ -239,15 +239,12 @@ func certIsRevokedOCSP(leaf *x509.Certificate, strict bool) (revoked, ok bool) {
 	}
 
 	for _, server := range ocspURLs {
-		resp, err := sendOCSPRequest(server, ocspRequest, issuer)
+		resp, err := sendOCSPRequest(server, ocspRequest, leaf, issuer)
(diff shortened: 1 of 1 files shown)


lziest:
lgtm, but please rebase on top of the master.

vanbroup:
@lziest, done

kisom:
@vanbroup I've fixed our tests on master; can you rebase?

vanbroup:
When resp.Certificarte is present it should contain a delegated signing certificate (not the issuer, you should already have that). When it doesn't contain a delegated signing certificate the response should not be delegated and be signed by the issuer directly. 

I started thinking why ParseResponseForCert in "golang.org/x/crypto/ocsp" would not already verify the signature. After verification I concluded that it does already verify the signature. I think it makes more sense to remove the OCSP signature verification from CFSSL as it has no value to do it twice.

https://github.com/golang/crypto/blob/master/ocsp/ocsp.go#L524

lziest:
Much appreciated.

PR #737, opened 2017-03-13
Update checkout path for staticcheck.

kisom:
This fixes the build errors from #736; this is due to staticcheck moving repos.


kisom:
It looks like the latest updates to staticcheck break with Go < 1.8.

lziest:
ok, I agree we should drop Mac CI env.

Issue #735, opened 2017-03-09
Pass the CA and key to sign or gencert as text rather than files?


deitch:
Normally, the way to generate a cert is something like:

```
cfssl gencert -ca ca-cert.pem -ca-key ca-key.pem csr.json
```

Is there any way to pass it in a command-line, e.g.

```
-ca-material lots-of-text-here ...
```

Specifically, I am thinking of scenarios in which access to the filesystem is constrained, or we don't want to write the files out, everything in the docker cfssl/cfssl container (without volume mounts), etc.

gebi:
Something like this would be awesome, though i'd prefer to put e.g the CA key into the process environment.

Our usecase would be certificate creation from ansible and the CA key is ansible-vault encrypted.

deitch:
Yeah, I am all for passing it into the environment. Either way, as long as we can pass the key material as data and not just a file.

kisom:
There isn't a way to do this right now, no; it's not something we're likely to prioritise working on here right now because it's not how we're using it (generally for something similar, we write out files to a tmpfs). That being said, we're happy to review PRs. 

One approach might be to support a similar construct to that used in the [auth package](https://github.com/cloudflare/cfssl/blob/master/auth/auth.go#L50), such as prefixing an environment variable with "env:".

Reading from standard input is trickier, because there are multiple items that you might want to read from stdin (keys, config files with API keys, etc) and it becomes an issue of figuring out what order things need to be in and what's expected to come in.

deitch:
> Reading from standard input is trickier, because there are multiple items that you might want to read from stdin (keys, config files with API keys, etc) and it becomes an issue of figuring out what order things need to be in and what's expected to come in.

Yeah, that's true. Of course cfssl already _outputs_ several things that way by using json, so that could be a valid way to _input_ things.

kisom:
It could --- we'd just need to figure out what that looks like and how it'd work.

tkellen:
Hi all!

It looks like some functionality was implemented around this. Was a new version incorporating these changes released?

tkellen:
PS: If support for this is in the source tree and just waiting on someone to do a release I'd be happy to tackle that with a bit of guidance.

kisom:
I believe there is support for this in the tree, we have a few things to close out before our next release.

PR #750, opened 2017-04-16
Add support for passing CA data to gencert from environment (#735)

sfurman3:


sfurman3:
@lziest @kisom The Travis CI test failed only for go 1.6 (on os x). Is it possible that running the tests again might fix this?

helpers/helpers.go
```diff
@@ -620,3 +622,27 @@ func SCTListFromOCSPResponse(response *ocsp.Response) ([]ct.SignedCertificateTim
 	}
 	return *sctList, err
 }
+
+// ReadDERBytes reads a []byte either from a file or an environment variable.
+// If valFile has a prefix of 'env:', the []byte is read from the environment
+// using the subsequent name. If the prefix is 'file:' the []byte is read from
+// the subsequent file. If no prefix is provided, valFile is assumed to be a
+// file path.
+func ReadDERBytes(valFile string) ([]byte, error) {
```
lziest: can it be named as ReadBytes?

cli/gencert/gencert.go
```diff
@@ -31,6 +31,8 @@ Usage of gencert:
 
 Arguments:
         CSRJSON:    JSON file containing the request, use '-' for reading JSON from stdin
+	cert:	    file or environment variable containing the CA certificate -- use '[file:]fname' or 'env:varname'
```
lziest: update the helper text here https://github.com/cloudflare/cfssl/blob/master/cli/config.go#L78 (and line 79 too)
and then we don't need to specify this at `gencert` because it will be available to all cil command that uses `ca` and `ca-key`

lziest:
In general looks good to me.

lziest:
LGTM, @kisom would you share your though on this?

PR #770, opened 2017-05-19
Strip leading slash from OCSP request if present

rolandshoemaker:
Fixes certain issues we've seen in the wild. Based on the ASN.1 structure of OCSP requests we should never see a valid request that starts with a slash.


lziest:
LGTM

PR #771, opened 2017-05-19
OCSP: Use url.PathUnescape instead of url.QueryUnescape

rolandshoemaker:
Turns out there is already [a method](https://golang.org/pkg/net/url/#PathUnescape) that handles the `'+' -> ' '` encoding issue so we don't have to do it manually.


rolandshoemaker:
Blerp, didn't realize `url.PathUnescape` was added in 1.8, unless we bump the minimum version this change is a no-go.

lziest:
I need to update the test suite a little bit to suit the 1.8. So I will try to re-run the test and merge this after the update.

rolandshoemaker:
For what it's worth the main issue with tests seems to be that a number of the test certificates are malformed (`x509: RSA key missing NULL parameters`) and need to be renegerated to conform with the new stricter `crypto/x509` parsing rules.

PR #772, opened 2017-05-23
Expand OCSP Source interface

### LINKED ISSUE #732: API changes for OCSP responder
Currently the OCSP responder uses this interface to allow user-defined data sources:

```
type Source interface {
	Response(*ocsp.Request) ([]byte, bool)
}
```

This has two problems:

 - Since it returns only a bool (`found`), it has no way to distinguish between "no response found" and "internal server error." For instance, a client that asks for OCSP for a non-existent serial should get an OCSP response indicating "unauthorized," and an HTTP 200 status. A database failure or timeout should result in HTTP 500 or 503.

 - There's not a good way to override HTTP headers. For instance, Let's Encrypt's OCSP responder is fronted by Akamai. The caching headers that responder.go provides are RFC 5019 compliant, and indicate that a response may be cached until its NextUpdate (effectively, until the response expires). That's correct behavior, but for our own edge cache we would like to send special Edge-Control headers that tell Akamai to re-fetch after 3 days instead of the 7 days indicated by the NextUpdate field in the response, because otherwise no one can actually retrieve the fresh responses when we generate them, because they older ones are still in the Akamai cache.

For the first problem, I suggest changing the Source interface to return an error instead of a bool, and to define a `NotFoundError` that implementations can return for not found. Any other error would be trigger an HTTP 5xx status code.

I'm not sure of the best way to solve the second problem. Thoughts welcome!

cc @rolandshoemaker @cpu

rolandshoemaker:
Changes the `ocsp.Source` interface from `Response(*ocsp.Request) ([]byte, bool)` to `Response(*ocsp.Request) ([]byte, http.Header, error)` so that a source can provide more insight into what has happened when looking for a response than found/not found as well as allowing it to provide specific headers to be set in the subsequent response.

Will add a few more tests + better documentation of the new interface style but wanted to get maintainers thoughts before getting too much deeper.

Fixes #732.


lziest:
is this ready?

ocsp/responder.go
```diff
@@ -161,6 +166,19 @@ func NewResponder(source Source) *Responder {
 	}
 }
 
+func writeExtraHeaders(response http.ResponseWriter, headers http.Header) {
```
lziest: this feels more like overrideHeaders, you intentionally delete original key value and add headers

ocsp/responder.go
```diff
@@ -30,14 +31,18 @@ var (
...
+	ErrNotFound = errors.New("Request OCSP Response not found")
 )
 
 // Source represents the logical source of OCSP responses, i.e.,
 // the logic that actually chooses a response based on a request.  In
 // order to create an actual responder, wrap one of these in a Responder
 // object and pass it to http.Handle.
 type Source interface {
-	Response(*ocsp.Request) ([]byte, bool)
+	Response(*ocsp.Request) ([]byte, http.Header, error)
```
lziest: I prefer some docs around why we design this func. It implies that each response will have its own set of header. And in this stub library we actually never use this feature.  Can header be specific to one response, or just specific to the entire source.

ocsp/responder.go
```diff
@@ -243,12 +261,21 @@ func (rs Responder) ServeHTTP(response http.ResponseWriter, request *http.Reques
...
-	if !found {
-		log.Infof("No response found for request: serial %x, request body %s",
-			ocspRequest.SerialNumber, b64Body)
-		response.Write(unauthorizedErrorResponse)
-		return
+	ocspResponse, extraHeaders, err := rs.Source.Response(ocspRequest)
+	if err != nil {
+		if err == ErrNotFound {
+			if extraHeaders != nil {
+				writeExtraHeaders(response, extraHeaders)
```
jsha: Go idiom is to avoid returning meaningful values when an err return is non-nil, so we should probably ignore the extraHeaders on error returns.

ocsp/responder.go
rolandshoemaker: Makes sense.

ocsp/responder.go
rolandshoemaker: I'll write some extra docs here, the main impetus is that we'd like to calculate the cache control period for each response based on the ThisUpdate/NextUpdate fields instead of setting a static cache period over the entire source (as well as set some CDN specific headers on a per-response basis).

ocsp/responder.go
rolandshoemaker: :+1:

rolandshoemaker:
Ready for re-review.

jsha:
LGTM

lziest:
lgtm

lziest:
Hoping osx build won't time out this time, I just restarted it.

rolandshoemaker:
OS X build still seems unhappy.

lziest:
I will deal with it at master branch

PR #776, opened 2017-06-02
DoS attack possible when reading from CRLDistributionPoints

saadk10:
In cfssl/revoke/revoke.go, there are multiple places which use ioutil.ReadAll to read the CRL. If a rogue certificate is sent with a CRLDistributionPoint to http://mal-server/100G.file, then this code will consume all memory.

Added setter functions that allow you to pass in a function that handles how to read the http response body for various HTTP Gets allowing for more flexibility.

### DIFF
--- a/revoke/revoke.go
+++ b/revoke/revoke.go
@@ -11,6 +11,7 @@ import (
 	"encoding/base64"
 	"encoding/pem"
 	"errors"
+	"io"
 	"io/ioutil"
 	"net/http"
 	neturl "net/url"
@@ -104,7 +105,7 @@ func fetchCRL(url string) (*pkix.CertificateList, error) {
 		return nil, errors.New("failed to retrieve CRL")
 	}
 
-	body, err := ioutil.ReadAll(resp.Body)
+	body, err := crlRead(resp.Body)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


lziest:
Thanks, will review soon, lgtm at first glance

revoke/revoke.go
```diff
@@ -303,3 +304,24 @@ func sendOCSPRequest(server string, req []byte, leaf, issuer *x509.Certificate)
 
 	return ocsp.ParseResponseForCert(body, leaf, issuer)
 }
+
+var crlRead = ioutil.ReadAll
+
+// SetCRLFetcher sets the function to use to read from the http response body
```
lziest: I hope we can mention 'DDoS' to justify it.

smithbk:
Yes, without this, there seems to be no way to avoid a DoS attack, so I think it is very important.
I think there are 3 options:
1) Allow CFSSL users to provide their own read function as this pull request does.
2) CFSSL provides its own read function which limits the read buffer to some configurable max size.  This would make it easier for all users of CFSSL.
3) Both 1 & 2, but given #2 I can't think of why someone would need #1 but you never know.

saadk10:
@lziest Any updates on this?

lziest:
LGTM

PR #785, opened 2017-07-13
new certdb method: GetRevokedAndUnexpiredCertificatesByLabel

lziest:


lziest:
static check seems to be too slow to be tolerated by CI

PR #794, opened 2017-08-16
Allow specifying/overriding OCSP validity period

rolandshoemaker:
Allows for backdating OCSP responses which is something we've wanted to do in [letsencrypt/boulder](https://github.com/letsencrypt/boulder) for a while.

Adds a test that validates this doesn't break current usage and tests the new behavior.


lziest:
lgtm, wait on build

rolandshoemaker:
OSX build appears unhappy, not entirely sure why.

lziest:
we use a staticcheck lib which usually took around 30 minutes to finish. So it will get shut down for no output for 30 minutes.

lziest:
yolo mac build.

PR #797, opened 2017-08-31
cfssljson: add --version flag

flandr:
As noted in https://github.com/kelseyhightower/kubernetes-the-hard-way/blob/master/docs/02-client-tools.md#verification, cfssljson lacks a way to check its version.

### DIFF
--- a/cli/version/version.go
+++ b/cli/version/version.go
@@ -27,9 +27,14 @@ Usage of version:
 	cfssl version
 `
 
+// FormatVersion returns the formatted version string.
+func FormatVersion() string {
+	return fmt.Sprintf("Version: %s\nRevision: %s\nRuntime: %s\n", versionString(), version.Revision, runtime.Version())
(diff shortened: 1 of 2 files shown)


lziest:
LGTM

cli/version/version.go
```diff
@@ -27,9 +27,14 @@ Usage of version:
 	cfssl version
 `
 
+// Returns the formatted version string.
```
lziest: golint error: the exported function comments/docs should start with the function name.

kisom: The Go standard is to have the function name in the docstring, so this should be 

```
// FormatVersion returns the formatted version string.
```

[golint](https://github.com/golang/lint) in our builds will catch this and cause the build to fail.

flandr:
NP, done.

flandr:
Fixed formatting :/

PR #803, opened 2017-09-11
Use stdlib CSR CheckSignature

### LINKED ISSUE #802: Replace CSR CheckSignature with stdlib one
When CFSSL was first implemented, there was not a method in Go's `x509` to check the signature on a CSR, so CFSSL copied an in-progress version. This is now available from the standard `x509` library, so the implementation should be replaced.

rolandshoemaker:
Fixes #802.


lziest:
LGTM.

Issue #801, opened 2017-09-11
Storing AKI in CertDB fails with Go 1.9


mastersingh24:
cfssl passes an X509 certificate template [here ](https://github.com/cloudflare/cfssl/blob/master/signer/local/local.go#L385)and then assumes that X509.CreateCertificate will fill in the AKI in the template and sets the AKI with this assumption [here ](https://github.com/cloudflare/cfssl/blob/master/signer/local/local.go#L395)

Unfortunately, Go 1.9 no longer fills in the AKI in the template passed to X509.CreateCertificate.  You can see that [here ](https://golang.org/src/crypto/x509/x509.go?s=19833:23432#L1786)

A proposed fix for this would be to parse the signed cert bytes which are returned and extract the AKI from the signed certificate rather than the template passed in

lziest:
thanks for reporting. I think I saw those lines of code prior to 1.9 (unless you are referring a different piece of code)
```
    authorityKeyId := template.AuthorityKeyId
    if !bytes.Equal(asn1Issuer, asn1Subject) && len(parent.SubjectKeyId) > 0 {
        authorityKeyId = parent.SubjectKeyId
    }
```


So my understanding is that:
AKI is still populated when the the cert to-be-signed is not self-signed, and the signer CA cert has Subject Key Id, then CA's subject key id is the certTBS's AKI.

mastersingh24:
Right - so the cert is signed and has the right AKI.  The problem is that the AKI does not get stored in CertDB (if you are using CertDB).

        var signedCert []byte
	signedCert, err = s.sign(&certTBS, profile)
	if err != nil {
		return nil, err
	}

	if s.dbAccessor != nil {
		var certRecord = certdb.CertificateRecord{
			Serial: certTBS.SerialNumber.String(),
			// this relies on the specific behavior of x509.CreateCertificate
			// which updates certTBS AuthorityKeyId from the signer's SubjectKeyId
			AKI:     hex.EncodeToString(certTBS.AuthorityKeyId),
			CALabel: req.Label,
			Status:  "good",
			Expiry:  certTBS.NotAfter,
			PEM:     string(signedCert),
		}

		err = s.dbAccessor.InsertCertificate(certRecord)
		if err != nil {
			return nil, err
		}
		log.Debug("saved certificate with serial number ", certTBS.SerialNumber)
	}

mastersingh24:
I'm happy to submit an attempt at a fix?

lziest:
[...]

PR #807, opened 2017-09-12
Store AKI in CertDB with Go 1.9 (#801)

mastersingh24:
Go 1.9 modified the x509.CreateCertificate
function to no longer update the AuthorityKeyId
on the template passed in.  cfssl signer relied
on this behavior in order to get the AKI and save
it in certdb.  This change now parses the signed
cert and uses the AKI from it.

### DIFF
--- a/signer/local/local.go
+++ b/signer/local/local.go
@@ -387,12 +387,17 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		return nil, err
 	}
 
+	// Get the AKI from signedCert.  This is required to support Go 1.9+.
+	// In prior versions of Go, x509.CreateCertificate updated the
+	// AuthorityKeyId of certTBS.
+	parsedCert, _ := helpers.ParseCertificatePEM(signedCert)
+
 	if s.dbAccessor != nil {
 		var certRecord = certdb.CertificateRecord{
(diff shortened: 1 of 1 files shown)


signer/local/local.go
```diff
@@ -387,12 +387,18 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		return nil, err
 	}
 
+	// Get the AKI from signedCert.  This is required to support Go 1.9+.
+	// In prior versions of Go, x509.CreateCertificate updated the
+	// AuthorityKeyId of certTBS.
+	block, _ = pem.Decode([]byte(signedCert))
+	parsedCert, _ := x509.ParseCertificate(block.Bytes)
```
lziest: you can do parsedCert, _ := helpers.ParseCertificatePEM(signedCert)
https://godoc.org/github.com/cloudflare/cfssl/helpers#ParseCertificatePEM

mastersingh24: Cool.  Missed that.  Will modify

mastersingh24:
What's the likelihood of this one getting merged?  My current project uses cfssl and we are moving to Go 1.9.  We can implement a workaround in our code, but we'd rather just vendor an updated version of cfssl with this fix in it

kisom:
@mastersingh24 100% ;)

PR #817, opened 2017-10-16
Graceful shutdown for ocsp server

evantill:
Also return an exit code to 0 if there are no errors.

Require Go 1.8

### DIFF
--- a/cli/ocspserve/ocspserve.go
+++ b/cli/ocspserve/ocspserve.go
@@ -5,10 +5,14 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-
 	"github.com/cloudflare/cfssl/cli"
 	"github.com/cloudflare/cfssl/log"
 	"github.com/cloudflare/cfssl/ocsp"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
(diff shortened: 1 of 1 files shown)


cli/ocspserve/ocspserve.go
```diff
@@ -41,11 +45,47 @@ func ocspServerMain(args []string, c cli.Config) error {
...
+			log.Fatalf("could not shutdown: %v", err)
+		}
+		close(exit)
+	}()
+	err = server.ListenAndServe()
+	<-exit
+	if err != http.ErrServerClosed {
+		log.Fatalf("listen: %s\n", err)
+		return err
+	} else {
```
lziest: go vet complaint:
if block ends with a return statement, so drop this else and outdent its block

evantill:
I have deleted my repository since.
reopening it in PR https://github.com/cloudflare/cfssl/pull/1272

cli/ocspserve/ocspserve.go
evantill: I have changed in the new PR https://github.com/cloudflare/cfssl/pull/1272

PR #820, opened 2017-10-20
Add Update to facilitate CA renewals.

kisom:


initca/initca.go
```diff
@@ -220,3 +219,17 @@ var CAPolicy = func() *config.Signing {
 		},
 	}
 }
+
+// CopyAndResign copies the CA certificate, updates the NotBefore and
+// NotAfter fields, and then re-signs the certificate.
+func CopyAndResign(ca *x509.Certificate, priv crypto.Signer) (cert []byte, err error) {
```
lziest: Re-issue? Resign is kinda... negative

lziest: Or just renew?

kisom: We're also not issuing anything, really. Renamed to update. Now we wait for the tests to re-run.

initca/initca.go
```diff
@@ -220,3 +219,17 @@ var CAPolicy = func() *config.Signing {
 		},
 	}
 }
+
+// Update copies the CA certificate, updates the NotBefore and
+// NotAfter fields, and then re-signs the certificate.
+func Update(ca *x509.Certificate, priv crypto.Signer) (cert []byte, err error) {
```
lziest: shall we check the CA bit of the input cert? And basically we don't need `Renew...` function any more, right?

kisom: I don't think it really matters. If the CA bit isn't set, you're just ending up with a self-signed cert anyways. This is really just a simple copy-and-update function. The Renew function is used elsewhere right now, so it should be deprecated, not removed right away.

PR #838, opened 2018-01-11
Added proxy to api/client. Closes #837

### LINKED ISSUE #837: Add Proxy to Client API
There is no way to use the client api behind a proxy in cfssl/api/client/client.go

We could add a new a new function like, AddProxy to add a proxy function to the http.Transport

ogimenezb:
Should close, if accepted #837


kisom:
Thanks, @ogimenezb!

PR #842, opened 2018-01-19
cfssljson: fixes parsing of bundle and root certificates

smcquay:
The format of the json returned is currently `resp["result"]["bundle"]["root"]` and `resp["result"]["bundle"]["bundle"]`.

The original implementation missed the outer "result" json object.

### DIFF
--- a/cmd/cfssljson/cfssljson.go
+++ b/cmd/cfssljson/cfssljson.go
@@ -154,26 +154,28 @@ func main() {
 		})
 	}
 
-	if contents, ok := input["bundle"].(map[string]interface{}); ok {
-		if certificateBundle, ok := contents["bundle"].(string); ok {
-			if rootCertificate, ok := contents["root"].(string); ok {
-				outs = append(outs, outputFile{
-					Filename: baseName + "-bundle.pem",
-					Contents: certificateBundle + "\n" + rootCertificate,
-					Perms:    0644,
-				})
-				outs = append(outs, outputFile{
-					Filename: baseName + "-root.pem",
-					Contents: rootCertificate,
-					Perms:    0644,
-				})
+	if result, ok := input["result"].(map[string]interface{}); ok {
+		if bundle, ok := result["bundle"].(map[string]interface{}); ok {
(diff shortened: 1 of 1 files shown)


lziest:
@kisom is it related to the `-bare` param? This is like an "non-bare" response parsing.

smcquay:
@lziest when I use this command it's as:

```bash
$ our-cert-request | cfssljson -bare
```

lziest:
If you get rid of `-bare`, would that parse that result for you?

smcquay:
@lziest: The tool I mentioned above does the following: it sends: 

```json
{"hosts":null,"certificate_request":"-----BEGIN CERTIFICATE REQUEST-----< redacted >-----END CERTIFICATE REQUEST-----\n","profile":"","bundle":true} 
```

to the following api:

    /api/v1/cfssl/sign  

and get the following response (unrelated keys/values redacted):

```json
{
  "result": {
    "bundle": {
      "bundle": "-----BEGIN CERTIFICATE----- < redacted > -----END CERTIFICATE-----\n-----BEGIN CERTIFICATE----- < redacted > -----END CERTIFICATE-----\n-----BEGIN CERTIFICATE----- < redacted > -----END CERTIFICATE-----",
      "root": "-----BEGIN CERTIFICATE----- < redacted > -----END CERTIFICATE-----",
    },
  },
}
```

And we'd like to create two files:

- a `${name}-bundle.pem` file with the full chain as provided in `response["result"]["bundle"]["bundle"]`
- a `${name}-root.pem` which contains the value from `response["result"]["bundle"]["root"]`

The code that was there before had a bug where it wasn't looking at `response["result"]`; it was searching for the values at:

- `response["bundle"]["bundle"]`
- `response["bundle"]["root"]`

Which never existed. The patch I've provided fixes this latent bug.

lziest:
I will verify this behavior, to me it was working fine but my local copy may not be up to date.

kisom:
@lziest is this okay to merge?

lziest:
LGTM

PR #844, opened 2018-01-23
ocsp: Reduce user-triggerable logs to debug.

jsha:
There are some log lines that can be triggered by users submitting malformed
input. Since malformed input is unfortunately common on the internet, reduce
these log lines to debug level to decrease log spam.


lziest:
lgtm

kisom:
Thanks, @jsha.

PR #846, opened 2018-01-30
Update Dockerfile.minimal

jacob-salassi:
Path to cfssl_trust was nested under cfssl for the final copy block, but nothing in the build sequence satisfied this dependency. Trigger fetching of cfssl_trust via go get during first stage, then copy those files from 1st to 2nd stage.


jacob-salassi:
Hey all, I studied this problem for a while and it just doesnt seem like the original Dockerfile.minimal could work, or I was not smart enough to figure out how to satisfy the cfssl_trust requirement in the way it was used via the dockerfile. Hope this is a fix you can merge.

kisom:
Thanks, @jacob-salassi!

PR #847, opened 2018-01-31
activate usage of CNOverride flag in gencert

kron4eg:


kron4eg:
CNOverride flag is defined but wasn't used, despite it's very convenient to use (to avoid json editing every time).

kisom:
@kron4eg Thanks!

PR #767, opened 2017-05-16
update go version

lziest:

PR #852, opened 2018-02-09
Move template filling out of signer.sign

rolandshoemaker:
Paves the way for [further work](https://github.com/cloudflare/cfssl/compare/master...rolandshoemaker:precert-changes?expand=1) on making SCT embedding public. Since this makes changes to the flow of `signer.Sign` I wanted to check this didn't break any assumptions you folks had about the design.

Now `signer.sign` only really does signing (and ca initialization stuff if you are doing that).


signer/local/local.go
```diff
@@ -336,14 +326,23 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		}
 	}
 
+	var distPoints = safeTemplate.CRLDistributionPoints
+	if distPoints != nil && len(distPoints) > 0 {
+		safeTemplate.CRLDistributionPoints = distPoints
```
kisom: It might just be not having had enough coffee yet, but I don't see what this is doing.

rolandshoemaker: I blindly copied this from `Signer.sign` but yeah you're right, this doesn't appear to actually _do anything_. It appears [when this was introduced](https://github.com/cloudflare/cfssl/commit/6728e3be9365e6fa8bc699014a2046ca6c82c27a#diff-8d79bfc4bb0987be2c41ab248b24b30fR101) the `if` was after `FillTemplate`, which allowed the profile `CRLDistributionPoints` to be overridden by those in the template. In #767 it appears to have been moved before `FillTemplate` where it now does nothing.

rolandshoemaker: (If the overriding behavior is something we wish to preserve I'm happy to move it back to after `FillTemplate` in this PR.)

rolandshoemaker:
@kisom @lziest the two test DB changes that are here seem to appear every time I run `test.sh`, any reason to not have them included in the `.gitignore` file?

lziest:
@rolandshoemaker I haven't found a good way to express in gitconfig that "a file is good to be in SCM but ignore further changes to that file"

.gitignore can ignore files that are not in the tree, but once it is in the tree, it is tracked. :(

rolandshoemaker:
>@rolandshoemaker I haven't found a good way to express in gitconfig that "a file is good to be in SCM but ignore further changes to that file"
> 
>.gitignore can ignore files that are not in the tree, but once it is in the tree, it is tracked. :(

Ah, I thought those files could be created from scratch each run but it appears not, perhaps I'll take a look at this in another PR.

rolandshoemaker:
CI seems to have timed itself out on one of the builds.

kisom:
@rolandshoemaker restarted the build.

kisom:
Thanks for your patience (and the PR), @rolandshoemaker.

PR #855, opened 2018-02-14
Switch to dep for dependency management.

kisom:
All the cool kids are doing, we maybe should, too.


lziest:
seems like we should move to use go 1.8+ only first. aka, deprecate go1.7 syntax

cbroglie:
Addressed in #863

PR #860, opened 2018-02-20
Add ability to return precerts from Signer.Sign and create certs from precerts with new API

rolandshoemaker:
This PR makes two changes:
* Adds the ability to request a poisoned precert be returned from `Signer.Sign` instead of attempting to retrieve SCTs and incorporate them within the method. This allows callers who may have their own SCT retrieval logic to get SCTs for precerts
* Add a new method `Signer.SignFromPrecert` that accepts a poisoned precert and a list of SCTs and creates a certificate from them

`SignFromPrecert` doesn't do any template filling or checks that the certificate matches the signing profile, it only checks that the passed in precert was properly poisoned and signed by the signer CA. The rationale for this is that the caller should have checks in place to verify if they really want to turn the poisoned precert into a certificate that can be used for trust decisions (i.e. in the case where the signing profile has changed and the precert no longer matches it).

This allows users of cfssl to manage SCT retrieval for certificates themselves in cases where they may want to use a different method to the internally specified one in `Signer.Sign`.

### DIFF
--- a/errors/error.go
+++ b/errors/error.go
@@ -201,8 +201,6 @@ const (
 	// and has a invalid CT poison extension value or the extension is not
 	// critical.
 	PrecertInvalidPoison
-	// SCTInvalid occurs when attempting to parse a SCT with trailing garbage
-	SCTInvalid
 )
 
 // Certificate persistence related errors specified with CertStoreError
@@ -378,6 +376,10 @@ func New(category Category, reason Reason) *Error {
 			msg = "Certificate transparency parsing failed due to unknown error"
 		case PrecertSubmissionFailed:
 			msg = "Certificate transparency precertificate submission failed"
+		case PrecertMissingPoison:
+			msg = "Precertificate is missing CT poison extension"
+		case PrecertInvalidPoison:
+			msg = "Precertificate contains an invalid CT poison extension"
 		default:
 			panic(fmt.Sprintf("Unsupported CF-SSL error reason %d under category CTError.", reason))
 		}
--- a/errors/error.go
+++ b/errors/error.go
@@ -194,6 +194,15 @@ const (
 	// CTClientConstructionFailed occurs when the construction of a new
 	// github.com/google/certificate-transparency client fails.
 	CTClientConstructionFailed
+	// PrecertMissingPoison occurs when a precert is passed to SignFromPrecert
+	// and is missing the CT poison extension.
+	PrecertMissingPoison
+	// PrecertInvalidPoison occurs when a precert is pass to SignFromPrecert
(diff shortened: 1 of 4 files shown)


rolandshoemaker:
Test failures appear to be unrelated.

rolandshoemaker:
@kisom @lziest friendly ping, if you folks get a chance to review this it would be super useful and help unblock some work on letsencrypt/boulder, thanks!

lziest:
LGTM, but we need to give some love to the Mac build plans.

lziest:
I will let @kisom and @cbroglie to review and merge.

rolandshoemaker:
Looks like we are just waiting on a review from @kisom, any idea on an ETA?

kisom:
I can get to it this afternoon.

rolandshoemaker:
Thanks! 😄

errors/error.go
```diff
@@ -194,6 +194,13 @@ const (
 	// CTClientConstructionFailed occurs when the construction of a new
 	// github.com/google/certificate-transparency client fails.
 	CTClientConstructionFailed
+	// PrecertMissingPoison occurs when a precert is passed to SignFromPrecert
+	// and is missing the CT poison extension.
+	PrecertMissingPoison
+	// PrecertInvalidPoison occurs when a precert is pass to SignFromPrecert
```
kisom: Nit: typo.

signer/local/local.go
```diff
@@ -414,6 +418,80 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 	return signedCert, nil
 }
 
+// SignFromPrecert creates and signs a certificarte from an existing precertificate
```
kisom: Typo:  certificate

signer/local/local.go
```diff
@@ -414,6 +418,80 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
...
+// that was previously signed by Signer.ca and inserts the provided SCTs into the
+// new certificate. The resulting certificate will be a exact copy of the precert
+// except for the removal of the poison extension and the addition of the SCT list
+// extension. SignFromPrecert does not verify that the contents of the certificate
+// still match the signing profile of the signer, it simply requires that the precert
```
kisom: I'd suggest changing simply → only to clear up the language a little.

rolandshoemaker:
Pushed fixes.

rolandshoemaker:
CI seems to be unhappy about an unrelated import.

cbroglie:
I re-ran the CI build and only the known-bad OSX one failed.

PR #867, opened 2018-03-02
Expose errors from CheckSignatureFrom instead of masking

jacob-salassi:
Signature verification failures are asked behind a generic "IssuerMismatch" error which is not accurate. Instead, check the return value and propagate the error to the caller to inform the user about the true nature of the failure.


jacob-salassi:
The CI failure seems unrelated to this change, can confirm?

PR #871, opened 2018-03-05
Expose errors from CheckSignatureFrom()

jacob-salassi:
Signature verification failures are asked behind a generic "IssuerMismatch" error which is not accurate. Instead, check the return value and propagate the error to the caller to inform the user about the true nature of the failure.

### DIFF
--- a/ocsp/ocsp.go
+++ b/ocsp/ocsp.go
@@ -164,8 +164,10 @@ func (s StandardSigner) Sign(req SignRequest) ([]byte, error) {
 	if bytes.Compare(req.Certificate.RawIssuer, s.issuer.RawSubject) != 0 {
 		return nil, cferr.New(cferr.OCSPError, cferr.IssuerMismatch)
 	}
-	if req.Certificate.CheckSignatureFrom(s.issuer) != nil {
-		return nil, cferr.New(cferr.OCSPError, cferr.IssuerMismatch)
+
+	err := req.Certificate.CheckSignatureFrom(s.issuer)
+	if err != nil {
+		return nil, err
 	}
 
 	var thisUpdate, nextUpdate time.Time


jacob-salassi:
@kisom any feelings on this one?

jacob-salassi:
Any feedback on what it would take to get this merged? Maybe wrap the error in a cfssl error?

kisom:
@jacob-salassi I just haven't had the bandwidth to go over it, but I'll try to do it in the next day or two.

ocsp/ocsp.go
```diff
@@ -164,8 +164,10 @@ func (s StandardSigner) Sign(req SignRequest) ([]byte, error) {
 	if bytes.Compare(req.Certificate.RawIssuer, s.issuer.RawSubject) != 0 {
 		return nil, cferr.New(cferr.OCSPError, cferr.IssuerMismatch)
 	}
-	if req.Certificate.CheckSignatureFrom(s.issuer) != nil {
-		return nil, cferr.New(cferr.OCSPError, cferr.IssuerMismatch)
+
+	err := req.Certificate.CheckSignatureFrom(s.issuer)
+	if err != nil {
```
kisom: The should probably be

```
return nil, cferr.New(cferr.CertificateError, cferr.VerifyFailed, err)
```

kisom: Sorry, it's the tests point out that it's cferr.Wrap, not cferr.New.

jacob-salassi: No worries, i saw the build failure but had not had a chance to test it locally before I saw it. I wasnt thinking about PR refresh when I made the change from github directly ;). Update will be forthcoming.

jacob-salassi: I went with cferr.OCSPError instead of CertificateError given the pattern in the rest of this function. I think it makes sense to me, context of OCSP signing, Verify failed, then verify real error.

kisom: So I just double checked something:

* https://github.com/cloudflare/cfssl/blob/master/errors/error.go#L417 `CertificateError` will cause it to add more information to the error code (maybe not useful here).
* When we added the `OCSPError`, we didn't add it to [the default case](https://github.com/cloudflare/cfssl/blob/master/errors/error.go#L428). So, OCSP error will make cfssl panic right now.

I'll merge this, then submit a PR to fix the default case.

kisom:
Thanks for your patience on this, @jacob-salassi.

jacob-salassi:
@kisom no thanks needed, appreciate you working with me to get it merged. I'll dig deeper into future bug fixes.

PR #874, opened 2018-03-10
add -mutual-tls-san option

adamjacobmuller:
similar to -mutual-tls-cn option, but allows the regular expression to
match against any SAN on the supplied certificate

### DIFF
--- a/cli/serve/serve.go
+++ b/cli/serve/serve.go
@@ -51,8 +51,8 @@ Usage of serve:
                     [-metadata file] [-remote remote_host] [-config config] \
                     [-responder cert] [-responder-key key] \
                     [-tls-cert cert] [-tls-key key] [-mutual-tls-ca ca] [-mutual-tls-cn regex] \
-                    [-tls-remote-ca ca] [-mutual-tls-client-cert cert] [-mutual-tls-client-key key] \
-                    [-db-config db-config] [-disable endpoint[,endpoint]]
+					[-mutual-tls-san regex] [-tls-remote-ca ca] [-mutual-tls-client-cert cert] \
+					[-mutual-tls-client-key key] [-db-config db-config] [-disable endpoint[,endpoint]]
 
 Flags:
 `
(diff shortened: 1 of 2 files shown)


AdamJacobMuller:
not sure if there is any interest in merging this, but, I needed this so I added it.

kisom:
@AdamJacobMuller will try to go over the PR in the next day or two.

AdamJacobMuller:
Thanks @kisom!

cli/serve/serve.go
```diff
@@ -334,6 +334,31 @@ func serverMain(args []string, c cli.Config) error {
...
+
+		if conf.MutualTLSSANRegex != "" {
+			log.Debugf(`Requiring any SAN matches regex "%s" for client connections`, conf.MutualTLSSANRegex)
+			re, err := regexp.Compile(conf.MutualTLSSANRegex)
+			if err != nil {
+				return fmt.Errorf("malformed SAN regex: %s", err)
+			}
+			server.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
+					for _, name := range r.TLS.PeerCertificates[0].DNSNames {
```
kisom: The SAN list includes more than DNS names; if you look at the [`Certificate` structure](https://golang.org/pkg/crypto/x509/#Certificate), it has

```
        // Subject Alternate Name values
        DNSNames       []string
        EmailAddresses []string
        IPAddresses    []net.IP
        URIs           []*url.URL
```

so this should probably also check more than the DNS names.

AdamJacobMuller: Hi Kisom,

I'd considered this, but, should it?

I wasn't too sure what the use-case would be for validating based on those other attributes, but, validating based on the DNSNames is ultimately arbitrary.

If I updated to validate against all of them, do you consider this merge ready/worthy?

-Adam

kisom: @AdamJacobMuller Other than that, it looks fine.

PR #877, opened 2018-03-21
Allow tests to run on s390x

vivkong:


test.sh
```diff
@@ -45,7 +46,11 @@ COVPROFILES=""
 for package in $(go list -f '{{if len .TestGoFiles}}{{.ImportPath}}{{end}}' $PACKAGES)
 do
     profile="$GOPATH/src/$package/.coverprofile"
-    go test -race -tags "$BUILD_TAGS" --coverprofile=$profile $package
+    if [ $ARCH = 's390x'  ]; then
```
kisom: I think the logic should be

```
if [ $ARCH = 'amd64' ]
then
        go test -race -tags "$BUILD_TAGS" --coverprofile=$profile $package
else
        go test -tags "$BUILD_TAGS" --coverprofile=$profile $package
fi

kisom:
Thanks @vivkong!

PR #894, opened 2018-05-13
[WIP] Fix failing tests

cbroglie:
Changes:
- Update CA bundle with latest from https://github.com/cloudflare/cfssl_trust
- Update test hostnames to use intentionally bad ones from badssl.com, since accidentally bad ones tend to get fixed (e.g. www.capitol.state.tx.us now works)
- Update test output to make it more clear which test cases are failing

TODO:
- [x] TestBundleFromRemote
- [ ] TestBundleFromRemoteFlavor


endform:
I took a whack at seeing what was going wrong with the TestBundleFromRemoteFlavor test and if I could just point it at another badssl domain. The answer is... yes, almost. I pointed the ecc cert site to

` 	ECCCertSite            = "ecc256.badssl.com"`

This caused another error complaining about the length of the chain from the ubiquitous flavor is not 3. I stepped through the code to see why, and it's because of this code here appears to be working as designed (we hit the else block to remove the root).

```
	if flavor != Force {
		// Include at least one intermediate if the leaf has enabled OCSP and is not CA.
		if bundle.Cert.OCSPServer != nil && !bundle.Cert.IsCA && len(bundle.Chain) <= 2 {
			// No op. Return one intermediate if there is one.
		} else {
			// do not include the root.
			bundle.Chain = bundle.Chain[:len(bundle.Chain)-1]
		}
	}
```

Now whether or not you are trying to test this first branch through the original benflare.us domain, I do not know. Nor do I fully understand the semantics of the test around the ubiquitous flavor because I can't see what the original certs were set up to test for on benflare.us.

Adjusting the domain and the length assertion got the tests to pass for me. Hope this helps.

bundler/bundle_from_remote_test.go
```diff
@@ -21,14 +22,14 @@ type remoteTest struct {
...
-	bundleCallback     func(*testing.T, *Bundle)
+	errorCallback      func(*testing.T, *remoteTest, error)
+	bundleCallback     func(*testing.T, *remoteTest, *Bundle)
 }
 
 const (
 	ValidSSLSite           = "google.com"
-	SelfSignedSSLSite      = "cacert.org"
-	MismatchedHostnameSite = "www.capitol.state.tx.us"
+	SelfSignedSSLSite      = "self-signed.badssl.com"
```
kisom: Happy to see this.

cbroglie:
All green, can you imagine that?

PR #896, opened 2018-05-17
allow endpoints to be disabled

knackaron:
```
$ cfssl serve -config=config/ca.json -db-config=config/db.json -disable=init_ca,revoke -address=0.0.0.0 -port=8888 -loglevel=1 -ca=data/intermediate.pem -ca-key=data/intermediate-key.pem -responder=data/ocsp.pem -responder-key=data/ocsp-key.pem
2018/05/17 02:34:22 [INFO] Initializing signer
...
2018/05/17 02:34:22 [INFO] endpoint 'init_ca' is explicitly disabled
2018/05/17 02:34:22 [INFO] endpoint '/api/v1/cfssl/crl' is enabled
2018/05/17 02:34:22 [INFO] endpoint 'revoke' is explicitly disabled
2018/05/17 02:34:22 [WARNING] endpoint 'sign' is disabled: {"code":5200,"message":"Invalid or unknown policy"}
...
2018/05/17 02:34:22 [INFO] Handler set up complete.
2018/05/17 02:34:22 [INFO] Now listening on 0.0.0.0:8888
```

### DIFF
--- a/cli/serve/serve.go
+++ b/cli/serve/serve.go
@@ -51,7 +51,7 @@ Usage of serve:
                     [-responder cert] [-responder-key key] [-tls-cert cert] [-tls-key key] \
                     [-mutual-tls-ca ca] [-mutual-tls-cn regex] \
                     [-tls-remote-ca ca] [-mutual-tls-client-cert cert] [-mutual-tls-client-key key] \
-                    [-db-config db-config]
+                    [-db-config db-config] [-disable endpoint[,endpoint]]
 
 Flags:
 `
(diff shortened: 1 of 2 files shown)


cli/serve/serve.go
```diff
@@ -247,7 +247,9 @@ var endpoints = map[string]func() (http.Handler, error){
 func registerHandlers() {
 	for path, getHandler := range endpoints {
 		log.Debugf("getHandler for %s", path)
-		if handler, err := getHandler(); err != nil {
+		if strings.Contains(conf.Disable, path) {
```
cbroglie: Thanks @knackaron. Unfortunately there are some endpoints which are substrings of others (e.g. `crl` and `gencrl`), so this check needs to be refined. Can you instead split `conf.Disable` on commas and do an exact comparison? Something like this should work:
```go
disabled := map[string]struct{}
if conf.Disable != "" {
	for _, endpoint := range strings.Split(conf.Disable, ",") {
		disabled[endpoint] = struct{}{}
	}
}

// ...

if _, ok := disabled[path]; ok {
	log.Infof("endpoint '%s' is explicitly disabled", path)
}
```

knackaron:
Thanks for the guidance @cbroglie.  A new implementation has been pushed.

cbroglie:
Looks like the CI failures are preexisting and unrelated. Thanks again @knackaron.

Issue #834, opened 2017-12-20
HealthCheck API


evident-sbellows:
Is there an API endpoint I can use as a health check? Something I can call from a load balancer to verify that a CFSSL instance is running and functional?

wouzar:
What do you think about adding healthcheck API for ocspserve also? Because currently it's available only for API Server, and in our case we don't use it.

PR #897, opened 2018-05-22
Adds a basic healthz endpoint

cmattoon:
Addresses #834 by adding a basic health check

    $ curl -s http://localhost:8888/api/v1/cfssl/healthz
    {"success":true,"result":{"healthy":true},"errors":[],"messages":[]}

### DIFF
--- a/api/health/health.go
+++ b/api/health/health.go
@@ -1,4 +1,4 @@
-package healthz
+package health
 
 import (
 	"encoding/json"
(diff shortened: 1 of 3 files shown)


cbroglie:
@cmattoon Why are you calling it `healthz`, and not simply `health`?

cbroglie:
Thanks for the link, that's good to know. I'd like to stick with `/health` for this project - do you mind making that change? Happy to merge after that, thanks.

cmattoon:
Fixed!

cbroglie:
Thanks again, @cmattoon!

PR #905, opened 2018-07-04
Create transports with same defaults as http.DefaultTransport

cbroglie:
Critically, these defaults have sane (non-infinite) timeouts.


cbroglie:
Note the test failures are unrelated (and I started #894 to address them).

PR #911, opened 2018-07-10
Remove remnants of PKCS #11 support in README.md and comments

mahrud:
https://github.com/cloudflare/cfssl/pull/522 removed PKCS11 code vendored from `miekg/pkcs11` and `letsencrypt/pkcs11key` 2 years ago, but the README still talked about it. Those remnants are removed until they need to be added.

ps: `gokeyless` now supports PKCS11, so maybe that code can move here, since `gokeyless` already vendors `cfssl`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ services:
   - postgresql
 before_install:
   # CFSSL consists of multiple Go packages, which refer to each other by
-  # their absolute GitHub path, e.g. github.com/cloudflare/crypto/pkcs11key.
+  # their absolute GitHub path, e.g. github.com/cloudflare/crypto/pkcs7.
   # That means, by default, if someone forks the repo and makes changes across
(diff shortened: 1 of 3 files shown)


kalabiyau:
@mahrud @cbroglie my two cents on all that story with pkcs11 carve-out, if I may. Without any trace and no reliable docs on how to use HSM with cfssl - one can suffer a lot re-investing the time you already once spent on the topic ('twas supported and now cleaned with no alternative more or less). It would be great if there can be one appointed reference project with cfssl with HSM support. So that, those who are a bit paranoid about offline keys are well informed on the matter. What do you think?

mahrud:
I agree. I recently added pkcs11 support in gokeyless using the `ThalesIgnite/crypto11` package, and it would have saved me a lot of time if I knew `pkcs11key` even existed. My goal is to benchmark `crypto11` against `pkcs11key` and add the better of the two here and have gokeyless import it from cfssl.

Issue #912, opened 2018-07-10
Flag for min version of TLS for cli/serve


endform:
We run CFSSL as a server in an environment where we need to meet policy guidelines around the min version of TLS. I can add a flag "MinTLSVersion" and do some string mapping to the constants needed to pass into tls.Config, but would like to start a conversation on how to implement this in a way that will make the maintainers happy as I have not contributed to this project before.

cbroglie:
Adding a new flag to `cli.Config`, and setting the `tls.Config.MinVersion` based on it sounds like a fine approach.

endform:
@cbroglie would love it if you got a chance to review the #913, I'm open to any changes. I went back and forth about how to represent the versions as args, but should probably just change it to use 1.1 & 1.2. The other consideration I should address is how this will change when go's underlying default gets changed.

cbroglie:
Fixed in #913

PR #913, opened 2018-07-11
Add a flag to support minimum TLS version for cli/serve

endform:
Solution for #912 

* Keep 1.0 as the default to prevent breakage
* Allowed values are 12, 11 and anything else defaults to 10


cbroglie:
@endform I'll try to get to this soon. In the mean time, can you rebase off of master? The tests are now passing there

endform:
@cbroglie Will do, thanks!

PR #914, opened 2018-07-11
Performance optimization for local Signer Sign()

devinlundberg:
When the local signer does not have a dbAccessor set, It does not need to get a AuthorityKeyId (the only reason ParseCertificatePEM is called). ParseCertificatePEM is expensive with ECC since it involves doing a verification. Moving this call into the if statement means that the Sign function runs nearly twice as fast on my machine when no dbAccessor is set (it likely doesn't matter as much for RSA since verification on RSA is relatively cheap).

### DIFF
--- a/signer/local/local.go
+++ b/signer/local/local.go
@@ -391,12 +391,11 @@ func (s *Signer) Sign(req signer.SignRequest) (cert []byte, err error) {
 		return nil, err
 	}
 
-	// Get the AKI from signedCert.  This is required to support Go 1.9+.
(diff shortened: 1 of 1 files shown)


devinlundberg:
I actually misread my flame graph. This is a needless operation that does effect performance, but it doesn't remove a signing operation. Editing description to match.

kisom:
Some of our consumer services use the AKI, even if it's not used elsewhere in this codebase; we'd rather every certificate signed by this gets an AKI.

Issue #758, opened 2017-05-04
Issue with cfssl-scan 


ptownbolt:
I'm running the latest code-base for cfssl-scan on OSX and Linux and get the following error when I scan some https sites like chase.com, wellsfargo.com. However when I scan google.com and cloudflare.com there is no problem. I debugged the code and the problem seems to on the line 290 of cfssl/scan/vendor/crypto/tls/handshake_client.go

285  switch certs[0].PublicKey.(type) {
286  case *rsa.PublicKey, *ecdsa.PublicKey:
287	 break
288	 default:
289		 c.sendAlert(alertUnsupportedCertificate)
290		 return fmt.Errorf("tls: server's certificate contains an unsupported type of public key: %T", 
 certs[0].PublicKey)
	}

$ cfssl scan -scanner=TLSDial  chase.com
Scanning chase.com...
=== chase.com ===
{
  "Connectivity": {
    "TLSDial": {
      "grade": "Bad",
      "error": "tls: server's certificate contains an unsupported type of public key: *rsa.PublicKey"
    }
  }
}

$ cfssl scan -scanner=TLSDial  cloudflare.com
Scanning cloudflare.com...
=== cloudflare.com ===
{
  "Connectivity": {
    "TLSDial": {
      "grade": "Good"
    }
  }
}

umbernhard:
I'm having this same issue. I think it might be the way that the crypto package is vendored; it's out of date with its corresponding package in golang source. Trying to confirm that this is the issue. If it is, I think the right answer is to change the way the crypto package is used (un-vendor it, or use tris or something else).

phpfs:
Hi there :)

Any updates on this?
Issue still exists for me with e.g. "yahoo.de" :(

Thanks so far!

phpfs:
Updating dependencies might fix it - but will require a lot of changes in code :/

elafargue:
Indeed, the cfssl code modified the system crypto package by adding its own hooks into the package, it would require taking those customizations out, but I am afraid the crypto package does not expose what the scan utility needs...

koshatul:
Sorry, the PR won't fix this, just different issue with the crypto/tls package being overridden.

PR #915, opened 2018-07-18
Moved crypto/tls into cfssl repository

koshatul:
Fixes CloudFlare/cfssl#758

By moving it out of the committed vendor it also helps stop issues with glide and strip-vendor when using cfssl in other projects.


koshatul:
Actually, it may not fix that without merging back updates to the crypto/tls package.

But it will fix the strip-vendor issue.

jmalloc:
Hey guys, I work with @koshatul, just to provide some more context; the problem this PR addresses occurs because we have other dependencies for which we must strip the `vendor` folder. In this case our tooling (Glide) would pull those excluded dependencies into our "root" package's `vendor` folder. Because cfssl's `crypto/tls` fork is not actually vendored (as in, not available in some other repo) Glide can not import it to our root package.

PR #917, opened 2018-07-19
Moved crypto/tls into cfssl repository

cbroglie:
This is just #915 after first merging it to a branch in the repo


cbroglie:
By moving it out of vendor, the `crypto/tls` tests are now executed during the build. And they are running too slowly, causing the Travis builds to get killed. Thinking about the best way to handle.

kisom:
Needs to be rebased.

cbroglie:
Ended up disabling the tests as they weren't previously being run from the vendor folder anyways. Build is green once again 😄

PR #919, opened 2018-07-23
Helpers for handling Ed25519 keys

cjpatton:
This implements the encoding of Ed25519 keys as defined in https://datatracker.ietf.org/doc/draft-ietf-curdle-pkix-04. This is necessary for integrating Ed25519 into goKeyless.


helpers/helpers_test.go
```diff
@@ -357,6 +358,12 @@ func TestParsePrivateKeyPEM(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	testEd25519PEM, _ := ioutil.ReadFile(testPrivateEd25519Key)
```
kisom: You should check for errors, even in a test file.

helpers/derhelpers/ed25519.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil, err
+	} else if len(rest) > 0 {
+		return nil, errors.New("PrivateKey too long")
+	}
+
+	if !priv.Algorithm.Algorithm.Equal(ed25519OID) {
+		return nil, errEd25519WrongID
+	}
+
+	// TODO(cjpatton) Figure out the idiomatic way to enforce this.
```
kisom: Can you add additional info to this comment (e.g. the section of the TLS 1.3 spec) where one can find this? Once it's no longer a TODO, it'd be fine to remove but it helps anyone else who wants to work on this in the meantime.

helpers/derhelpers/ed25519.go
```diff
@@ -0,0 +1,111 @@
...
+	return asn1.Marshal(spki)
+}
+
+// ParseEd25519PublicKey returns the Ed25519 public key encoded by the input.
+func ParseEd25519PublicKey(der []byte) (ed25519.PublicKey, error) {
+	var spki subjectPublicKeyInfo
+	if rest, err := asn1.Unmarshal(der, &spki); err != nil {
+		return nil, err
+	} else if len(rest) > 0 {
+		return nil, errors.New("SubjectPublicKeyInfo too long")
```
kisom: cfssl generally uses its own internal error codes for this. You can see [this](https://github.com/cloudflare/cfssl/blob/master/helpers/derhelpers/derhelpers.go#L27) as an example in this package.

helpers/helpers_test.go
cjpatton: Skipping this check is a convention in this particular `Test*()`. Shall I change this here and above?

helpers/derhelpers/ed25519.go
cjpatton: I went ahead and resolved the TOODs.

helpers/derhelpers/ed25519.go
cjpatton: I'm not sure if cferrs is appropriate here. This function is analogous to:
https://golang.org/src/crypto/x509/x509.go?s=1352:1421#L45

This is is meant to surface low-level details about the encoding. I suppose I could use `cferr.Wrap()` to wrap these low-level errors, but I'd need to add a `PublicKeyError` category. (`CertificateError isn't really appropriate.)

cjpatton:
Ready for round two, @kisom :)

cbroglie:
@cjpatton the backoff package is available here: https://github.com/cloudflare/backoff

helpers/helpers_test.go
kisom: If it is, maybe I didn't review it --- I would have said the same thing. I think it would be better to have the error check here.

helpers/derhelpers/ed25519.go
kisom: I'm not sure either; the other function returns a `cferr`, but we can probably get by without it here. @cbroglie do you have any thoughts on this?

cjpatton:
@kisom do you want me to ping anyone else for review or are you ready to merge this? Also, I noticed that `dep ensure` removes a bunch of stuff. Want me to add a commit with this change?

helpers/helpers_test.go
cjpatton: Fixed, here and elsewhere in this file.

kisom:
@cjpatton Please add the `dep` commit, then I can merge.

helpers/derhelpers/ed25519.go
cbroglie: There's a mix of `errors` and `cferr` in helpers.go, I'm ok with this

cjpatton:
Done.

PR #942, opened 2018-11-01
Support for custom extensions in CSRs

cbroglie:


csr/csr.go
```diff
@@ -392,6 +393,10 @@ func Generate(priv crypto.Signer, req *CertificateRequest) (csr []byte, err erro
 		}
 	}
 
+	if req.Extensions != nil {
+		tpl.ExtraExtensions = append(tpl.ExtraExtensions, req.Extensions...)
```
wbl: This line appears not to be working:  tpl.ExtraExtensions = req.Extensions does.

orbli: It seems this commit is not quite same from what I have submitted... refer to https://github.com/cloudflare/cfssl/pull/939/commits/eb8547577003bbbf976eaa7585a70aeabc5e5795

cbroglie: I did make some tweaks, but logically they should be the same: https://goplay.space/#CVSEhh3UxnI. Am I missing something obvious?

badie:
@cbroglie Any ideas when this (custom extensions in CSR) might make it into master?

ralphbragg-ob:
Hi plus one for this feature request

PR #953, opened 2018-11-28
Support extensions in the generation of CSRs and signing of certificates

wbl:
This adds a change to signer that ensures we don't drop extensions on the floor when signing the certificate. Also a mysterious change required to ensure we actually output the extensions when generating the CSR.

### DIFF
--- a/csr/csr.go
+++ b/csr/csr.go
@@ -140,11 +140,12 @@ type CAConfig struct {
 // certificate request functionality.
 type CertificateRequest struct {
 	CN           string
-	Names        []Name     `json:"names" yaml:"names"`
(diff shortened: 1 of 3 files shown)


wbl:
No, you're absolutely right: the signer policy should let you limit the extensions/have a whitelist.

orbli:
If I remember right I worked on this feature when `cfssl` provides a function to generate both csr and crt by feeding a single json file. If the csr and sign parts are separated (as i read readme of cfssl now) then it looks like more design is needed, as you might need to tell cfssl the limit/whitelist policy when signing...... =.=

Issue #921, opened 2018-07-27
Adding support for URI SANs


jacobcase:
RFC 5280 specifies support for URI SANs along with the existing supported types (DNS, IP, Email). Unfortunately Go did not make this easy until the 1.10 release.  It is still possible to support it in a backwards compatible way by marshaling the asn1 data manually.

I would like to add support for this feature to CFSSL and submit a PR. Before I do, I want to make sure it is work that would be accepted, and that I implement it in a way everyone is happy with. That leaves me with a few questions.

1) Is this an acceptable addition to cfssl?
2) Should it still fall under `hosts` in the CSR JSON?
3) Should it still be compatible with go 1.8, is it unreasonable to require go 1.10?

Example URI fields:
`https://cfssl.org/`
`urn:some:resource`

Thank you for any feedback.

RFC: https://tools.ietf.org/html/rfc5280#section-4.2.1.6

vjanelle:
Took a stab at this, but someone from cloudflare will have to make a decision about removing older versions of golang from the travis tests.

cbroglie:
Fixed by #957

PR #957, opened 2018-12-13
(#921) URI SAN records

vjanelle:
Support URI san records.

### DIFF
--- a/csr/csr.go
+++ b/csr/csr.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"net"
 	"net/mail"
+	"net/url"
 	"strings"
 
 	cferr "github.com/cloudflare/cfssl/errors"
@@ -268,6 +269,9 @@ func getHosts(cert *x509.Certificate) []string {
 	for _, email := range cert.EmailAddresses {
 		hosts = append(hosts, email)
 	}
+	for _, uri := range cert.URIs {
+		hosts = append(hosts, uri.String())
+	}
 
 	return hosts
 }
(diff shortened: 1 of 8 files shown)


vjanelle:
As kind of expected from #921  - Older versions of golang don't have URL support in the cypto/tls support.

cbroglie:
Thanks @vjanelle! I don't see any reason to support more than the most recent 2 versions of Go, so you can go ahead and update the Travis configuration as part of this change. And would you mind updating the README as well?

csr/csr.go
```diff
@@ -379,6 +383,9 @@ func Generate(priv crypto.Signer, req *CertificateRequest) (csr []byte, err erro
 			tpl.IPAddresses = append(tpl.IPAddresses, ip)
 		} else if email, err := mail.ParseAddress(req.Hosts[i]); err == nil && email != nil {
 			tpl.EmailAddresses = append(tpl.EmailAddresses, email.Address)
+		} else if uri, err := url.ParseRequestURI(req.Hosts[i]); err == nil && uri != nil {
+			log.Info("Found a URI: ", uri)
```
cbroglie: Please remove this log as well

vjanelle: oops.

vjanelle:
@cbroglie done.

vjanelle:
@cbroglie any opinions on if a squashed commit or not?

cbroglie:
Separate commits is good here. Thanks again!

PR #959, opened 2018-12-20
Check IP Addresses in name_whitelist

raimund-s:
The name_whitlist in the cfssl ca config ignores IP addresses completely therefore any IP could be requested by the client even if a white list was used on the server side.

All the groundwork seems to be in there yet an IP is never checked.


raimund-s:
The tests here seem to work differently from the local ones, therfore I can not reasonably test it and will close the MR

PR #966, opened 2019-01-29
Add missing sqlite import

andrewheberle:
Ensure that go-sqlite3 is imported so sqlite based databases work properly.

This ensures that sqlite3 based certificate databases work properly.


andrewheberle:
Builds now pass after the recent PR removing go1.10 from TravisCI.

Also my initial comment did not make it clear that this is so that the sqlite db backend can be used for "multirootca"

cbroglie:
Thanks @andrewheberle

Issue #929, opened 2018-09-07
PKCS11 support


adbias:
Hello, I'm using SoftHSM and I don't know if there is any documentation, or if someone already implemented a PKCS11 module for this project. I'm using the latest version of CFSSL (1.3.2). I know there was support for an old version and all of the PKCS11 support was moved to another repository, but I want to know if there is going to be official docs for this.

cbroglie:
There is no support for PKCS#11 currently, though pull requests are welcome. We've had success using the [ThalesIgnite/crypto11](https://github.com/ThalesIgnite/crypto11 ) package for PKCS#11 support in our [gokeyless](https://github.com/cloudflare/gokeyless) project.

techninja1008:
I'm interested in implementing this feature, at least in the `multirootca` binary. Currently, I've managed to hack together support based on ThalesIgnite/crypto11 for it, and I am able to successfully issue certificates signed with a key that is stored on a Yubikey 4 (As I have one anyway, and it's an easy test device).

@cbroglie I would like to seek your direction on how this feature should properly be integrated and contributed back to this repository.

cbroglie:
@techninja1008 Thanks for your contribution, and I'll be happy to work with you to get it merged. I should be able to take a look at this next week.

PR #977, opened 2019-03-16
[WIP] Add a PKCS#11 support for multirootca command.

techninja1008:
Relates to #929.

Text from commit message:

```
Add recognition of a `pkcs11` protocol in the URI parsing for the private key in the multirootca
config. Add other configuration options when this is used in order to identify the correct token
for the server to use, and then return a crypto.Signer that uses a Private Key located on the
token by its ID. Also add recognition for crypto11.PKCS11PrivateKeyRSA and
crypto11.PKCS11PrivateKeyECDSA in the multirootca binary as valid private key types.

Signed-off-by: Danny Wensley <danny@wensley.eu>
```


cbroglie:
@techninja1008 what are the atomicity issues you are encountering?

techninja1008:
There were a few bugs in a dependency of ThalesIgnite/crypto11 relating to atomic updating of unaligned fields on some structs, but those have now been fixed. I am currently waiting for ThalesIgnite/crypto11 to revendor, but I have also discovered another issue on ARM relating to part of the CGO interface - which I unfortunately don't have experience in - where a double free is occuring, but only on ARM, even when using the same `go` binary (cross-compiling to ARM) and the same version of gcc (7.3.0)

asyd:
Any news by chance? I'm really interest by PKCS#11 support in cfssl

Cheers

Benvandamme:
I would appreciate this to see this feature!

PR #891, opened 2018-05-06
Add SANs support for self signed certificate

ashphy:
Fixed #777 

Currently the `cfssl selfsign` command ignores `hosts` config, and produces certificate excluding Subject Alternative Names.

Now modern browsers require to inlclude SANs in the certificate, common name is no longer supported.
Self-signed certificate should also include SANs inside.

PR #983, opened 2019-03-28
Add support to lookup certificate by s/n and authkey. Closes #982

### LINKED ISSUE #982: Lookup certificate by serial number
Thanks for a great product!

In the setup we're using we are issuing short lived certificates to use for signing within the system. To verify the signature we're passing the serial number (and authority key id) since we, for several reasons, don't want to send the full certificate (or certificate chain). We then want to leave it up to the receiving party to fetch the certificate from CFSSL for verification.

I was a bit surprised that even though all the issued certificates and revocations are stored in the accessor database there was no API endpoint/CLI exposing the `GetCertificate()` function from the accessor interface.

Since I think it's a quite common usage to fetch the certificate based on the serial number I think it's reasonable to add support for it via the API and CLI.

I did a proposal of the implementation and will submit a PR shortly.

bombsimon:
This PR is a proposal to solve #982 which aims to implement support to lookup a certificate by serial number and authority key id.

The implementation does not break the API which as of today does **not** require an `certdb.Accessor` to run. Instead the CLI tool will yield an error like other tools if `db-config` is not specified and `dbconf.DBFromConfig()` fails.

If the server is being started without `db-config` flag the same handler for the certinfo API will be created with `certinfo.NewHandler()`. However, if there's an available DB the handler will be created with `certinfo.NewAccessorHandler()` which then supports querying the database.

I'm not sure if there's any need to keep the backwards compatibility but I guess it's unnecessary to remove support for those certinfo features that does not require an underlying data storage.

I only made the bare minimum changes for this implementation on purpose so even though there are a few low hanging fruits to do some cleanup I tried to avoid that.

And of course I added a new test for the new feature. Sadly there seems to be lacking tests for the other certinfo functions but I didn't bother to add those in this PR.


bombsimon:
Hmm, not sure why the Travis job is failing. It was failing at my personal account as well but since I couldn't figure out what was going on I thought it was related to the repository path not including `cloudflare/cfssl`.

These are the tests failing, and they're only failing under Linux 1.10.x. Linux 1.11.x is working and so is master under macOS. I run 1.12 locally on my mac where the tests also work.

```sh
-- FAIL: TestBundleFromFile (1.41s)
	bundle_from_file_test.go:346: expected no error. but an error occurred: {"code":1220,"message":"x509: certificate signed by unknown authority"}

--- FAIL: TestUbiquitousBundle (0.06s)
	bundler_test.go:301: Optimal bundle failed the chain length test. Chain length: 3

--- FAIL: TestUbiquityBundleWithoutMetadata (0.06s)
	bundler_test.go:338: Ubiquitous-fall-back-to-optimal bundle failed

--- FAIL: TestECDSAWarning (0.00s)
	bundler_test.go:741: Optimal bundle failed: {"code":1220,"message":"x509: certificate signed by unknown authority"}
```

Will try to look into those errors if no one beats me to it.

bombsimon:
Seems to be something with go 1.10.x, maybe something regarding the certificate pool in standard library.

```sh
go test ./bundler/. -run TestECDSAWarning
ok      github.com/cloudflare/cfssl/bundler     0.100s
```

```sh
GOROOT=~/go1.10.8 PATH=~/go1.10.8/bin/:$PATH go test ./bundler/... -run TestECDSAWarning
2019/03/28 19:53:13 [INFO] bundling certificate for CN=cloudflare-inter.com,OU=Systems Engineering,O=CloudFlare,L=San Francisco,ST=California,C=US
--- FAIL: TestECDSAWarning (0.00s)
        bundler_test.go:741: Optimal bundle failed: {"code":1220,"message":"x509: certificate signed by unknown authority"}
FAIL
FAIL    github.com/cloudflare/cfssl/bundler     0.025s
```

Also failing at `master` and in #891. I guess it would be reasonable to drop support for 1.10 and support 1.11 and 1.12 instead since 1.10 is now archived.

laverya:
@bombsimon my PR to drop go 1.10 support (and add 1.12 tests) has been merged - you should be able to rebase and have things pass now

bombsimon:
Sorry for force-pushes but I had some naming and spelling errors and didn't want to make the git history messy. Twice. 😔

Latest change uses the same flags and naming convention for serial and authority key id as the revoke API which feels better.

cli/certinfo/certinfo.go
```diff
@@ -21,12 +24,14 @@ Usage of certinfo:
         cfssl certinfo -csr file
 	- Data from certificate from remote server.
         cfssl certinfo -domain domain_name
+    - Data from CA storage
```
cbroglie: Whitespace looks slightly off here

cli/certinfo/certinfo.go
```diff
@@ -66,8 +71,21 @@ func certinfoMain(args []string, c cli.Config) (err error) {
 		if cert, err = certinfo.ParseCertificateDomain(c.Domain); err != nil {
 			return
 		}
+	} else if c.Serial != "" && c.AKI != "" {
```
cbroglie: Seems like we should check `c.DBConfigFile != ""` here as well? Otherwise the user would just see an error about failing to load an empty string?

cli/certinfo/certinfo.go
bombsimon: Oh, such a rookie mistake! 🤦‍♂️ Didn't see the mix of tabs and spaces. Will fix!

cli/certinfo/certinfo.go
bombsimon: Yeah you're absolutely right! I'll add the same message as when `-db-config` is missing while doing revoke to be consistent.

PR #981, opened 2019-03-27
update certificate-transparency-go, protobuf and mattn/go-sqlite3

laverya:
This fixes compilation errors on go 1.11+ and GOOS=windows

[replicatedhq/ship](https://github.com/replicatedhq/ship) imports this, and our goreleaser build failed after updating to go 1.12. Overriding the dependency in Gopkg.toml works https://github.com/replicatedhq/ship/pull/892/commits/faffe1c30141f613df1d25cd313d94fb2814eeee, but presumably we will not be the only project to encounter this issue.

PR #984, opened 2019-03-28
include go 1.12 in Travis CI and drop support for go 1.10

laverya:
1.12 is the latest stable version of Golang and should be supported, whereas go 1.10 is now archived

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: go
 
 matrix:
   include:
-    - go: 1.10.x
     - go: 1.11.x
     - go: 1.12.x
     - go: master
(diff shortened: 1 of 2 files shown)


cbroglie:
Thanks! Will merge after CI passes

laverya:
Apologies for the force push - I'm not sure what went wrong with that code signature

cbroglie:
The Travis build won't seem to run for this PR, not sure if it's related to the force push or what. We can keep waiting, or you could try another PR with the same changes.

laverya:
I think it's just waiting in a queue somewhere - previous force-pushes I've done (in #981 for instance) have been picked up and built by travis

cbroglie:
Of course it starts running as soon as I give up on it 😄

PR #986, opened 2019-04-04
feat(ocsp): add parameter to control max-age of OCSP responses

### LINKED ISSUE #623: Parameter to control OCSP response `Expires` header
The OCSP responder's `ServeHTTP` handler presently [sets the `Expires` cache control header to the OCSP response's `NextUpdate` value](https://github.com/cloudflare/cfssl/blob/437c6213e569dde639fd116272c27557fd38eaf9/ocsp/responder.go#L197) as recommended by [RFC 5019 §6.2](https://tools.ietf.org/html/rfc5019#section-6.2).

At Let's Encrypt we are generating fresh OCSP responses ahead of the `NextUpdate` value (generally ~3 days) to allow some wiggle room. With the `Expires` header presently set from the responder this strategy is not effective as our content distribution network will continue to honour the cached response until right up to the `NextUpdate`. Therefore we override the `Expires` header at the CDN layer to a shorter period ~12h after the current date so clients will pick up the new OCSP response before the last minute.

It would be better if we could instead set a new `maxMaxAge` parameter of ~3days in the OCSP Responder configuration. The responder's `ServeHTTP` could then be modified to set the `Expires` header to the  min of the `NextUpdate` and the `maxMaxAge` parameter. This would remove our need to override the `Expires` header later-on.

glb:
The previous behaviour of the OCSP responder would set the `Expires` and `Cache-Control` headers for OCSP responses to the `NextUpdate` value. This caused some providers to experience problems with their CDN caching the responses up until the last possible moment, which was undesirable.
    
This change allows users to set a max-age for OCSP responses to address this issue.

Fixes #623.

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -69,7 +69,8 @@ type Config struct {
 	AKI               string
 	DBConfigFile      string
 	CRLExpiration     time.Duration
-	Disable     	  string
+	Disable           string
+	MaxAge            time.Duration
 }
 
 // registerFlags defines all cfssl command flags and associates their values with variables.
@@ -132,6 +133,7 @@ func registerFlags(c *Config, f *flag.FlagSet) {
 	f.DurationVar(&c.CRLExpiration, "expiry", 7*helpers.OneDay, "time from now after which the CRL will expire (default: one week)")
 	f.IntVar(&log.Level, "loglevel", log.LevelInfo, "Log level (0 = DEBUG, 5 = FATAL)")
 	f.StringVar(&c.Disable, "disable", "", "endpoints to disable")
+	f.DurationVar(&c.MaxAge, "max-age", 0, "max-age for OCSP responses")
 }
 
 // RootFromConfig returns a universal signer Root structure that can
--- a/cli/ocspserve/ocspserve.go
+++ b/cli/ocspserve/ocspserve.go
@@ -15,13 +15,13 @@ import (
 var ocspServerUsageText = `cfssl ocspserve -- set up an HTTP server that handles OCSP requests from either a file or directly from a database (see RFC 5019)
 
   Usage of ocspserve:
-          cfssl ocspserve [-address address] [-port port] [-responses file] [-db-config db-config]
+          cfssl ocspserve [-address address] [-port port] [-responses file] [-db-config db-config] [-max-age duration]
 
   Flags:
   `
 
 // Flags used by 'cfssl serve'
-var ocspServerFlags = []string{"address", "port", "responses", "db-config"}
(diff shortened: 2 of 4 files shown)


glb:
@cpu have you had a chance to look at the implementation? Is this what you were looking for?

cpu:
@glb I admit I totally forgot about #732. I agree with the reasoning in #732 that it would be nicer to expose a richer interface that can return `http.Header`s generically. Its also a larger change and there didn't see to be any traction/input from the maintainers for this in 2017. If the smaller implementation you have in this PR is likely to be accepted I think it could still be useful to us and I don't want to block the good on the perfect.

@jsha @rolandshoemaker What do you think?

cpu:
>  there didn't see to be any traction/input from the maintainers for this in 2017. 

I just can't read today :sweat: It was merged in https://github.com/cloudflare/cfssl/pull/772/files and so maybe this PR isn't required anymore for the cache-control usage we had in mind. I'll check internally.

jsha:
Thanks for working on this, @glb! To be clear, when you say "some providers" are you talking about Let's Encrypt, or have other providers run into this too?

Right now we work around this by telling our CDN to ignore the caching headers and always cache for 12 hours. That's not ideal. However, there's a bit of complexity: [RFC 5019](https://tools.ietf.org/html/rfc5019#page-10) specifies what value Expires SHOULD have:

> expires         Specifies how long the response is considered fresh.
>                    This date and time will be the same as the nextUpdate
>                    timestamp in the OCSP response itself.

More generally, Expires / Max-Age are doing double duty:

1. Telling the CDN how long to cache, and
2. Telling the rest of the world how long to cache.

We'd like to keep the full cache period for (2), but for (1) use a shorter period. CDN providers appear to have special internal caching headers to deal with exactly this scenario, and that's our [current plan to approach this issue](https://github.com/letsencrypt/boulder/issues/2794). But as you can see, it hasn't been a top priority.

So, long and short, we probably wouldn't use this feature exactly as written, but I don't want to stand in the way if it's important for another use case.

glb:
@jsha I just wrote the PR specifically for #623 since it was marked as "help wanted", open, and I had some time on my hands... if this specific implementation doesn't meet your needs and you have another one on the go, then that's cool. If you aren't looking for a solution to #623 as written any more it might be worthwhile to close it 🙂 and I'll close this PR.

jsha:
Yep, will close. Sorry for leaving it open so long even though it wasn't needed anymore! :heart:

cpu:
> Yep, will close. Sorry for leaving it open so long even though it wasn't needed anymore! heart

Agreed! Sorry I dropped the ball on that!

PR #1007, opened 2019-06-14
ocsp: Add debug level request logging

### LINKED ISSUE #1004: OCSP Responder request logs
(@cbroglie I think you're the current lead maintainer, so tagging you for thoughts on this)

At LE we'd like to add OCSP request logs to our OCSP responder for debugging purposes. Given we use `cfssl/ocsp.Responder` we only actually get information about the OCSP request itself, rather than the HTTP request. Instead of double parsing the OCSP request in a top level mux hack I'd rather the logging happen at the `cfssl/ocsp` level.

This is slightly complicated for us though as we don't want to _always_ collect this request information as our privacy policy indicates we will only ever collect request information for OCSP requests (i.e. IP, UA, etc) for debugging purposes. This means it'd need to either be enabled by logging at a low syslog level (presumably `LOG_DEBUG`) or by explicitly enabling it as a bool in the `Responder` constructor. Do you have strong opinions on which way you'd prefer this to be done from the CF perspective? I'd lean towards having an explicit bool as it'd be harder to accidentally enable, but either would work for us.

rolandshoemaker:
Sorry for the lag on this. I think this hits most of the information that we would be looking to collect. Adding JSON marshaling to every request feels a bit excessive but it avoids having to do a bunch of more complicated string building (or logging lots of empty fields) and I doubt the added overhead will be that much of a burden.

Fixes #1004.

### DIFF
--- a/ocsp/responder.go
+++ b/ocsp/responder.go
@@ -8,9 +8,11 @@
 package ocsp
 
 import (
+	"crypto"
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


rolandshoemaker:
One thought: Perhaps it would be worthwhile to check `cfssl/log.Level` and only do the marshaling/logging if it is at `LevelDebug`.

cbroglie:
I expect the performance impact to be negligible, so I'd prefer not to introduce a conditional code path. My fear is some future refactor inadvertently introduces a crash that only occurs at a non-default log level.

PR #1013, opened 2019-06-27
Fix yaml support for CertificateRequest objects, convert KeyRequest into concrete class

ferringb:
Two main things here:
* add explicit yaml tag's for the various objects of a CertificateRequest.  yaml.v2 requires explicit tagging for this sort of thing, thus add it to make it happy.
* Drop the KeyRequest interface and instead just rename BasicKeyRequest as it.  Asking yaml.v2 to unmarshal through that interface triggers a panic- unlike json, it can't see through to the underlying type thus unhappily panics in that case.
  Nothing in the codebase provides an alternative implementation of KeyRequest, thus- via discussion with @cbroglie - it seems best to just drop the interface altogether.

### DIFF
--- a/api/generator/generator.go
+++ b/api/generator/generator.go
@@ -115,7 +115,7 @@ func (g *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
 	r.Body.Close()
 
 	req := new(csr.CertificateRequest)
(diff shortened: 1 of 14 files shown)


ferringb:
@cbroglie confirmed from my end, thanks

Issue #1008, opened 2019-06-19
Pre-issuance certificate linting support.


cpu:
Hi folks,

At Let's Encrypt we would like to build optional pre-issuance certificate linting into CFSSL. Pre-issuance linting is repeatedly cited as a best-practice in incident post-mortems in the web PKI community. Adding support to CFSSL will help interested parties add more safe-guards to their issuance pipeline. I'd be happy to commit to doing the work provided the proposed design outlined below sounds good. 

[Boulder](https://github.com/letsencrypt/boulder) already uses [zlint](https://github.com/zmap/zlint), a Go based X.509 Certificate Linter, as a library for post-issuance linting. I think it makes sense to integrate this project as a CFSSL dependency to accomplish pre-issuance linting.

## High Level

At a high level we would like to:

* Add a dependency on `zlint`
* Update the `config.SigningProfile` struct to accept two new optional fields:
  1. `lintErrLevel int`
  2. `ignoredLints []string`
* Update the `signer.local.Signer` struct to generate one throw-away private key at creation for linting.
* Update the `signer.local.Signer`'s `sign(*x509.Certificate)` function to:
  1. Sign the TBSCert template with the Signer's throw-away key.
  2. Pass the the resulting cert to zlint for linting.
  3. Handle any lint results above the lintErrLevel as errors (iff the lint is not in the ignoredLints list)

## Adding a dep.

We think integrating `zlint` directly so it can work out-of-box with a config change rather than building in a callback for the same purpose and supplying the linting externally on a project-specific basis will help more folks benefit from the linting. zlint has a bias towards CA/B reqs and RFC 5280 but it is an extensible way to add additional lints and many of the existing lints vet best-practices that are broadly applicable.

If it helps you feel better about a new dependency I'm also a committer/maintainer of `zlint` so you would have a contact on that side of the fence from the get-go.

## Config changes

The `lintErrLevel` will default to 0 for existing configs. We'll match this up with the `zlint.LintResult.LintStatus` in the local signer to know when to treat a result as an error. Any `LintStatus` > the `lintErrLevel` will be an error, except in the case where `lintErrLevel` is zero and linting is disabled.

* 0 = `zlint.lints.Reserved` -> no linting will be done. no results considered. Existing CFSSL behaviour.
* 1, 2 = `zlint.lints.NA` and `zlint.lints.NE` -> these don't really make sense in this context.
* 3 = `zlint.lints.Pass` -> all lint results except pass will be considered errors.
[...]

PR #1015, opened 2019-07-01
local signer: pre-issuance linting support with zlint.

cpu:
See #1008 for more background/design.

* 797ea50 adds a dependency on `github.com/zmap/zlint`. The tip of master is vendored.
* f02ee2f adds the `lint_error_level` and `ignored_lints` configuration to `config.SigningProfile`.
* 787a454 updates `config.SigningProfile.populate` to create a bool map from the `ignored_lints`.
* 66f7d5d updates the `signer.local.Signer` to generate a linting private key when required.
* 0434ea0 updates the `signer.local.Signer` to lint certificates before signing them.
* a4b817c uses the native `lints.LintStatus` type instead of `int` in config/parameters.

Points of note for reviewers:
* As promised the default behaviour/existing configurations will perform no linting. It is opt-in by way of the `lint_error_level` being > 0.
* A new concrete `signer.local.LintError` error type is returned from `Sign` when there is a linting error. I originally implemented this so that `Sign` used `errors.Wrap` with a new error code range but this made it impossible to return all of the lint results through to library users except by concatinating them into one giant message. Using the error type makes it easier to handle the lint errors specifically in calling code.
* Only `Sign` performs linting when configured. `SignFromPrecert` can not lint the TBS Certificate with SCTs created from the precert because it has no link to the profile that created the precert. Addressing this would require some significant API rework and I think the advantages are small.
* I successfully integrated this branch with a WIP [Boulder](https://github.com/letsencrypt/boulder) branch to test end-to-end.

@cbroglie, @lgarofalo PTAL when you have a chance. Thanks! 

Resolves https://github.com/cloudflare/cfssl/issues/1008


config/config.go
```diff
@@ -104,6 +117,9 @@ type SigningProfile struct {
 	NameWhitelist               *regexp.Regexp
 	ExtensionWhitelist          map[string]bool
 	ClientProvidesSerialNumbers bool
+	// IgnoredLintsMap is a bool map created from IgnoredLints when the profile is
+	// loaded. It facilitates set membership testing.
+	IgnoredLintsMap map[string]bool
```
jsha: This should probably be unexported since it's an internal implementation detail, right?

signer/local/local.go
```diff
@@ -29,14 +32,21 @@ import (
...
 // Signer contains a signer that uses the standard library to
 // support both ECDSA and RSA CA keys.
 type Signer struct {
-	ca         *x509.Certificate
-	priv       crypto.Signer
+	ca   *x509.Certificate
+	priv crypto.Signer
+	// lintPriv is generated randomly when pre-issuance linting is configured and
+	// used to sign TBSCertificates for linting.
+	lintPriv   crypto.Signer
```
jsha: Since this is an internal implementation detail, we have the luxury of explicitly typing it (`*ecdsa.PrivateKey`). That saves a `switch` later on, and also makes it a little clearer what the field does when reading this part of the code.

cpu: I thought it was slightly cleaner to use an interface/type switch to make it easier to use different key types in the future. It's _probably_ overkill, I'd be open to changing it to the explicit type if folks feel strongly.

config/config.go
cpu: It needs to be exported to be used by the `signer/local` package. I could move the map out of the profile config to an unexported field of the `local.Signer` type alongside `lintPriv` but it will be more complicated because unlike the private key there can be multiple profiles per `Signer` and each can have its own configured ignore map. Overall I'd prefer to keep the map in the `SigningProfile` as an exported field. This also matches how other constructed fields are handled in `SigningProfile.populate`.

cpu:
:wave: @cbroglie, @lgarofalo Do you folks think you will have a chance to give this PR a look in the next week? No troubles if it isn't possible, I'm just trying to get a sense to balance other commitments so I can be available for iteration. Thanks!

cbroglie:
I should be able to get to it today or tomorrow, sorry for the delay

cpu:
@cbroglie No problem at all! Thanks for the update.

config/config.go
```diff
@@ -90,6 +90,19 @@ type SigningProfile struct {
...
 	CertStore           string       `json:"cert_store"`
+	// LintErrLevel controls preissuance linting for the signing profile.
+	// 0 = no linting is performed [default]
+	// 2..3 = reserved
+	// 3 = all lint results except pass are considered errors
+	// 4 = all lint results except pass and notice are considered errors
+	// 5 = all lint results except pass, notice and warn are considered errors
+	// 6 = all lint results except pass, notice, warn and error are considered errors.
+	// 7 = lint is performed, no lint results are treated as errors.
+	LintErrLevel int `json:"lint_error_level"`
```
cbroglie: What do you think about using https://godoc.org/github.com/zmap/zlint/lints#LintStatus directly here?

cpu: Good idea, thanks! Implemented in a4b817c

cpu:
@cbroglie Is there anyone else that needs to give this PR a once-over before it can be merged?

cpu:
@cbroglie Thanks!

cpu:
I discovered there was a slight bug with adding a `zlint.lints.LintStatus` field to the CFSSL signer config object: Prior to [zlint 9971d62](https://github.com/zmap/zlint/commit/9971d62266e74547157ff95b5413227e21d8fe23) this type could only be marshaled to JSON, but not unmarshaled.

If you marshal a CFSSL config (with or without setting the new `LintErrLevel` field) then unmarshaling it produces an error like:
> Failed to create CA: {"code":5200,"message":"failed to unmarshal configuration: json: cannot unmarshal string into Go struct field SigningProfile.lint_error_level of type lints.LintStatus"}

The fix is simply to update your `zlint` package. This allows the `lints.LintStatus` type to be unmarshaled without error.

cpu:
> The fix is simply to update your zlint package. This allows the lints.LintStatus type to be unmarshaled without error.

I realized that this also needed to be fixed in the `Gopkg.lock` and vendored deps: https://github.com/cloudflare/cfssl/pull/1025

PR #1021, opened 2019-07-23
Have the API sign endpoint accept -expiry argument

abezard:
Currently the `sign` API endpoint doesn't accept any `expiry` argument.

This is an attempt to fix it. I believe it's not possible to pass any time dimension into the CSR which means that when a user hits the `sign` API endpoint to have its certificate signed, the `Expiry` value used to sign that certificate will always be the one configured on the server's side.

This code makes sure the user can't overwrite the server's configuration if the `expiry` value passed as an argument is greater than the one configured on the server's side.

### DIFF
--- a/api/signhandler/signhandler.go
+++ b/api/signhandler/signhandler.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"math/big"
 	"net/http"
+	"time"
 
 	"github.com/cloudflare/cfssl/api"
 	"github.com/cloudflare/cfssl/auth"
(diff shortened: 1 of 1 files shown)


api/signhandler/signhandler.go
```diff
@@ -127,6 +139,27 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) error {
 		return errors.NewBadRequestString("Unable to parse sign request")
 	}
 
+	if req.NotAfter.IsZero() {
```
abezard: right now `req.NotAfter` / `req.NotBefore` will always be 0 as the user doesn't have the ability to update it via his POST but I still put that piece here for future implementation (in case in the future we want users to be able to post `req.NotAfter` / `req.NotBefore` values and we want `req.NotAfter` to have the priority over `Expiry`).

PR #1022, opened 2019-07-25
Fix formatted logging calls.

QuLogic:
These pass format strings, but don't call the formatting version of the function.


cbroglie:
Thanks!

PR #1025, opened 2019-08-06
Gopkg: update zlint/zcrypto to latest.

cpu:
This is a follow-up to https://github.com/cloudflare/cfssl/pull/1015

There was a slight bug with adding [a `zlint.lints.LintStatus` field to the CFSSL signer config object](https://github.com/cloudflare/cfssl/blob/633726f6bcb7574626ae05ae72ca3c8dbc51810f/config/config.go#L94-L102): Prior to [zlint](https://github.com/zmap/zlint) commit [9971d62](https://github.com/zmap/zlint/commit/9971d62266e74547157ff95b5413227e21d8fe23) this type could only be marshaled to JSON, but not unmarshaled.

As a result if you marshal a CFSSL config (with or without setting the new `LintErrLevel` field) then unmarshaling it produces an error like:

```
Failed to create CA: {"code":5200,"message":"failed to unmarshal configuration: json: cannot unmarshal string into Go struct field SigningProfile.lint_error_level of type lints.LintStatus"}
```

Updating to `github.com/zmap/zlint` >= [9971d62](https://github.com/zmap/zlint/commit/9971d62266e74547157ff95b5413227e21d8fe23) resolves the problem.

This also has the tangential benefit of removing an indirect dependency on `github.com/asaskevich/govalidator` (as a result of some cleanup work done upstream: https://github.com/zmap/zlint/pull/294, https://github.com/zmap/zcrypto/pull/199)


cpu:
@cbroglie It looks like CI is failing because of `zlint` using `strings.ReplaceAll`, which I didn't realize wasn't available in Go 1.11.x.

Would you rather that I:
A) close this PR and fix Go 1.11.x compatibility upstream in `zlint`
B) update CFSSL to use Go 1.12.x

My light preference is to pursue option B because both Boulder and Zlint are both on 1.12.x already. WDYT?

cbroglie:
Since Go 1.13 is nearing release, I'm OK with updating the minimum supported version to 1.12.

cpu:
@cbroglie Sounds good, thanks. I'll open a second PR for that momentarily.

cpu:
@cbroglie This branch is ready for review when you have a chance. Thanks!

cpu:
Thanks!

PR #1027, opened 2019-08-07
project: require Go 1.12.x

cpu:
Dependencies are starting to require Go 1.12.x, Go 1.13 is near ready, and the project maintainers are comfortable dropping support for 1.11.x.

Ref: https://github.com/cloudflare/cfssl/pull/1025#issuecomment-519195077

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@ language: go
 
 matrix:
   include:
-    - go: 1.11.x
-    - go: 1.12.1
+    - go: 1.12.7
     - go: master
       os: osx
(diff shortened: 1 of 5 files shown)


cpu:
I'm not able to reproduce the test failures locally with `./test.sh` but I'm a couple point releases behind 1.12.7 on my dev system. I'm going to update and see if I can repro.

cbroglie:
Looks like it was just a transient network issue, kicked the build and it passed.

cpu:
Great, I was able to see tests pass locally with 1.12.7 as well.

PR #1035, opened 2019-10-11
Database support added for CFSSL in remote mode

pboguslawski:
CFSSL does not create cert record in SQL database if
signed in remote mode (i.e. using multirootca). With
this mod all certs will be saved id DB regardless of
signing mode (local/remote).

Author-Change-Id: IB#1094762


pboguslawski:
Obsoleted by #1042

PR #1048, opened 2019-10-31
Support building/releasing binaries with goreleaser

cbroglie:
Since this project requires cgo, the builds are performed in a
container which includes the MinGW compiler (for Windows) and an OSX
SDK.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,3 +30,23 @@ bin/goose: $(shell find . -type f -name '*.go')
 .PHONY: clean
 clean:
 	rm -rf bin
+
+# Check that given variables are set and all have non-empty values,
+# die with an error otherwise.
+#
+# Params:
+#   1. Variable name(s) to test.
+#   2. (optional) Error message to print.
+#
+# cf: https://stackoverflow.com/questions/10858261/abort-makefile-if-variable-not-set
+check_defined = \
+	$(strip $(foreach 1,$1, \
+		$(call __check_defined,$1,$(strip $(value 2)))))
+__check_defined = \
+	$(if $(value $1),, \
+		$(error Undefined $1$(if $2, ($2))))
+
+.PHONY: release
+release:
+	@:$(call check_defined, GITHUB_TOKEN)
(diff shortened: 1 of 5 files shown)


cli/version/version.go
```diff
@@ -8,17 +8,9 @@ import (
...
-	Minor    int
-	Patch    int
-	Revision string
-}{1, 3, 4, "release"}
-
-func versionString() string {
-	return fmt.Sprintf("%d.%d.%d", version.Major, version.Minor, version.Patch)
-}
+var (
+	version = "dev"
```
cbroglie: This will now be set to the latest git tag at build time

Makefile
```diff
@@ -30,3 +30,23 @@ bin/goose: $(shell find . -type f -name '*.go')
...
+	$(strip $(foreach 1,$1, \
+		$(call __check_defined,$1,$(strip $(value 2)))))
+__check_defined = \
+	$(if $(value $1),, \
+		$(error Undefined $1$(if $2, ($2))))
+
+.PHONY: release
+release:
+	@:$(call check_defined, GITHUB_TOKEN)
+	docker run -e GITHUB_TOKEN=$(GITHUB_TOKEN) --rm  -v $(PWD):/workdir -w /workdir cbroglie/goreleaser-cgo:1.12.12 goreleaser --rm-dist
```
cbroglie: This uses the image I built here: https://hub.docker.com/r/cbroglie/goreleaser-cgo.

It simply adds the goreleaser binary to the `dockercore/golang-cross` image which is used for cross builds of https://github.com/docker/cli. Will move it to a more official location soon.

eiginn: Doesn't have to block this but we should move this to an org owned image

cbroglie: Yep, will do

cbroglie:
By adding the tag `v1.4.0-rc1` and running `make release`, the tools were able to generate https://github.com/cloudflare/cfssl/releases/tag/v1.4.0-rc1 automatically

nickysemenza:
💯 🚢

PR #1063, opened 2019-11-24
also output sha256 digest when generating certificate

izolight:
The newkey and newcert endpoints now will print the sha256 digest in addition to md5 and sha1


izolight:
I just went through my open PRs and saw that this is approved but not merged.
Anything in the way for a merge?

PR #1073, opened 2020-01-21
Fix readme still referencing deleted doc file

PeterLincsafe:
Commit ee73190 removed doc/bootstrap.txt but it is still mentioned in README.md

I would like to suggest replacing the entire "Additional documentation" section with a reference to doc/README.txt to prevent repeating it's content.


cbroglie:
Thanks @PeterLincsafe

PR #1077, opened 2020-01-30
Add helper function to trim EC parameters from pem files

darakian:
This PR addresses https://github.com/cloudflare/cfssl/issues/1074

It adds a quick little helper function which checks for the existence of the substring `-----END EC PARAMETERS-----` in the PEM encoded file and returns the content after it. This allows for openssl generated key files to be used directly with cfssl.


helpers/helpers.go
```diff
@@ -378,7 +378,8 @@ func ParsePrivateKeyPEMWithPassword(keyPEM []byte, password []byte) (key crypto.
 
 // GetKeyDERFromPEM parses a PEM-encoded private key and returns DER-format key bytes.
 func GetKeyDERFromPEM(in []byte, password []byte) ([]byte, error) {
-	keyDER, _ := pem.Decode(in)
+	parameterlessKey := ParsePEMPrivateKey(in)
```
cbroglie: Thanks for taking a stab at this @darakian.

A better way to do this would be to loop over the different PEM blocks until we find one which isn't of type `EC PARAMETERS`, something like:
```go
var keyDER *pem.Block
for {
	keyDER, in = pem.Decode(in)
	if keyDER == nil || keyDER.Type != "EC PARAMETERS" {
		break
	}
}
```
Please all include a test case for this, thanks!

darakian: @cbroglie Thanks for the reply. I've updated the PR and replaced the `pem.Decode` step in `GetKeyDERFromPEM`. A test has been added with an openssl generated pem file. Let me know if you want anything else done :)

PR #1080, opened 2020-02-19
Update to ZLint 2.0.0, support filtering out lint sources.

cpu:
:wave: Hi folks,

The [Zlint project](https://github.com/zmap/zlint) just announced version [2.0.0](https://github.com/zmap/zlint/releases/tag/v2.0.0). This major release brings some breaking API changes that are resolved in this branch.

For CFSSL the existing signing profile configuration field `"ignored_lints"` is joined by a new field `"ignored_lint_sources"`. This relies on features in the new 2.0.0 release and is useful for CAs that know certain classes of ZLint lints are never applicable (e.g. CABF EV guideline lints, ETSI ESI lints,
etc). We're also able to drop some of the CFSSL code that supported ignoring lint results by name after running all of the lints. In v2.0.0 we can exclude lints by name before they are run.

Issue #1076, opened 2020-01-29
errors/error_test.go: TestWrap fails wish Go 1.14beta1


eclipseo:
Fedora Rawhide with Go 1.14 beta1

```
github.com/cloudflare/cfssl/errors
--- FAIL: TestWrap (0.00s)
    error_test.go:244: Error message construction failed.
FAIL
```

See upstream change: https://github.com/golang/go/commit/30d7b6400860d87d810a0db3593b28dfb72879f2

claucece:
With the current stable Go 1.14 on MacOS this also fails.

Issue #1075, opened 2020-01-29
api/client/client_test.go: TestNewMutualTLSServer fails with Golang 1.14beta1


eclipseo:
Fedora Rawhide with Go 1.14beta1, cfssl 1.4.1:

```
--- FAIL: TestNewMutualTLSServer (0.00s)
    client_test.go:101: no error message {"code":7400,"message":"failed POST to https://nohost:8888/api/v1/cfssl/sign: Post \"https://nohost:8888/api/v1/cfssl/sign\": dial tcp: lookup nohost on [::1]:53: read udp [::1]:60785-\u003e[::1]:53: read: connection refused"}
FAIL
```

See https://github.com/golang/go/commit/64cfe9fe22113cd6bc05a2c5d0cbe872b1b57860#diff-9667474d33400190658e4758cf28dbf5
and
golang/go#29261

PR #1081, opened 2020-02-19
deps: update /x/crypto, fix Go 1.13 test breakage.

cpu:
When using modern `golang.org/x/crypto/ed25519` on Go 1.13 the `x` library is a small wrapper around the stdlib version. In this case the CFSSL `derhelper` `ParsePrivateKeyDER` function needs to match on the stdlib type.

To maintain backwards compat with Go 1.12 the helper code is split by a build tag. The legacy code can use the `golang.org/x/crypto/ed25519` import while the new Go 1.13+ code can use the `crypto/ed25519` import.

Build tags and code duplication are ugly but I couldn't think of another solution that wouldn't require dropping Go 1.12.x support.


jsha:
FYI it looks like Travis tests passed even though the GitHub UI is still showing the yellow "pending" icon.

cpu:
:wave: @cbroglie @mihirjham @lgarofalo hi folks. Do you think you'll be able to give this PR (and #1080) a once over in the next week or two? My employment situation changed from the time I opened these PRs and it would be nice to get them wrapped up to unblock Let's Encrypt from updating their ZLint dependency. If you don't think you'll have cycles that's OK but I'll probably hand these PRs off to someone else in that event.

Thanks!

claucece:
Hi, @cpu!

Thanks so much for this!

The only problem I see with this is that it will start failing for go.1.14, as it only checks for anything that is !go.1.13. Have you tested it with go1.14?

Something that can also be done is to provide an appropriate error message when the build fails for a given version. Another thing that can be done is to use go contraints (https://medium.com/@theckman/version-constraints-and-go-c9309be15773).. will you be interested into looking into that?

I can also change the code to match that pattern, so let me know ;)

Thanks!

cpu:
Hi @claucece, thanks for the reply!

> The only problem I see with this is that it will start failing for go.1.14, as it only checks for anything that is !go.1.13. Have you tested it with go1.14?

Good point. I'll test this with Go 1.14. I'm not familiar enough with build tags to know for sure what the outcome will be. I _thought_ that `!go 1.13` and `go 1.13` meant "less than Go 1.13" and "Go 1.13 and higher" but could be mistaken.

> Something that can also be done is to provide an appropriate error message when the build fails for a given version. Another thing that can be done is to use go contraints

I think adding another dependency might be overkill for this case. I'll see what I can come up with based on vanilla build tags and we can go from there. Sounds good?

Thanks!

cpu:
Hi again @claucece :wave:

> The only problem I see with this is that it will start failing for go.1.14, as it only checks for anything that is !go.1.13. Have you tested it with go1.14?

I verified the fix continues to work on Go 1.14:

```bash
~/go/src/github.com/cloudflare/cfssl$ go version
go version go1.14 linux/amd64

~/go/src/github.com/cloudflare/cfssl$ git rev-parse HEAD
81e54732c78bc0320d4f6061b9d27b441d8ff8f7

~/go/src/github.com/cloudflare/cfssl$ go test -count=1 --test.run TestParsePrivateKeyPEM ./helpers/...
ok  	github.com/cloudflare/cfssl/helpers	0.004s
ok  	github.com/cloudflare/cfssl/helpers/derhelpers	0.005s [no tests to run]
ok  	github.com/cloudflare/cfssl/helpers/testsuite	0.006s [no tests to run]
```

I ran the specific test I fixed in this branch with `--test.run` because there are other tests that are broken with Go 1.14. E.g. the case you ran into described in https://github.com/cloudflare/cfssl/issues/1076 (and another in https://github.com/cloudflare/cfssl/issues/1075). I think it would be best if those were addressed in a separate PR (at the same time as adding Go 1.14 to the CI build matrix).

> Something that can also be done is to provide an appropriate error message when the build fails for a given version.

I think knowing that the fix works for all of the versions supported in CI (1.12 and 1.13) and the current newest version (Go 1.14) makes me tend towards thinking trying to add a special build time error message for Go 1.11 and older isn't worth the complexity/effort. Do you concur?

Let me know if there is anything else you'd like me to test!

cpu:
> I thought that !go 1.13 and go 1.13 meant "less than Go 1.13" and "Go 1.13 and higher" but could be mistaken.

Re-reading the [Build Constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints) docs made this clearer for me.

The `// +build go1.13` I added to the top of `helpers/derhelpers/derhelpers.go` in this branch will only be satisfied in Go version 1.13 onward (e.g. it is selected for Go 1.14 as well). From that it follows that the `// +build !go1.13` at the top of `helpers/derhelpers/derhelpers-legacy.go` will only be satisfied for versions < Go 1.13. I think that's exactly what we want in this scenario :+1:

cpu:
Thanks!

claucece:
Hi @cpu! 

You are right! Thanks so much for all the work!

PR #1082, opened 2020-02-21
add copy_extensions configuration to local signer to allow 

darkedges:
This Pull Requests add `copy_extensions` to the local signer so that you can specify if you want to copy the additional extensions presented in a CSR. This way a client can make a request say with `qcStatement` and get a certificate issued with those extensions included.

This makes it possible to produce eIDAS like certificates where the details are included in the CSR and the signer can then validate those details before issuing the certificate,

```
{
	"signing": {
		"default": {
			"expiry": "8000h",
			"copy_extensions": true
		}
	}
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ bin
 *.deb
 *.rpm
 test
+
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ profile.out
 bin
 *.deb
 *.rpm
+test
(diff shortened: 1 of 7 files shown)


wbl:
I think this works for our immediate needs, but I'll see what others think. Thanks for doing this!

lgarofalo:
Looks good, only minor thing is it looks like some files are missing a `go fmt`

darkedges:
@lgarofalo Thanks for approving the pull request.

Could you please elaborate on the 

> some files are missing a `go fmt`

bit, as I am not sure what you mean.

PR #1083, opened 2020-02-26
Correctly check for different golang versions #1076

claucece:
This test was failing for golang 1.14. With this, it only checks the string until it differs due to https://github.com/golang/go/commit/30d7b6400860d87d810a0db3593b28dfb72879f2


lgarofalo:
Thank you for this PR, LGTM

PR #1084, opened 2020-03-03
Bump golang to v1.14

chenrui333:

PR #1092, opened 2020-03-23
Add go 1.14 to the CI

claucece:


claucece:
This PR also adds version 1.14 to the go.mod and to the dockerfiles.

lgarofalo:
Got a conflict from merging https://github.com/cloudflare/cfssl/pull/1080

claucece:
It should be updated now @lgarofalo ;)

lgarofalo:
I was comparing to this PR https://github.com/cloudflare/cfssl/pull/1084/files

I think we should add the 
```
    - arch: s390x
      go: 1.14.x
```
And bump the `Dockerfile.alpine` as well.

.travis.yml
```diff
@@ -5,6 +5,7 @@ matrix:
   include:
     - go: 1.12.x
     - go: 1.13.x
+    - go: 1.14
```
chenrui333: It should be 1.14.x to introduce the latest 1.14.1

```suggestion
    - go: 1.14.x
```

Dockerfile.alpine
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13.3-alpine3.10@sha256:40278d43a27b6e0563fcc4dd52c991a25741b1a775402aea342ab9a80158e69e as builder
+FROM golang:1.13.3-alpine3.10@sha256:f98761ca3b0855d554085eac8a55e6162f859b3f409c68fa4a7e3302c9d8bf5f as builder
```
chenrui333: please also use `alpine3.11` as well.

chenrui333: Would be something like this
`FROM golang:1.14.1-alpine3.11@sha256:xx as builder`

chenrui333:
```suggestion
FROM golang:1.14.1-alpine3.11@sha256:244a736db4a1d2611d257e7403c729663ce2eb08d4628868f9d9ef2735496659 as builder
```

claucece: Done! ;)

.travis.yml
claucece: Done! ;)

Dockerfile.alpine
chenrui333: Thanks!

PR #1061, opened 2019-11-23
Add support for generating ed25519 keys and certs

izolight:
It is now possible to generate ed25519 keys and certificates by specifying ed25519 as algo in the json.
This should fix #926

PR #1097, opened 2020-04-16
Add support for generating ed25519 keys and certificates

claucece:
This PR is WIP, and it is built upon: https://github.com/cloudflare/cfssl/pull/1061 by @izolight . 

@lgarofalo , I will be putting the new commits over here, so let me know if it is ok ;)


claucece:
One of the things that this does is remove the go 1.12 as the built target, as the ed25519 package is part of std library...

armfazh:
Have you considered to use CIRCL for ed signatures?

claucece:
Oh, neat idea @armfazh ! I'll start playing with that ;)

claucece:
On the same note @armfazh .. I think openssl also supports ed448, so maybe we can support here as well with your implementation?

claucece:
So, several tests are missing and some needs to be checked for the ed algorithm:

- TestBundleWithECDSAKeyMarshalJSON
- TestGenerate
- TestUbiquitousBundle
- TestUbiquityBundleWithoutMetadata
- TestForceBundleNoFallback
- TestSHA2HomogeneityAgainstUbiquity

armfazh:
> I think openssl also supports ed448, so maybe we can support here as well with your implementation?

That is a great idea, let me know if the current CIRCL API needs changes. Happy to review concise PRs.

claucece:
For sure! I'll finish this PR, and then create another one for ed448 @armfazh . Thanks so much!

claucece:
I jumped into some interfaces conversion problems with using circl, as detailed on issue: https://github.com/cloudflare/circl/issues/109 . Let's see if that can be modified and if not, I'll try something else ;)

claucece:
This also should be taken into account: https://github.com/golang/go/issues/20544

claucece:
So, @armfazh , I've been playing with different ideas but it does not seem like it will work with CIRCL: the 'CreateCertificate' from the X509 package only supports "The currently supported key types are *rsa.PublicKey, *ecdsa.PublicKey and ed25519.PublicKey" (https://golang.org/pkg/crypto/x509/#CreateCertificate), which forces to use the ed25519 from Golang, or do a "conversion" of some kind (which still will mean that package 25519 is needed, ruling out go 1.12). I'll finish this PR by switching back to Golang's ed25119; but I could maybe start an issue into Golang x/crypto libraries for support of ed448 (now that some applications like Privacy Pass will actually need it) and adding it to x509. 

cc./ @alxdavids

armfazh:
One can use circl for signature operations, while being compliant with stdlib types. 
Take a look on https://github.com/claucece/cfssl/pull/1

claucece:
Yes! @armfazh , just did the same! ;) It is sad that we will probably only be able to signing (with ed448, as well).. but well. I'll try to see if in the actual Golang libraries, ed448 can be implemented. I found this issue that mentions that it will be good to have ed448: https://github.com/golang/go/issues?q=is%3Aissue+is%3Aopen+ed448

claucece:
So, @lgarofalo , as part of this PR, I realized that some of the data used for testing is expired, which shouldn't matter while testing as if not we will have to change the data every time that it gets expired. I added a small fix on the previous commit. Let me know if you think is ok ;)

claucece:
@armfazh .. maybe you can help me on this (this current PR has lots of debugging, sorry for that): for some reason using the circl ed25519 fails on the s390x machines, and not on the 64 archs. Using the stdlib ed21559 works just fine. Any ideas?

armfazh:
Let's do baby steps first, so we can make small incremental changes.
This PR is for integrating eddsa into cfssl (use stdlib first).
In another PR, we can make changes to replace stdlib with CIRCL.

In the meantime, we can check what makes circl to fail on s390x architecture.

claucece:
Perfect @armfazh ! 

Then, I will:

* close the PR with stdlib
* make a PR/issue to track what is happening with eddsa in CIRCL at s390x
* integrate eddsa with CIRCL here

Thanks so much!

claucece:
Ok, this PR is basically done.

There is a small thing I found out: the tls package inside the scan folder, that is vendored, has not been update from stdlib in a while. Not sure if it is the best thing to do right now, but the latest version adds there the support for ed25519.

@armfazh @lgarofalo

ltv511:
> There is a small thing I found out: the tls package inside the scan folder, that is vendored, has not been update from stdlib in a while. Not sure if it is the best thing to do right now, but the latest version adds there the support for ed25519.

This should be addressed with https://github.com/cloudflare/cfssl/pull/1101 (still a WIP).

claucece:
Awesome @ltv511 ! I'll keep and eye on it and wait until that one is merged ;)

csr/csr.go
```diff
@@ -216,11 +231,21 @@ func ParseRequest(req *CertificateRequest) (csr, key []byte, err error) {
 			Bytes: key,
 		}
 		key = pem.EncodeToMemory(&block)
+	case ed25519.PrivateKey:
+		key, err = derhelpers.MarshalEd25519PrivateKey(priv)
+		if err != nil {
+
```
bwesterb, line 237: Is there a line missing here?

claucece: Oh, yes! I'll add it ;) Thanks for pointing it out!

armfazh:
A small comment: run go mod tidy to remove the unused dependencies and update vendor folder.

claucece:
Just did it @armfazh . I started a conversation on #1101 to see if I can chime in there. If there is no answer till Friday, I'll start doing it ;)

csr/csr.go
```diff
@@ -216,11 +231,22 @@ func ParseRequest(req *CertificateRequest) (csr, key []byte, err error) {
...
+			Type:  "Ed25519 PRIVATE KEY",
+			Bytes: key,
+		}
+		key = pem.EncodeToMemory(&block)
 	default:
 		panic("Generate should have failed to produce a valid key.")
 	}
 
-	csr, err = Generate(priv.(crypto.Signer), req)
+	csr, err = Generate(priv.(crypto.Signer), req) // TODO: solve
```
bwesterb, line 249: What is the problem the todo refers to?

transport/kp/key_provider.go
```diff
@@ -241,9 +244,30 @@ func (sp *StandardProvider) Generate(algo string, size int) (err error) {
 		}
 		sp.internal.keyPEM = pem.EncodeToMemory(p)
 
+		sp.internal.priv = priv
+	case "ed25519":
+		if size != (ed25519.PublicKeySize * 8) {
```
bwesterb, line 249: The size parameter doesn't refer to the actual size of the public key, does it?

csr/csr.go
claucece: Oh, this was a remanent of thinking how to change some x509 things..

transport/kp/key_provider.go
claucece: Oh, yes.. this should probably refer to the private key size. I'll change this.

centromere:
Thank you for working on this. Is it ready to go?

lukevalenta:
> Thank you for working on this. Is it ready to go?

Looks like it's not quite ready to go. @claucece is no longer at Cloudflare, so we may need someone else to champion this PR.

johannww:
I stumbled into this trying to support ed25519 in [fabric-ca](https://github.com/hyperledger/fabric-ca). I merged the current master branch with this pull request's branch. Everything compiles. The only test failing is the one on **github.com/cloudflare/cfssl/bundler**, but it is also failing in the current master branch. I generated an ed25519 key successfully.

It seems to work. I will leave some comments in the commit.

transport/kp/key_provider.go
johannww: This is counterintuitively right. ed25519.PrivateKeySize accounts for both the public key (32 bytes) and the seed (32 bytes), totaling 64 bytes. ed25519.go

However, I do not see the point of specifying the key size, since ed25519 is always the same, as I know. I think the field size should be simply ignored when generating ed25519 keys.

johannww: This is counterintuitively right. ed25519.PrivateKeySize accounts for both the public key (32 bytes) and the seed (32 bytes), totaling 64 bytes. [ed25519.go](https://github.com/golang/go/blob/8c0256b398f3b23f24a1a3ac0cb00c2d8fb7f506/src/crypto/ed25519/ed25519.go#L154-L170)

**However**, I do not see the point of specifying the key size, since ed25519 is always the same, as I know. I think the field **size** should be simply ignored when generating ed25519 keys.

izolight:
I merged this PRs changes and master into my original PR https://github.com/cloudflare/cfssl/pull/1061
I've also addressed the last two open points on this MR in it.

nickysemenza:
closing in favor of #1061

Issue #1110, opened 2020-05-26
`TestBundleFromFile` broken on master


cpu:
:wave: Hi folks, 

It looks like the `TestBundleFromFile` unit test is failing on master [and breaking builds](https://travis-ci.org/github/cloudflare/cfssl/jobs/691492882):

```go
--- FAIL: TestBundleFromFile (0.16s)
    bundle_from_file_test.go:346: expected no error. but an error occurred: {"code":1211,"message":"x509: certificate has expired or is not yet valid: current time 2020-05-26T18:33:28-04:00 is after 2020-05-07T22:37:00Z"}
FAIL
FAIL	github.com/cloudflare/cfssl/bundler	0.180s
FAIL
```

It seems like the `bundler/testdata/cfssl-leaf-rsa3072.pem` test file has expired as of May 7th and is causing the test failure:

```
cfssl on  master [!] via 🐹 v1.14.2
💀 openssl x509 -in bundler/testdata/cfssl-leaf-rsa3072.pem -noout -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            2f:6d:7b:52:07:47:fd:b9:0e:82:4b:65:27:30:2b:87:99:81:17:f1
        Signature Algorithm: ecdsa-with-SHA384
        Issuer: C = US, ST = California, L = San Francisco, O = CloudFlare, OU = Systems Engineering, CN = cloudflare-inter.com
        Validity
            Not Before: May  8 22:37:00 2019 GMT
            Not After : May  7 22:37:00 2020 GMT
       <snipped>
```

cpu:
Here's a quick-fix (_with a lucky Github PR number :sparkles:_): https://github.com/cloudflare/cfssl/pull/1111 Unfortunately I don't have the cycles at the moment to try and properly regenerate the testcase with a new expiry. If maintainers would prefer that feel free to close the quick-fix PR.

jsha:
Hi Cloudflare team,

Would you mind merging this PR to fix the tests, unblocking #109?

Thanks,
Jacob

cpu:
> Would you mind merging this PR to fix the tests, unblocking #109?

(_I think you meant #1109_)

jsha:
Heh, thanks. Fixed. :-D

mihirjham:
@jsha - We'll get that merged today to unblock you.
Thanks for the PR!

PR #1111, opened 2020-05-26
bundler: temporarily remove `leafRSA3072` testcases.

cpu:
The `leafRSA3072` test file, (`bundler/testdata/cfssl-leaf-rsa3072.pem`) expired and breaks tests. A proper fix would be to regenerate this test file but in the meantime removing the testcases fixes CI.

This branch also updates `TestBundleFromFile` to print the filename of the failing certificate when there is a bundling error. This may help identify which piece of testdata is at fault when CI fails.

Resolves https://github.com/cloudflare/cfssl/issues/1110


mihirjham:
We can merge this to unblock you today and we will look into regenerating the certificate (dynamically preferably) in a future PR.

PR #1112, opened 2020-05-27
Document revocation parameter formats

RocketPuppy:
I struggled with this for a few hours before piecing together this information from a few different issues. I'm proposing this update so nobody else needs to do that struggling. If this is unwanted or incorrect, feel free to close.

### DIFF
--- a/doc/api/endpoint_revoke.txt
+++ b/doc/api/endpoint_revoke.txt
@@ -5,10 +5,10 @@ Method:   POST
 
 Required parameters:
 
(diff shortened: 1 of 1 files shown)


nickysemenza (approved):
Thanks! Looks correct, in https://github.com/cloudflare/cfssl/blob/master/api/certadd/insert.go#L131 we're using `hex.DecodeString` which doesn't accept colons (as per https://golang.org/pkg/encoding/hex/#DecodeString)

PR #1123, opened 2020-08-19
Add support for Circl's signature schemes

bwesterb:
This PR adds support for the signature schemes from Cloudflare's [Circl library](https://github.com/cloudflare/circl), such as the hybrid scheme Ed25519-Dilithium3. This is preparation to set up a secondary post-quantum secure CA.  (See Jira ticket RTG-49 for internal context.)

cfssl relies on Go's internal `crypto/x509`, which makes it impossible to add new signature schemes which are not supported by Go itself.  Thus we have [created a fork of Go](https://github.com/cloudflare/go/tree/cf) that adds support for the signature schemes in Circl to `crypto/x509` and `crypto/tls`.  If cfssl is built with vanilla Go, then nothing changes.  If cfssl is built using our Go fork, then it will include support for the additional signature schemes.  (This is controlled by the `cfgo` build-tag.)

This PR does not yet update cfscan as it does not support TLS 1.3 (yet). We'll add that later.


bwesterb:
@claucece

PR #1126, opened 2020-09-10
feat(authsign): store additional metadata/fields in `certdb`

nickysemenza:
This is a major change in that the included DB migrations *must* be run before the new version of `cfssl` is deployed.
This allows for clients (i.e. https://github.com/cloudflare/certmgr) to send some additional optional fields to `/api/v1/cfssl/authsign` to be stored in `certdb`. It also starts saving SANs, common name, and NotBefore from the issued certificates so that they can be queried without having to parse the PEM.


mihirjham (approved):

lgarofalo (approved):

certdb/pg/migrations/002_AddMetadataToCertificates.sql
```diff
@@ -0,0 +1,21 @@
...
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE certificates
+ADD COLUMN issued_at timestamptz,
+    ADD COLUMN not_before timestamptz,
+    ADD COLUMN originating_host bytea,
+    ADD COLUMN sans bytea,
+    ADD COLUMN tags bytea,
+    ADD COLUMN common_name bytea,
+    ADD COLUMN filename bytea,
+    ADD COLUMN application_name bytea;
```
eiginn, line 11: Maybe we should start the process of not using `bytea` type where it doesn't make sense, starting with these new columns.

nickysemenza: Yeah, I'll update the pg + mysql migrations to be `TEXT`

nickysemenza: done

maxnystrom (approved):

nickysemenza:
updated to move these arbitrary fields into a json column

jmunson (approved):

mihirjham (approved):

szechuen (approved):
Looks great!

Issue #1031, opened 2019-08-23
revoke: race condition accessing CRLSet


mdempsky:
The Go memory model allows concurrent *reads* of variables, but a write cannot be concurrent with any other operation whether read or write. This includes maps.

For example, see the [Go 1.6 release notes](https://golang.org/doc/go1.6#runtime) [emphasis added]:

> The runtime has added lightweight, best-effort detection of concurrent misuse of maps. **As always, if one goroutine is writing to a map, no other goroutine should be reading or writing the map concurrently.** If the runtime detects this condition, it prints a diagnosis and crashes the program. The best way to find out more about the problem is to run the program under the race detector, which will more reliably identify the race and give more detail.

Package revoke declares:

```
var CRLSet = map[string]*pkix.CertificateList{}
var crlLock = new(sync.Mutex)
```

Within `certIsRevokedCRL`, `crlLock` is used to protect `delete(CRLSet, url)` and `CRLSet[url] = crl` (i.e., write operations on `CRLSet`), but there's no protection on the `CRLSet[url]` read operation at the top of the function.

Moreover, since `crlLock` isn't exported, there's no way for users of package revoke to safely access `CRLSet` unless they know there aren't any concurrent certificate revocation checks happening (e.g., through higher-level application mutexes).

Two possible options (not exhaustive):

1. If you can safely rely on Go 1.9 and can break your API, you can switch to using [sync.Map](https://golang.org/pkg/sync/#Map).

2. If backwards compatibility is important, you can at least add locking around the `CRLSet[url]` read operation; export `crlLock` (e.g., as `CRLLock`); and document to users that they need to protect access to `CRLSet` using `CRLLock`.

PR #1132, opened 2020-09-25
Fix race condition in revoke

asayyah:
This change fixes a race condition in revoke where read operation to CRLSet is not protected by its associated mutex.

There is an open issue for the same problem already:

https://github.com/cloudflare/cfssl/issues/1031

### DIFF
--- a/revoke/revoke.go
+++ b/revoke/revoke.go
@@ -133,13 +133,13 @@ func getIssuer(cert *x509.Certificate) *x509.Certificate {
 // check a cert against a specific CRL. Returns the same bool pair
 // as revCheck, plus an error if one occurred.
 func certIsRevokedCRL(cert *x509.Certificate, url string) (revoked, ok bool, err error) {
+	crlLock.Lock()
 	crl, ok := CRLSet[url]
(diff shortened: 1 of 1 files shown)


nickysemenza:
Thanks - to address the second half of #1031, I think it's a bit un-idiomatic to export the mutex, so I think we should either:
1. Stop exporting the unsafe map itself
2. Create exported helpers which safely access the map - although i'm not too familiar with the use cases here, it seems like every public repo that imports this is currently using the existing exported functions (https://github.com/search?q=%22revoke.VerifyCertificate%22&type=Code) as opposed to directly accessing the map (https://github.com/search?q=%22revoke.CRLSet%22&type=Code), so (1) seems like it wouldn't be a breaking change. 

But either way this is a great fix as-is :)

asayyah:
Yes either of those of work, but both options are potentially going to break customers and probably better to be done with care. Does it make sense to go forward with this change so that the underlying race is fixed first?

nickysemenza:
@asayyah yep certainly, i'll merge this and keep the linked issue open.

PR #1135, opened 2020-10-13
fix selecting rows created before migration introduced in #1126

nickysemenza:
fix(certdb): use `sql.NullString` for CertificateRecord.CommonName

Rows inserted before the migration in #1126 will have the `common_name` set to NULL. This fixes selects for those rows.

### DIFF
--- a/certdb/certdb.go
+++ b/certdb/certdb.go
@@ -1,6 +1,7 @@
 package certdb
 
 import (
+	"database/sql"
 	"encoding/json"
 	"time"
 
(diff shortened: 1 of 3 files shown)


nickysemenza:
CI run from after I added a test to catch this but before I fixed it: https://travis-ci.org/github/cloudflare/cfssl/builds/735543224

mitalirawat (approved):

mihirjham (approved):

PR #1136, opened 2020-10-13
fix(certdb): allow reading other null columns (part 2 of #1135)

nickysemenza:
This follows up #1135 to properly handle the case when columns written pre-migration have null values.

### DIFF
--- a/certdb/sql/sql_test.go
+++ b/certdb/sql/sql_test.go
@@ -172,7 +172,12 @@ func testInsertCertificateAndGetUnexpiredCertificateNullCommonName(ta TestAccess
 	}
 
 	// simulate situation where there are rows before migrate 002 has been run
-	ta.DB.MustExec("update certificates set common_name = NULL")
+	ta.DB.MustExec(`update certificates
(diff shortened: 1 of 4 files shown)


certdb/sql/sql_test.go
```diff
@@ -172,7 +172,12 @@ func testInsertCertificateAndGetUnexpiredCertificateNullCommonName(ta TestAccess
 	}
 
 	// simulate situation where there are rows before migrate 002 has been run
-	ta.DB.MustExec("update certificates set common_name = NULL")
+	ta.DB.MustExec(`update certificates
```
nickysemenza, line 175: Right now when CI starts it immediately migrates everything, I think that an improvement could be to migrate up to n-1 migrates when CI starts, and then apply/rollback the most recent migration during tests to ensure that new and old versions of the code behave with new and old data. However it's kind of complex given that we support 3 different database types, so i'm not sure if there's much value given that this 002 was the first migration in 5+ years.

mihirjham (approved):

PR #1137, opened 2020-10-15
Update dependency zlint to 2.2.1

hugoboos:
Zlint has released a new version to fix a vulnerability in `golang.org/x/text`: https://github.com/zmap/zlint/releases/tag/v2.2.1.

See: https://nvd.nist.gov/vuln/detail/CVE-2020-14040 for more information about the vulnerability.

PR #1138, opened 2020-10-15
Update to zlint 2.2.1

jsha:


cpu:
@jsha I think you're duplicating https://github.com/cloudflare/cfssl/pull/1137 here (and it has a maintainer +1)

jsha:
Thanks for spotting!

cpu:
My ZLint sense was tingling! 😆

Issue #1152, opened 2020-11-30
OCSPrefresh multiple CA's


b4ldr:
I have configured mutliroot ca with two different intermidiate CA's which both use the same db.conf and hence the same database e.g.

```
[Wikimedia_SRE_Internal]
private = file:///etc/cfssl/signers/Wikimedia_SRE_Internal/ca/Wikimedia_SRE_Internal-key.pem
certificate = /etc/cfssl/signers/Wikimedia_SRE_Internal/ca/Wikimedia_SRE_Internal.pem
config = /etc/cfssl/signers/Wikimedia_SRE_Internal/cfssl.conf
dbconfig = /etc/cfssl/db.conf

[Wikimedia_testing_CA]
private = file:///etc/cfssl/ssl/Wikimedia_testing_CA/Wikimedia_testing_CA-key.pem
certificate = /etc/cfssl/ssl/Wikimedia_testing_CA/Wikimedia_testing_CA.pem
config = /etc/cfssl/signers/Wikimedia_testing_CA/cfssl.conf
dbconfig = /etc/cfssl/db.conf
```

This seems to work well and i can issue and revoke certs in both CA's without issue.  however when i attempt to configure OCSP using ocsprefresh i get the following error

```
$ sudo /usr/bin/cfssl ocsprefresh \                                                                                        
          -db-config /etc/cfssl/db.conf \
          -responder /etc/cfssl/ssl/ocsp/OCSP_Wikimedia_SRE_Internal_ocsp_discovery_wmnet.pem \
          -responder-key /etc/cfssl/ssl/ocsp/OCSP_Wikimedia_SRE_Internal_ocsp_discovery_wmnet-key.pem \
          -ca /etc/cfssl/signers/Wikimedia_SRE_Internal/ca/Wikimedia_SRE_Internal.pem -loglevel=0
2020/11/30 18:24:45 [DEBUG] Loading issuer cert: /etc/cfssl/signers/Wikimedia_SRE_Internal/ca/Wikimedia_SRE_Internal.pem
2020/11/30 18:24:45 [DEBUG] Loading responder cert: /etc/cfssl/ssl/ocsp/OCSP_Wikimedia_SRE_Internal_ocsp_discovery_wmnet.pem
2020/11/30 18:24:45 [DEBUG] Loading responder key: /etc/cfssl/ssl/ocsp/OCSP_Wikimedia_SRE_Internal_ocsp_discovery_wmnet-key.pem
[...]

PR #956, opened 2018-12-05
add support for multi AKI in ocsprefresh. only one DB is needed for ocspserver

mitch-geht-ab:
Hi,

I've added support for ocsprefresh based on the SKI of the cert given with ocsprefresh command.
`cfssl ocsprefresh -db-config db-config -ca cert -responder....`

With multiple AKI in DB table certificates i got allways
```
./cfssl ocsprefresh -db-config db-maria.json -ca $ROOTCA.pem -responder $OCSPCERT.pem -responder-key $OCSPCERT-key.pem -loglevel=0
2018/12/05 13:43:08 [DEBUG] Loading issuer cert: xxx-root-cai.pem
2018/12/05 13:43:08 [DEBUG] Loading responder cert: cert_ocsp.pem
2018/12/05 13:43:08 [DEBUG] Loading responder key: cert_ocsp-key.pem
2018/12/05 13:43:08 [DEBUG] loading db configuration file from db-maria.json
2018/12/05 13:43:08 [CRITICAL] Unable to sign OCSP response: {"code":8100,"message":"Certificate not issued by this issuer"}
{"code":8100,"message":"Certificate not issued by this issuer"}
```
when I tried to refresh the ocsp DB.
Because I had certs from RootCA and IntermediateCA in my DB.

`cfssl2 ocspserve ...`
works also very well and can service both CA (root and intermediate).

```
Root
|--- Intermediate I
      |--- Entity1
      |--- Entity2
|--- Intermediate II
      |--- Entity1
```

Of cource, the patch might not be very well coded. Sorry for that, I'm new with go.

Best regards 
Thomas

PR #1153, opened 2020-12-01
(#1152) OCSP Refresh multi CA support

b4ldr:
This is an attempt to add multi CA support to ocsprefresh to address the
issue discussed in #1152

(fits go PR so apologize in advance for any stupid issues)


b4ldr:
I notice a similar request is already open #956

PR #1157, opened 2020-12-09
fix: cfssl provider support SAN URI

mayocream:
`transport` package doesn't send `URIs` column in CSR, change this behavior.

### DIFF
--- a/transport/ca/cfssl_provider.go
+++ b/transport/ca/cfssl_provider.go
@@ -206,13 +206,17 @@ func (cap *CFSSL) SignCSR(csrPEM []byte) (cert []byte, err error) {
 		return nil, err
 	}
 
-	hosts := make([]string, len(csr.DNSNames), len(csr.DNSNames)+len(csr.IPAddresses))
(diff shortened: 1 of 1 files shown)


transport/ca/cfssl_provider.go
```diff
@@ -206,13 +206,17 @@ func (cap *CFSSL) SignCSR(csrPEM []byte) (cert []byte, err error) {
 		return nil, err
 	}
 
-	hosts := make([]string, len(csr.DNSNames), len(csr.DNSNames)+len(csr.IPAddresses))
+	hosts := make([]string, len(csr.DNSNames), len(csr.DNSNames)+len(csr.IPAddresses)+len(csr.URIs))
```
nickysemenza, line 209: I know this was existing code, but there's no sense in making the array the right length if it's being appended to, see https://goplay.space/#6CL-nQJOxsS

```suggestion
	hosts := make([]string, 0)
```

mayocream: Thanks!

PR #1163, opened 2020-12-31
Added support for general OID names in CSRs

akgmartin:


csr/csr.go
```diff
@@ -157,8 +160,24 @@ func appendIf(s string, a *[]string) {
...
+func OIDFromString(s string) (asn1.ObjectIdentifier, error) {
+	var oid []int
+	parts := strings.Split(s, ".")
+	if len(parts) < 1 {
+		return oid, fmt.Errorf("invalid OID string: %s", s)
+	}
+	for _, p := range parts {
+		i, err := strconv.Atoi(p)
+		if err != nil {
+			return oid, fmt.Errorf("invalid OID part %s", p)
```
benbuzbee, line 172: may want to return  nil slice instead of partially constructed oid

signer/signer.go
```diff
@@ -172,10 +172,29 @@ func DefaultSigAlgo(priv crypto.Signer) x509.SignatureAlgorithm {
...
+	return (len(t) == 4 && t[0] == 2 && t[1] == 5 && t[2] == 4 && (t[3] == 3 || (t[3] >= 5 && t[3] <= 11) || t[3] == 17))
+}
+
 // ParseCertificateRequest takes an incoming certificate request and
 // builds a certificate template from it.
 func ParseCertificateRequest(s Signer, p *config.SigningProfile, csrBytes []byte) (template *x509.Certificate, err error) {
 	csrv, err := x509.ParseCertificateRequest(csrBytes)
+
+	var r pkix.RDNSequence
+	_, err = asn1.Unmarshal(csrv.RawSubject, &r)
```
benbuzbee, line 185: this results in the err from line 182 (ParseCertificateRequest) being overwritten and not checked for nil

csr/csr.go
```diff
@@ -234,6 +260,7 @@ func ParseRequest(req *CertificateRequest) (csr, key []byte, err error) {
 // from an existing certificate. For a root certificate, the CA expiry
 // length is calculated as the duration between cert.NotAfter and cert.NotBefore.
 func ExtractCertificateRequest(cert *x509.Certificate) *CertificateRequest {
+	fmt.Printf("ExctractCertificateRequest %+v\n", *cert)
```
benbuzbee, line 263: This line is causing the tests to fail because the tests parse stdout

benbuzbee (changes requested):

benbuzbee (approved):

PR #1164, opened 2021-01-05
Allow users to use custom HTTP client

grongor:
This is by no means the best solution, but at least it allows setting the options like timeouts just for this package.


nickysemenza (approved):
> This is by no means the best solution, but at least it allows setting the options like timeouts just for this package.

Yeah these mutable global configs aren't ideal IMO, but this seems to be the least disruptive way to do it.

PR #1170, opened 2021-01-24
fix error check when check signature

### LINKED ISSUE #1169: error will always bee 'nil' here
https://github.com/cloudflare/cfssl/blob/3cc617f27de99f13068b41960b52a62015616e97/selfsign/selfsign.go#L36

```
csr.CheckSignature()
	if err != nil {
		err = cferr.Wrap(cferr.CSRError, cferr.KeyMismatch, err)
		return
	}
```

qinghai5060:
Resolves #1169


nickysemenza:
Thanks!

Issue #1171, opened 2021-02-05
Leak issues in revoke package


aloababa:
I found leak issues in the revoke package.

The first one is in the `fetchCRL` function:

```
func fetchCRL(url string) (*pkix.CertificateList, error) {
	resp, err := HTTPClient.Get(url)
	if err != nil {
		return nil, err
	} else if resp.StatusCode >= 300 {
		return nil, errors.New("failed to retrieve CRL")
	}

	body, err := crlRead(resp.Body)
	if err != nil {
		return nil, err
	}
	resp.Body.Close()

	return x509.ParseCRL(body)
}
```

So here if status code is `>= 300` or if `crlRead` fails the body is never closed.

The second one is in the `fetchRemote` function:

```
func fetchRemote(url string) (*x509.Certificate, error) {
	resp, err := HTTPClient.Get(url)
	if err != nil {
		return nil, err
	}

	in, err := remoteRead(resp.Body)
	if err != nil {
		return nil, err
	}
	resp.Body.Close()

	p, _ := pem.Decode(in)
	if p != nil {
		return helpers.ParseCertificatePEM(in)
	}

	return x509.ParseCertificate(in)
}
```

If remoteRead fails the body is never closed.

This can lead to too many open file error and this is not good.

As a fix you can defer resp.Body.Close() rigth away after the http call this is meant for that.

```
resp, err := HTTPClient.Get(url)
if err != nil {
    return nil, err
}
defer resp.Body.Close()
```

nickysemenza:
Good find! Are you planning to make a PR for this? If not I can.

aloababa:
Yes i can i will submit a proposal this week end

PR #1175, opened 2021-02-16
Add http response close in client post function

chienfuchen32:


nickysemenza:
Thanks! similar to #1171, some linters would be helpful to catch this.

PR #1180, opened 2021-03-05
CRL distribution points having same CRL should not reject certificate

Yogi1994:
**Problem**:There are cases when there is same CRL hosted at multiple locations for high availability.
Some if the URLs(CRL Distribution points) may not be reachable. In this case, it is not required for the certificate to pass every CRL(as the CRL is same in every hosted location).
**Solution**: Verify at least 1 hosted location and not fail if host is not reachable.
If any of the URL is reachable and verified as not revoked then we should not reject the certificate.

### DIFF
--- a/revoke/revoke.go
+++ b/revoke/revoke.go
@@ -33,6 +33,9 @@ var HTTPClient = http.DefaultClient
 // verification to fail (a hard failure).
 var HardFail = false
 
+//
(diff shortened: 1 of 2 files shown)


Yogi1994:
@nickysemenza, @maxnystrom , @mihirjham , @cbroglie

jovillas:
Any update on this?

Issue #829, opened 2017-11-29
certadd endpoint not working


johanneslanger:
When posting against the endpoint /api/v1/cfssl/certadd I get a 404, example

```
curl -d '{ "serial": "568433120340650732114301233628109166578730740368", "authority_key_id": "01b712db6a966ef0ddd8c50d6ddecdb25128ad59", "status": "good" , "pem" : "-----BEGIN CERTIFICATE-----\nMIIGFD................\n-----END CERTIFICATE-----" }' http://cfssl-api:8080/api/v1/cfssl/certadd
```

returns:

404 page not found

minaevmike:
Hi, have same issue. Can you help me? thank you.

r0039:
> I guess this is a little outdated, but for anyone who faces this problem again. It is related to file permissions.
> I was having the same issue, after correcting the permission from the ca-key.pem, the cfssl service is able to response http calls correctly.

hi @rchicoli  I did chmod to 644 for ca-key.pem, but it didn't work. I still receive 404 Not Found. Could you please let me know what did you do? Thank you.

My case:
`-rw-r--r--  1 root root   227 Mar  4 06:31 ca-key.pem
`
```
MacBook-Pro:~ tien$ curl -I http://x.x.x.x:8888/api/v1/cfssl/revoke
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Wed, 06 Mar 2019 06:32:55 GMT
Content-Length: 19
```

rchicoli:
it is hard to remember, because it happens a long time ago. But I guess this is how i fixed:
```yaml
- name: correct ca-key permissions
  file:
    path: "{{ cfssl_data_dir }}/ca-key.pem"
    owner: "{{ cfssl_user }}"
```

ttinkr:
@r0039 I know this is a really old thread, but I am facing the same issues even with correct permissions. Did you manage to get it working after all?

PR #1181, opened 2021-03-23
Make certadd API available

myokoyama28:
There is not the certadd endpoint and it inserts a record that does not match the current structure of the certificates table,
so I fix them.

related issue: https://github.com/cloudflare/cfssl/issues/829


nickysemenza:
Thanks for this!

PR #1195, opened 2021-06-17
pick columns to get revoked/unexpired certs

BowonY:
Selecting all columns to get all revoked and unexpired certificates is expensive and unnecessary. 
This PR adds a function updating existing database accessor function to select only serial_number and revoked_at

### DIFF
--- a/certdb/sql/database_accessor.go
+++ b/certdb/sql/database_accessor.go
@@ -36,6 +36,10 @@ SELECT %s FROM certificates
 SELECT %s FROM certificates
 	WHERE CURRENT_TIMESTAMP < expiry AND status='revoked' AND ca_label= ?;`
 
+	selectRevokedAndUnexpiredWithLabelSQL = `
+SELECT serial_number, revoked_at FROM certificates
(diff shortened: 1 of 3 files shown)


certdb/sql/database_accessor.go
```diff
@@ -202,6 +206,21 @@ func (d *Accessor) GetRevokedAndUnexpiredCertificatesByLabel(label string) (crs
 	return crs, nil
 }
 
+// GetRevokedAndUnexpiredCertificatesSelectColumnsByLabel gets serial_number and revoed_at from all revoked and unexpired certificate from db (for CRLs) with specified ca_label.
+func (d *Accessor) GetRevokedAndUnexpiredCertificatesByLabelSelectColumns(label string) (crs []certdb.CertificateRecord, err error) {
```
nickysemenza, line 210: do you think it makes sense to make another struct that's a subset of `certdb.CertificateRecord` that way someone doesn't try to use one of the other cols in the future (which will always be empty)

nickysemenza (approved):

mihirjham:
Approved

mihirjham (approved):

PR #1199, opened 2021-07-28
WIP: add custom client to revoke

ddhawal:


ddhawal:
If a proxy is set in the ENV then there is no need to add a proxy in the client for service running as a daemon. However, if required will reopen this MR.

PR #1218, opened 2021-12-20
Implement the bundle parameter in multirootca authsign

jayme-github:
`/api/v1/cfssl/authsign` now supports the bundle parameter and returns an "optimal" bundle in case it is set in a request (as documented in cfssl/doc/api/endpoint_authsign.txt).

For this to work the rootCA must either be in the systems root store or it might be provided by filename in `multirootca.conf`, like:
```
[intermediate1]
private = file:///intermediate/intermediate1-key.pem
certificate =/intermediate/intermediate1.pem
config =/intermediate/intermediate-config.json
ca = /ca/ca.pem
```

### DIFF
--- a/cmd/multirootca/api.go
+++ b/cmd/multirootca/api.go
@@ -8,7 +8,9 @@ import (
 	"net/http/httputil"
 
 	"github.com/cloudflare/cfssl/api"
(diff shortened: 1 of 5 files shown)


elukey:

PR #1219, opened 2021-12-21
Add BundleSign and BundleAuthSign methods to client Remote

jayme-github:
BundleSign/BundleAuthSign methods can be used like Sign/AuthSign to sign
CSRs but will return the "optimal" bundle instead of only the
certificate.
For that to work, the caller needs to set the "bundle" parameter in the
json POST data to true.

This is definitely not the best way to implement this, but I think it's
the least intrusive one and does not break the interface of the client.


elukey:

PR #1224, opened 2022-01-16
add support for m1 macs

romantomjak:
Adding support for arm64 builds on darwin


troyanov:
@nickysemenza do you have an ETA for a new cfssl version? 
It looks like there are not that many changes https://github.com/cloudflare/cfssl/compare/v1.6.1...master

Would be great to have have binaries for Apple arm64.

Thanks in advance.

PR #1255, opened 2022-11-19
go.mod: update for go1.18 and update/remove obsolete dependencies

thaJeztah:
see individual commits for details.

### go.mod: tidy and vendor with go1.18

go1.17 and older are deprecated and no longer tested in CI

### remove deprecated golint (golang.org/x/lint)

CI is already using golangci-lint, which should cover whatever
golint was still covering.

### go.mod: github.com/google/certificate-transparency-go v1.1.4

This version drops a great number of (indirect) dependencies 🎉 

Screenshot 2022-11-20 at 01 07 12

### go.mod: github.com/cloudflare/redoctober v0.0.0-20211013234631-6a74ccc611f6

Adds go module support and drops various obsolete dependencies;

full diff: https://github.com/cloudflare/redoctober/compare/99c99a8e7544...6a74ccc611f6

### go.mod: github.com/jmhodges/clock v1.2.0

Adds go module support. No code changes in vendored files

full diff: https://github.com/jmhodges/clock/compare/880ee4c33548...v1.2.0

### go.mod: golang.org/x/net v0.2.0

the golang.org/x projects now tag releases.

### go.mod: golang.org/x/crypto v0.3.0

the golang.org/x projects now tag releases.

### go.mod: github.com/matttproud/golang_protobuf_extensions v1.0.2

adds go module support

PR #1256, opened 2022-11-20
certdb/sql: remove uses of github.com/stretchr/testify/require

thaJeztah:
- [x] follow-up of / depends on https://github.com/cloudflare/cfssl/pull/1255

This was the only location where this assertion library was used; all
other code in this repository used Go's stdlib for assertions. The
remaining uses in this package could be replaced without too much
trouble, which allowed for the dependency (including its indirect
dependencies) to be removed altogether.


thaJeztah:
Rebased; this one should be ready for review as well, @nickysemenza (thanks for reviewing the other ones!)

thaJeztah:
@nickysemenza PTAL 😅

PR #1257, opened 2022-11-20
transport/ca/localca: New(): panic instead of os.Exit(1), remove github.com/kisom/goutils/assert

thaJeztah:
- This will conflict with https://github.com/cloudflare/cfssl/pull/1255, but happy to do a rebase if that one gets merged first

### transport/ca/localca: New(): panic instead of os.Exit(1)

This code was added in 56dfed7c82597110116d549dc1c990a5e24d1db7, but
now has become the only use of github.com/kisom/goutils/assert, which
previously was used in tests, and now is archived (moved to a new module).

There were a couple of issues with this code; 

The `assert.NoError` appears to have a bug; it accepts optional arguments,
but those are ignored; https://github.com/kisom/goutils/blob/v1.4.3/assert/assert.go#L90-L99
In this case, it meant that the additional information to describe the
error won't be printed.
 
Looking at the code (https://github.com/kisom/goutils/blob/v1.4.3/assert/assert.go#L35-L45),
it defaults (`GOTRACEBACK` anything other than "crash") using `os.Exit(1)`.
While (from the description), program execution MUST be terminated, there
are some downsides to using `os.Exit` here, as it terminates execution
immediately (which is desirable), but has no way to recover. While users
should NOT use the result in this case, they still may want to catch this
error (without terminating the program as a whole, which may be problematic
if this module is used as part of a service).

Unlike `panic`, `os.Exit` also does not execute pending `defer` statements,
which may still be desirable to handle state cleanup.

### transport/ca/localca: remove uses of github.com/kisom/goutils/assert

This package was the only package using this assertion library. Looking
for a replacement, all packages (except for one) in this repository were
not using an assertion library, so replacing it with standard checks.

### transport/ca/localca: remove unused testGenerateKeypair utility

This was added in 56dfed7c82597110116d549dc1c990a5e24d1db7, but never
used.

### transport/ca/localca: remove uses of deprecated io/ioutil

Using their replacements instead. Also making use of t.TempDir(),
to let Go's testing take care of cleaning up.


transport/ca/localca/signer.go
```diff
@@ -146,13 +148,22 @@ func New(req *csr.CertificateRequest, profiles *config.Signing) (*CA, error) {
 	// CFSSL has become inconsistent, and it can't be trusted.
 
 	priv, err := helpers.ParsePrivateKeyPEM(keyPEM)
-	assert.NoError(err, "CFSSL-generated private key can't be parsed")
+	if err != nil {
+		println("CFSSL-generated private key can't be parsed")
```
nickysemenza, line 152: since the function signature supports it, I think returning an error would be more idiomatic than panic (or os.exit for that matter) 
```suggestion
		return nil, fmt.Errorf("CFSSL-generated private key can't be parsed: %w", err)
```

thaJeztah: Yup, works for me! I tried to stay close to the original approach, but agreed that if we return nil, nobody would be able to use the result anyway. Let me update.

thaJeztah:
Updated 👍

nickysemenza:
thanks again!

thaJeztah:
@nickysemenza you're welcome! Was juggling complex module dependency resolution in some projects, and noticed that I could reduce some (indirect) dependencies here, so thought I'd roll up my sleeves and do a bit of legwork.

Last one (for now at least 😅) is in https://github.com/cloudflare/cfssl/pull/1256

PR #1258, opened 2022-11-20
chore: remove uses of deprecated io/ioutil

thaJeztah:
Grouped per top-level package; there's one remaining use in transport/ca/localca, which is being removed in https://github.com/cloudflare/cfssl/pull/1257


nickysemenza:
thanks!

PR #1259, opened 2022-11-21
Add linux/arm64 Build

sonork:
Add arm64 build for linux.
Disable CGO except for darwin.
Use gythialy/golang-cross for snapshot builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,11 @@ __check_defined = \
 
 .PHONY: snapshot
 snapshot:
-	docker run --rm  -v $(PWD):/workdir -w /workdir cbroglie/goreleaser-cgo:1.12.12-musl goreleaser --rm-dist --snapshot --skip-publish
+	docker run --rm --privileged \
+    -v $(PWD):/cross \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -52,7 +52,11 @@ __check_defined = \
 
 .PHONY: snapshot
 snapshot:
-	docker run --rm  -v $(PWD):/workdir -w /workdir cbroglie/goreleaser-cgo:1.12.12-musl goreleaser --rm-dist --snapshot --skip-publish
+	docker run --rm --privileged \
+    -v $(PWD):/cross \
+    -v /var/run/docker.sock:/var/run/docker.sock \
```
thaJeztah, line 57: Is the `--privileged` and access to the docker socket needed? (effectively, this gives full root access to the host)

sonork: Removed `--privileged` and docker socket from `snapshot` and `github-release` make targets.
Thanks for pointing it out.

nickysemenza:
> Disable CGO except for darwin.
This should be fine - there is a usecase where sqlite is used which uses cgo, but I think that's not super common, can revisit later if need be.

Issue #1265, opened 2023-01-06
Adding multiarch support to cfssl docker image


shahidhs-ibm:
The cfssl docker images supports [amd64](https://hub.docker.com/r/cfssl/cfssl/tags) arch only. Are there any plans to extend its support to other architectures? 

Also we did not got much clarity on how the docker images for cfssl are getting build and published. Could you please elaborate on how its done?

We feel that enabling docker buildx support will be the simplest way to achieve multi-arch support and we would like to provide assistance (if the idea sounds good). Please let us know your thoughts.

nickysemenza:
the docker build process is pretty rudimentary right now - just a make target that pushes to the registry: https://github.com/cloudflare/cfssl/blob/master/Makefile#L69-L73 that I run manually.

I think a github action that builds a multi-arch image (and then pushes to dockerhub) would be great. If you'd like to PR something that builds the image in a Github Action I'd be happy to add in the docker push portion with the relevant secrets!

shahidhs-ibm:
Thanks @nickysemenza for the positive response. We'll work on this feature and create a PR for same.

nickysemenza:
@shahidhs-ibm check out https://hub.docker.com/layers/cfssl/cfssl/master/images/sha256-4eebea119b2c2c0d994f1d0046b7531e510870814b81fd742a9f6512929c8e80?context=repo and let me know if that works!

shahidhs-ibm:
@nickysemenza Thanks for publishing multi-arch images for cfssl.
I am able to pull the cfssl docker image on my s390x arch VM and able to run cfssl commands:
```
$ docker inspect cfssl/cfssl:master | grep -i arch
        "Architecture": "s390x",
        
$ docker run -it --entrypoint cfssl  cfssl/cfssl:master version
Version: 1.6.3
Runtime: go1.20.2
```
All looks good to me. 👍

nickysemenza:
great!

PR #1267, opened 2023-01-12
Add support to multi architecture docker image using github action

shahidhs-ibm:
Raising this PR as per the discussion in issue [#1265](https://github.com/cloudflare/cfssl/issues/1265#issuecomment-1373979871).
The GitHub Action will run on the event when we tag a commit. I had to upgrade the Go version to `1.19.3` for docker buildx compatibility.

Let me know if you have any questions/comments.


shahidhs-ibm:
Not sure why Go/lint GitHub Action failed but seems unrelated to the PR'ed code changes. Kindly confirm.

thaJeztah:
Looks like it hit a timeout (perhaps it needs longer to run);


```
Running [/home/runner/golangci-lint-1.50.1-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
```

Makefile
```diff
@@ -9,7 +9,7 @@ all: bin/cfssl bin/cfssl-bundle bin/cfssl-certinfo bin/cfssl-newkey bin/cfssl-sc
 
 bin/%: $(shell find . -type f -name '*.go')
 	@mkdir -p $(dir $@)
-	go build -ldflags $(LDFLAGS) -o $@ ./cmd/$(@F)
+	GOOS=$(TARGETOS) GOARCH=$(TARGETARCH) go build -ldflags $(LDFLAGS) -o $@ ./cmd/$(@F)
```
thaJeztah, line 12: Is this makefile only to be used inside the Dockerfile? If not, then it probably should not contain the `TARGETOS` and `TARGETARCH` variables here (as they won't be present in other situations)

Dockerfile
```diff
@@ -1,11 +1,13 @@
-FROM golang:1.16.15@sha256:35fa3cfd4ec01a520f6986535d8f70a5eeef2d40fb8019ff626da24989bdd4f1
+FROM --platform=${BUILDPLATFORM} golang:1.19.3@sha256:d388153691a825844ebb3586dd04d1c60a2215522cc445701424205dffc8a83e
+
+ARG TARGETOS TARGETARCH
```
thaJeztah, line 3: I _think_ both formats are supported, but defining multiple build-args in a single `ARG` can sometimes be somewhat ambiguous; the `ENV` instruction supports `ENV NAME VALUE` (without `=` in between). To prevent confusion, it may be clearer (although slightly more verbose) to use two separate `ARG` instructions;

```dockerfile
ARG TARGETOS
ARG TARGETARCH
```

Dockerfile
```diff
@@ -1,11 +1,13 @@
-FROM golang:1.16.15@sha256:35fa3cfd4ec01a520f6986535d8f70a5eeef2d40fb8019ff626da24989bdd4f1
+FROM --platform=${BUILDPLATFORM} golang:1.19.3@sha256:d388153691a825844ebb3586dd04d1c60a2215522cc445701424205dffc8a83e
```
thaJeztah, line 1: I see the "pinning by digest" was added in https://github.com/cloudflare/cfssl/pull/1047 (but not mentioned). Wondering if there was a strong reason to use these

Note that when using a combination of `tag` and `digest`, the `tag` is considered "for informational purposes" and ignored. When specifying a digest, no "lookup" is performed for the tag, and the image is pulled by its digest. 

While this allows for using an immutable reference, it can also be error-prone. I'm wondering if there was a strong reason to pin it to a digest, or if a regular `:tag` would be okay for this.

Makefile
shahidhs-ibm: Inside the Dockerfile, we are git cloning the `cfssl_trust` and executing its Makefile. This made the things tricky as I didn't wanted to touch `cfssl_trust` repository. This is a workaround solution that I could come up with for the issue. I have tested and can confirm that absence of values in `TARGETOS` and `TARGETARCH` variables does not cause any issues and `go build` command works fine.

Dockerfile
shahidhs-ibm: Agreed. Will update.

Tyrael:
hi,

any update on this?
we were wondering why there are mo recent images from cfssl on dockerhub and then found https://github.com/cloudflare/cfssl/issues/1265 and then this PR

cc @nickysemenza

Issue #1280, opened 2023-04-07
CSR JSON supports specifying custom extensions, but selfsign ignores them


sthussey:
When building the certificate for a self-signed cert from a CSR json file, the Extensions field in the CSR is ignored. I'll open a PR with a fix (I used it locally and got a self-signed cert w/ my needed extension).

PR #1281, opened 2023-04-07
Respect custom x509 ext in selfsign

sthussey:
- No longer ignore specified x509 extensions in the CSR when generating a self-signed certificate.


sthussey:
This resolves issue #1280

nickysemenza:
this looks good; is there a simple test you could add for this?

sthussey:
@nickysemenza Sorry for the long delay, lost track of this. I've added a test.

PR #1282, opened 2023-04-18
goreleaser: Add ARMv7 binaries

patrickelectric:
Such boards are still available

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,6 +24,7 @@ builds:
       - linux
     goarch:
       - amd64
+      - arm
       - arm64
       - s390x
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -14,6 +14,7 @@ builds:
       - darwin
     goarch:
       - amd64
+      - arm
```
nickysemenza, line 17: is this one meant to be darwin? all the other additions are linux

patrickelectric: oooooops, you are correct sir.

patrickelectric: fixed.

PR #1312, opened 2023-06-26
Add DER support for generating and parsing CSR

kongweiguo:
Add new function `GenerateDER` to support generating DER format CSR;
Refactor the old function `Generate`, using `GenerateDER`;
Add parsing DER format CSR support in helpers package;


kongweiguo:
Hi team, 
please help to re-run the golang lint, thks!

PR #1360, opened 2024-02-22
Build images on base image for target platform

### LINKED ISSUE #1359: bug: cfssl/cfssl image builds with wrong base arch
The `cfssl/cfssl:latest` image is currently available for `arm64`, but the base image appears to be `amd64`. This causes issues when trying to run multiple `cfssl` commands with bash.

### Steps to reproduce

On an `arm64` host:

```bash
docker run --entrypoint bash cfssl/cfssl:latest
docker run cfssl/cfssl:latest
```

The first command failes with `exec /bin/bash: exec format error`. The second command runs as expected.  

On an `amd64` host

```base
docker run --platform linux/arm64 --entrypoint bash cfssl/cfssl:latest
docker run --platform linux/arm64 cfssl/cfssl:latest
```

In this case the fist command runs successfully but the second case produces `exec /bin/bash: exec format error` - showing that there are tools with different architectures in the pod.

The solution to this is - I believe - to use a base image that matches the architecture of the target platform.

killianmuldoon:
This change ensure the docker image is built on the same architecture that it's building for. This ensures that other tools in the docker image - e.g. `bash` - will be usable in the image.

Fixes #1359


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20
+FROM --platform=${TARGETPLATFORM} golang:1.20
```
killianmuldoon, line 1: I tested that golang images are available for all platforms currently supported - `linux/arm64`, `linux/s390x` and `linux/amd64`

PR #1387, opened 2024-07-30
Build binaries and container images for ppc64le.

kishen-v:
#### Changes in this PR:
1.  Add support to build binaries and images for ppc64le architecture.

2. Address warnings regarding labels in `LegacyKeyValueFormat` used in the image:
```
 - LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 7)
 - LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 8)
```
Fixed and validated through a build.
```
            "Labels": {
                "org.opencontainers.image.description": "Cloudflare's PKI toolkit",
                "org.opencontainers.image.source": "https://github.com/cloudflare/cfssl"
            }
```

### DIFF
--- a/.github/workflows/docker-builds.yml
+++ b/.github/workflows/docker-builds.yml
@@ -53,7 +53,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
-          platforms: linux/amd64,linux/arm64,linux/s390x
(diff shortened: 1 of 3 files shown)


kishen-v:
Hi @nickysemenza,
Can you PTAL at the changes when time permits? I'm looking forward to having the artifacts built for ppc64le! 

Thanks.

Issue #1237, opened 2022-08-19
testdata certificates should are expiring again


nickysemenza:
similar fix as https://github.com/cloudflare/cfssl/pull/996/files#diff-bce1e9ac488e338e3aa3534151936e9f6e852a224ce9c9cb5e3f55529ca2fb76 is likely needed. Even better would be if they were ephemeral - generated during the tests.

PR #1410, opened 2025-02-08
Mitch292/1237 fix test cases

mitch292:
The goal of this PR is to get the tests in the builds back in working order. It only partially resolves issue #1237.

There were essentially 3 types of cert updates done.
1. Files that were copy pastes of our cf-ssl-trust, we could potentially automate this in the future
2. mTLS test case data where scripts were available for regenerating the certs 
3. Extending a test CA cert via openssl that had the key in the repo.


.github/workflows/go.yml
```diff
@@ -70,9 +70,10 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
           go-version: 1.18
-      - uses: actions/checkout@v4
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v4
+        uses: golangci/golangci-lint-action@v6
```
mitch292, line 78: no harm in bumping this dependency, but lint job fails. Created #1411 to address fixing the lint job

.golangci.yaml
mitch292, line 1: Both of these linters are deprecated and no longer need to be ignored

Makefile
```diff
@@ -60,7 +60,7 @@ snapshot:
 	--rm \
     -v $(PWD):/cross \
     -w /cross \
-    ghcr.io/goreleaser/goreleaser-cross:latest --clean --snapshot --skip-publish
+    ghcr.io/goreleaser/goreleaser-cross:latest --clean --snapshot --skip=publish
```
mitch292, line 63: syntax of this flag has changed and we're using latest. This was causing goreleaser to fail

Makefile
```diff
@@ -73,15 +73,8 @@ github-release:
     -w /cross \
     ghcr.io/goreleaser/goreleaser-cross:latest --clean
 
-.PHONY: docker-build
-docker-build:
-	docker build -f Dockerfile -t cfssl/cfssl:$(VERSION) .
-.PHONY: docker-push
-docker-push:
-	docker push cfssl/cfssl:$(VERSION)
-
```
mitch292, line 82: These are now done as part of the docker-build action

nickysemenza:
nice

